PR #30, opened 2018-01-03
Add transaction.context

mikker:


lib/elastic_apm/context.rb
```diff
@@ -0,0 +1,136 @@
...
+        @headers = {}
+        @cookies = {}
+        @env = {}
+      end
+
+      attr_accessor :body, :cookies, :env, :headers, :http_version, :method,
+        :socket, :url
+
+      # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+      def add_rack_env(rack_env)
```
simitt: I find the handling in `add_rack_env` relatively confusing as I cannot find a consistent usage of virtual attrs, initializing hashes in the init method, accessing and setting attrs, etc. 

If `add_rack_env` would be called twice on the same object, then the information in `socket` would be set according to the second `req`, but `headers` would be basically merged from the first and the second `req`. This is an outcome of once overwriting `socket` completely, but only adding key-values to `headers`, while both initializing in the init method. 
I know this is theoretical, but it shows how different you are initializing and accessing the attrs here. 

Maybe you find a way to make this more consistent.

mikker: Thanks @simitt. You are completely right that this is a bit confusing. I'll try and refactor it to make more sense.

mikker:
@simitt While this in some ways just moves things around, I think it also might make its behaviour less weird -- _theoretically™_ 😀 

The distance between the specs testing this and the functionality might be too long now though. Will have to move things around there as well to keep everyone sane in the long run.

lib/elastic_apm/context.rb
```diff
@@ -0,0 +1,46 @@
...
+module ElasticAPM
+  # @api private
+  class Context
+    # @api private
+    class Request
+      def initialize
+        @socket = {}
+        @headers = {}
+        @cookies = {}
+        @env = {}
```
simitt: I think you are missing `url` here.

mikker: I don't think there's any reason to initialize these anymore.

PR #37, opened 2018-01-12
Add in_app to stacktrace frames

mikker:


beniwohli:
@mikker FYI, we call it `library_frame` now, with the opposite value of `in_app`

mikker:
@beniwohli when you first start renaming stuff:

PR #42, opened 2018-01-17
Freeze config after init to avoid modification.

simitt:
* Freeze config after creation to avoid overwriting some config by incident.
* Remove lock from setting logger as it is not memoized any longer, but directly set when initializing the config.

@mikker I don't have the big picture yet but this is an attempt to avoid overwriting the config by accident anywhere, as it is needed in many places.

### DIFF
--- a/lib/elastic_apm/config.rb
+++ b/lib/elastic_apm/config.rb
@@ -34,8 +34,6 @@ module ElasticAPM
       root_path: Dir.pwd
     }.freeze
 
-    LOCK = Mutex.new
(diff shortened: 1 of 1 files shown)


mikker:
This is good! One thing: We're only `freeze`ing if given a block. The early return should probably just be a regular `if` statement.

mikker:
Perfect! Can I make you squash to a single commit? (Error has nothing to do with you. It's a weird one that I'm still investigating.)

PR #47, opened 2018-01-22
Freeze config during agent start

mikker:


lib/elastic_apm/agent.rb
```diff
@@ -28,7 +28,8 @@ def self.start(config)
 
       LOCK.synchronize do
         return @instance if @instance
-        @instance = new(config).start
+
+        @instance = new(config.freeze).start
```
simitt: You probably want to remove the `freeze` from https://github.com/elastic/apm-agent-ruby/blob/master/lib/elastic_apm/config.rb#L66 then.

mikker: I didn't? wut.

lib/elastic_apm/config.rb
```diff
@@ -62,8 +62,6 @@ def initialize(options = nil)
       end
 
       yield self if block_given?
-
-      freeze
```
mikker: @silvia :)

PR #48, opened 2018-01-25
Return empty process object if pid is not given.

simitt:
Comply to apm-server spec that pid is required
if process object is sent.

### DIFF
--- a/lib/elastic_apm/process_info.rb
+++ b/lib/elastic_apm/process_info.rb
@@ -8,9 +8,11 @@ module ElasticAPM
     end
 
     def build
+      pid = $PID || Process.pid
+      return unless pid
       {
         argv: ARGV,
-        pid: $PID,
+        pid: pid,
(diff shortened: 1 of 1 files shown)


lib/elastic_apm/process_info.rb
```diff
@@ -8,9 +8,11 @@ def initialize(config)
     end
 
     def build
+      pid = $PID || Process.pid
```
beniwohli: I guess `Process.pid` is guaranteed to always have a value?

I'm just wondering why we even check `$PID` first

simitt: Because `Process.pid` is [not available on all platforms](http://ruby-doc.org/core-2.2.0/Process.html#method-c-pid).

beniwohli: ah, gotcha! LGTM then

PR #50, opened 2018-01-31
Update docs from Alpha to Beta.

simitt:
Update to Beta and minor changes in example.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -11,7 +11,7 @@ The recommended way to configure Elastic APM for Rails is to create a file `conf
 [source,yaml]
 ----
 ---
-server_url: 'http://localhost:8200'
+server_url: 'https://localhost:8200'
 secret_token: 'very_very_secret'
 ----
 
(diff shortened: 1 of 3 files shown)


docs/configuration.asciidoc
```diff
@@ -11,7 +11,7 @@ The recommended way to configure Elastic APM for Rails is to create a file `conf
 [source,yaml]
 ----
 ---
-server_url: 'http://localhost:8200'
+server_url: 'https://localhost:8200'
```
mikker: You sure? We provide a self-signed cert for `localhost`?

simitt: No we don't provide a cert. 
I wouldn't create an example with a `secret_token` and `http` though.

simitt: @mikker I removed other changes than setting the `Beta` flag and leave it up to you if you want to change the example in the future.

mikker: @simitt Thank you!

Issue #53, opened 2018-02-09
JRuby concurrency issues


mikker:
The sinatra specs sometimes fail on JRuby because of what looks like a race condition eg. [here](https://apm-ci.elastic.co/job/elastic+apm-agent-ruby+pull-request+multijob-test/134/FRAMEWORK=sinatra-2.0,RUBY_VERSION=jruby-9.1,label=linux/console).

mikker:
Seemingly fixed in #68

PR #57, opened 2018-02-12
Add dropped span count to transaction payloads

### LINKED ISSUE #54: Add max span count to transactions
env: `ELASTIC_APM_TRANSACTION_MAX_SPANS`

APM supports capping span count. When a max span count is reached we should stop collecting more and instead bump `transaction.span_count.dropped.total`.

mikker:
Closes #54


mikker:
CI fail is known jruby bug #53 ☹️

PR #62, opened 2018-02-14
tag docker containers to avoid name clashes

beniwohli:


.gitignore
```diff
@@ -6,6 +6,7 @@
 /pkg/
 /spec/reports/
 /tmp/
+/.vscode/
```
mikker: Shouldn't you put this in your `~/.gitignore` file instead?

beniwohli:
@mikker done. Also renamed the docker image from `spec` to `apm-agent-ruby`

mikker:
Thanks @beniwohli !

PR #68, opened 2018-02-21
(Maybe) Fix JRuby spec woes

mikker:


mikker:
Have reported something that looks like a bug in Webmock. Still investigating. https://github.com/bblimke/webmock/issues/744

PR #69, opened 2018-02-23
Fix typo in inspector.rb

ste26054:


mikker:
Good catch! Thank you.

Jenkins, please test this

PR #80, opened 2018-03-14
Add Elasticsearch instrumentation

mikker:


mikker:
Might have to come up with a way to disable `Net::HTTP` tracing when done inside ES queries. Otherwise we always end up with nested http traces inside.

mikker:
CI error is `rails#master`

PR #87, opened 2018-03-15
Always load config from file if it exists

### LINKED ISSUE #83: Always load config file on boot
Now that we support worker libs that have their own threads and processes we can't count on the user to either use Rails (and load config file in Railtie) or load the file themselves.

Instead let's always load `config/elastic_apm.yml` in `ElasticAPM.start` if it exists.

mikker:
This moves the responsibility to check for and load the config file
from the Railtie to the individual Config. This means every app should
configure by file, Rack/Sinatra as well as Rails.

Closes #83

### DIFF
--- a/lib/elastic_apm/agent.rb
+++ b/lib/elastic_apm/agent.rb
@@ -27,7 +27,7 @@ module ElasticAPM
     def self.start(config)
       return @instance if @instance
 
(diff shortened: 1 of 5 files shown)


mikker:
And the award for Most Persistent CI Error goes to, once again, `rails#master` + Delayed::Job

PR #90, opened 2018-03-23
Change from enabled_ to disabled_injectors

### LINKED ISSUE #85: Make injectors opt-out instead of opt-in
Instead users having to discover available integrations by themselves we should come set up with all bells and whistles and instead let them disabled the stuff they don't want.

Changes `enabled_injectors` to `disabled_injectors`.

mikker:
Closes #85


mikker:
Injectors are undocumented rn. All the docs need a refresher but I'm pushing it off until things have settled a bit.

Issue #84, opened 2018-03-15
Add timer thread to trigger flushing


mikker:
Instead of always waiting for _the next_ transaction to trigger sending recorded transactions to apm-server we should follow the example of the Python agent and have a timer thread that guarantees to at least send every x seconds.

PR #92, opened 2018-04-04
Replace Worker with TimedWorker

mikker:
Instead of having to wait for the next incoming req to flush
batched transactions, TimedWorker will check at least every
flush_interval.

Fixes elastic/apm-agent-ruby#84 

Would love to hear what you think of the approach, @simitt @Qard.


lib/elastic_apm/timed_worker.rb
```diff
@@ -0,0 +1,124 @@
...
+          Thread.exit
+        end
+      end
+    rescue ThreadError
+      # queue empty
+    end
+    # rubocop:enable Metrics/MethodLength
+
+    def process_request(req)
+      @adapter.post(req.path, req.payload)
```
Qard: So is this sending each transaction in a separate http request then?

mikker: No, it's sending a bundle of batched transactions from `batched_transactions`.

lib/elastic_apm/timed_worker.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
```
simitt: Have you considered using a [TimedWorker](http://ruby-concurrency.github.io/concurrent-ruby/Concurrent/TimerTask.html)? I haven't used this myself, but it seems to be created exactly for the purpose of running a task on a scheduled time basis. I think it would be a good design, to have the timer thread separated from the business logic here.

lib/elastic_apm/timed_worker.rb
```diff
@@ -0,0 +1,124 @@
...
+      @batched_transactions = batched_transactions
+
+      @last_sent_transactions = Time.now.utc
+
+      @serializers = Struct.new(:transactions, :errors).new(
+        Serializers::Transactions.new(config),
+        Serializers::Errors.new(config)
+      )
+
+      @rand = rand(999)
```
simitt: What is this for?

lib/elastic_apm/timed_worker.rb
```diff
@@ -0,0 +1,124 @@
...
+      loop do
+        run_once
+        sleep SLEEP_INTERVAL
+      end
+    end
+
+    def run_once
+      process_messages
+
+      flush_transactions if should_flush_transactions?
```
simitt: I think the order of `process_messages` and `flush_transactions` should be changed, or `process_messages` should be called from `flush_transactions` at the end.

lib/elastic_apm/timed_worker.rb
mikker: Nothing. Used it for debug – good catch!

mikker:
@simitt Do you like these renames?

lib/elastic_apm/timed_worker.rb
```diff
@@ -0,0 +1,113 @@
...
+    def process_messages
+      while (msg = messages.pop(true))
+        case msg
+        when ErrorMsg
+          post_error msg
+        when StopMsg
+          # empty collected transactions before exiting
+          collect_and_send_transactions
+
+          Thread.exit
```
Qard: Rather than exiting here, would it perhaps be better to mark a flag and exit when the empty queue exception raises? Otherwise you might still have messages in the queue.

mikker: Good idea!

lib/elastic_apm/timed_worker.rb
mikker: @Qard I’m hesitant to add dependencies. I’m not completely sure if it’s really a wasted effort but at least I’m putting it off as much as possible.

lib/elastic_apm/timed_worker.rb
```diff
@@ -0,0 +1,113 @@
...
+        debug e.backtrace.join("\n")
+        nil
+      end
+    end
+
+    def collect_batched_transactions
+      batch = []
+
+      begin
+        while (transaction = pending_transactions.pop(true))
```
simitt: Consuming the the `transactions` here could become an issue, when events are enqueued faster than they can be processed. If new events get pushed to the queue fast enough, the condition `while queue.pop(true)` is always fulfilled. This would eventually lead to only a few requests to the server with large payloads. 

It might be an edge case, but if there are tons of requests to an endpoint I think this could happen. You probably need to additionally set a limit when consuming from the queue so the number of transactions that are sent within 1 request cannot grow endlessly.

mikker: Oh yeah, you are right. What do you suggest? Something like a limit to how many transaction can be in a batch? 
`while pop && batch.length < MAX_BATCH_SIZE` ?

simitt: Yes adding a `batch.length` check. How about you use the configured `max_queue_size` for an upper limit?

lib/elastic_apm/timed_worker.rb
```diff
@@ -50,21 +50,25 @@ def run_once
 
     private
 
+    # rubocop:disable Metrics/MethodLength
     def process_messages
+      should_exit = false
+
       while (msg = messages.pop(true))
```
simitt: Giving my previous comment a second thought, something similar could happen here: E.g a bad deploy happens that causes tons of errors to be sent, the `while` loop would go on for quite some time, blocking transactions to be sent, as the thread gets stuck in processing the enqueued `messages`. 

With the changes that you are proposing in this PR the error messages are not directly sent when they occur anymore, but eventually when the thread processes the queue. With this change, have you considered to send all the currently queued errors at once, instead of one by one?

mikker: So essentially doing the same for errors as for transactions?

simitt: basically yes, but with a `should_flush_errors?` check that returns `true` as soon as an error is enqueued.

Issue #94, opened 2018-04-10
Binary Redis payloads break APM reporting


cheald:
We have some legacy code which is using Redis for session management, which among other things uses Marshal.dump to store Ruby objects in Redis. This results in non-UTF-8 byte sequences in the command, which breaks Elastic APM when it tries to dump the payload to UTF-8 JSON.

Here's a quick repro case (I added this to the Redis injector, but it doesn't really belong there. It demonstrates the break though.)

```ruby
    it "breaks on binary query payloads" do
      redis = ::Redis.new
      ElasticAPM.start

      transaction = ElasticAPM.transaction 'T' do
        redis.set("session", Marshal.dump(0xdeadbee))
      end.submit 200

      ElasticAPM.agent.instance_variable_get("@serializers").transactions.build_all([transaction]).to_json
      ElasticAPM.stop
    end
```

```
  2) Injectors::RedisInjector breaks on binary query payloads
     Failure/Error:
                 alias :"__without_apm_#{method}" :"#{method}"

                 def #{method}(*args, &block)
                   unless ElasticAPM.current_transaction
                     return __without_apm_#{method}(*args, &block)
                   end

                   ElasticAPM.span "#{name}", "#{type}" do
                     __without_apm_#{method}(*args, &block)

     Encoding::UndefinedConversionError:
       "\xEE" from ASCII-8BIT to UTF-8
     # /home/chris/.rvm/gems/ruby-2.4.3/gems/activesupport-4.2.10/lib/active_support/core_ext/object/json.rb:34:in `encode'
     # /home/chris/.rvm/gems/ruby-2.4.3/gems/activesupport-4.2.10/lib/active_support/core_ext/object/json.rb:34:in `to_json'
     # /home/chris/.rvm/gems/ruby-2.4.3/gems/activesupport-4.2.10/lib/active_support/core_ext/object/json.rb:34:in `to_json_with_active_support_encoder'
     # /home/chris/.rvm/gems/ruby-2.4.3/gems/activesupport-4.2.10/lib/active_support/json/encoding.rb:57:in `to_json'
     # ./lib/elastic_apm/span_helpers.rb:20:in `generate'
     # /home/chris/.rvm/gems/ruby-2.4.3/gems/activesupport-4.2.10/lib/active_support/json/encoding.rb:101:in `stringify'
[...]

PR #95, opened 2018-04-11
Omit collecting full redis statements

mikker:
Statements could very well include very large or confidential data, so for now we'll just omit them from APM.

Closes elastic/apm-agent-ruby#94


mikker:
CI failure is related to delayed_job things fixed in #92

Issue #93, opened 2018-04-10
Compress payloads before sending


mikker:
apm-server accepts compressed requests and there's a lot of bytes to be saved with our kind of data.

Prob. using https://ruby-doc.org/stdlib-2.5.0/libdoc/zlib/rdoc/Zlib.html

PR #97, opened 2018-04-11
Add support for HTTP compression

cheald:
Non-compressed payloads are frequently failing to post with broken pipe errors due to the excessive size  of the payloads. This is due in part to the tremendous size of backtraces captured for transactions, but is helped by deflating the content prior to posting. This satisfies #93, which helps, but probably won't be sufficient for production usage by itself.

Adds 3 config options:

* http_compression: bool (default true)        - Turns compression on or off
* compression_minimum_size: int (default 5120) - Sets the minimum payload length to deflate.
                                                 Smaller payloads will not be deflated.
* compression_level: int (default 6)           - Sets the zlib deflate level to use.

The 5k minimum size is arbitrarily chosen. I've seen recommendations as low as 1kb and as high as 10kb.

### DIFF
--- a/lib/elastic_apm/config.rb
+++ b/lib/elastic_apm/config.rb
@@ -35,6 +35,9 @@ module ElasticAPM
       debug_transactions: false,
       debug_http: false,
       verify_server_cert: true,
+      http_compression: true,
+      compression_minimum_size: 1024 * 5,
+      compression_level: 6,
 
       source_lines_error_app_frames: 5,
       source_lines_span_app_frames: 5,
@@ -119,6 +122,9 @@ module ElasticAPM
     attr_accessor :http_open_timeout
     attr_accessor :debug_transactions
     attr_accessor :debug_http
+    attr_accessor :http_compression
+    attr_accessor :compression_minimum_size
+    attr_accessor :compression_level
 
     attr_accessor :source_lines_error_app_frames
     attr_accessor :source_lines_span_app_frames
--- a/lib/elastic_apm/http.rb
+++ b/lib/elastic_apm/http.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'net/http'
+require 'zlib'
 
 require 'elastic_apm/service_info'
 require 'elastic_apm/system_info'
@@ -32,6 +33,7 @@ module ElasticAPM
     def post(path, payload = {})
       payload.merge! @base_payload
       filters.apply(payload)
+
       request = prepare_request path, payload.to_json
       response = @adapter.perform request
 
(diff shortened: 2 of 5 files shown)


mikker:
This is great, @cheald! Thank you. Taking a look right away. (Is this payment for solving the redis issue? 😀)

Jenkins. please test this

cheald:
Hah, I'm just trying to get Elastic APM to the point that we can use it in production. :)

To that point, payload size is still an issue (each request is generating ~3-4MB of JSON because every span in the transaction has very chatty backtraces! I'm seeing if I can figure out how to wire up a config option to turn off (or limit) transaction traces. Does that seem like a sane approach to you?

cheald:
That definitely helps. With sampling and compression turned on, my test payloads have dropped from multiple megabytes to 50kb - far better!

A possibly-related issue - when testing with compression _off_ I am periodically seeing "end of file reached" errors from apm-server, with payloads in the 1MB range. I'm not sure if it's an issue with the payload size, or an incorrect accounting of the content-length header yet, or something else, but it's showing up often enough that it's worth noting. I haven't seen it with compression on yet.

(Apologies for sort of wandering all over the place with this PR!)

mikker:
I've sometimes seen that as well. I'll keep an eye out for it. Do you have any feeling for the performance impact from the compression? And maybe also regarding choosing the right default compression level?

cheald:
I haven't measured the performance impact yet, but Ruby's zlib is relatively quick, as Ruby goes. Since the actual compression isn't happening in the request threads, I wouldn't expect there to be much user-facing impact outside of the CPU burn, and even that is a lower concern than usual - as of Ruby 2.0, [Ruby releases the GIL while zlib is inflating/deflating](https://bugs.ruby-lang.org/issues/6615), so in theory, the compression work may be running on a separate core while Ruby is serving requests, and it shouldn't block the Ruby program from running unless all CPUs are saturated.

I set the default level at 6 since that's the library default if left unspecified. It may be appropriate to lower that since we expect that this is likely to be more time-sensitive than space-sensitive on the margin.

Documentation to the effect of "lower compression levels mean less CPU, but larger payloads" is probably appropriate - the tradeoff will be different for different users. It might be worth generating a largeish payload and then benchmarking different compression for speed and size to get a feel for a good middle ground default.

cheald:
Ah, that could be `Zlib::Inflate.inflate`. I'll patch it. Curious that JRuby's API isn't compatible in this case.

mikker:
Yeah it's weird that they missed the module class methods. Thanks again!

mikker:
jenkins, please test this (again)

mikker:
Can you rebase this on master? I think the CI error is fixed in https://github.com/elastic/apm-agent-ruby/commit/999481c90bcb9f4f3170919e510ae49f3eb4bff3

Can I see on GH what commit goes before this PR?

cheald:
Yup, will rebase.

https://github.com/cheald/apm-agent-ruby/commits/http-compression has the history. It looks like it's currently based off of 9616d869de090e19117741857d5c459bd6a9a7c8

cheald:
Nevermind, it looks like it's already rebased! I'm not sure what the cause for that error is, then.

mikker:
Looks like it's the same error but in the Rails spec instead. You can totally just copy-paste the _fix_ from the sinatra spec.

cheald:
Okay, I _think_ that should fix it. I have two unrelated specs failing locally under JRuby, but the Rails integration is fixed.

Incidentally, the JRuby failure is fixed in https://github.com/jruby/jruby/commit/a512032e60cc2a3011a44ce74569eb968c5129c0 :)

mikker:
jenkins, once again, please test this

mikker:
All green, merging 💙💚💛💜❤️

mikker:
Again, thanks a bunch for this. I'd **love** to hear how APM works for you if you get it running for a few days – don't hesitate to write me at _mikkel@elastic.co_.

cheald:
Sounds good. I'm currently evaluating it, and it looks like it's going to do what we need quite competently. It'll be a bit before it ends up in prod, but I'll not hesitate to send any feedback your way!

PR #100, opened 2018-04-16
Add Mongo injector

### LINKED ISSUE #78: Add Mongo support

mikker:
Fixes elastic/apm-agent-ruby#78

### DIFF
--- a/spec/elastic_apm/injectors/mongo_spec.rb
+++ b/spec/elastic_apm/injectors/mongo_spec.rb
@@ -1,0 +1,41 @@
+require 'spec_helper'
+require 'mongo'
+
+module ElasticAPM
+  RSpec.describe 'Injectors::MongoInjector' do
+    it 'instruments calls', :with_fake_server do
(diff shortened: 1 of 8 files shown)


spec/elastic_apm/injectors/mongo_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+require 'mongo'
+
+module ElasticAPM
+  RSpec.describe 'Injectors::MongoInjector' do
+    it 'instruments calls', :with_fake_server do
+      ElasticAPM.start flush_interval: nil
+
+      client =
+        Mongo::Client.new(
+          ['127.0.0.1:27017'],
```
beniwohli: this needs to be `mongodb:27017`. You could set that as environment variable in the `ruby_rspec` service, and fall back to `127.0.0.1:27017` if the environment variable isn't set, so the test also can be run outside of the docker setup

docker-compose automatically makes

Issue #75, opened 2018-03-13
Add Resque instrumentation and error reporting


mikker:
As Resque (by default) forks the process around every job this was a bit tricky. We want this eventually but for now it's put off until after 1.0.

alvarolobato:
There were some additional problems with this integration (integration with the external process), we've decided to wait after GA to implement this integration.

darrellnash:
Today I upgraded to v2.0.1 of this gem and found our Resque jobs stopped sending data to the apm-server. After some digging through the gem code, I found a workaround:

You want to force the Agent to restart in a fresh state for each newly forked process, which can be done by adding an `after_fork` hook to Resque. For example, you can create a rails initializer such as `config/initializers/resque_initializer.rb`:

```ruby
Resque.after_fork do
  ElasticAPM.stop if ElasticAPM.running?
  ElasticAPM.start
end
```

The problem is that the workers are booted before the fork, and apparently some state in the ConcurrentRuby thread pool doesn't transfer over well to the child process (or maybe just the threads are in a dead state in the child?). So while transactions and spans were getting queued up, there were no worker threads to pull jobs off the queue to send to the apm-server.

When the `ElasticAPM.stop` call was triggered at the end, the ConcurrentRuby `@pool.shutdown` and `@pool.wait_for_termination(5)` calls would work fine because there are no _valid_ worker threads to shutdown, even though there are plenty of queued up messages. And apparently `@workers` still had the initial pre-fork booted worker objects, but again, they weren't actually executing. Rebooting the Agent fixed my woes.

Not the first time mixing multi-threaded and multi-processed models has caused me brain damage ;)

Hope this helps troubleshoot the auto-instrumentation effort.

mikker:
Hi @darrellnash. Resque instrumentation has never been officially supported. As you see it caused us some trouble. Thanks for your thorough investigation and description. I'll take another look at it when the dust of 2.0 settles.

tmedford:
Any chance at supporting resque?

mikker:
I'm afraid the story is still the same as above: We'd love to support Resque but are working on other things at the moment. PRs are welcome 😊

sundarv85:
Any chance of this being supported? As we are also using resque.

mikker:
We are still not actively working on a solution, so the status hasn't changed.

We do however have more devs on the agent now, so perhaps it's time to take another look. I'll bring it up 👍

sundarv85:
[...]

PR #112, opened 2018-05-08
Resque instrumentation (WIP)

mikker:
Closes elastic/apm-agent-ruby#75

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,10 +13,11 @@ gem 'mongo'
 gem 'pry'
 gem 'rack-test'
 gem 'redis', require: nil
+gem 'resque', require: nil
 gem 'rspec'
 gem 'rubocop'
 gem 'sequel'
-gem 'sidekiq'
+gem 'sidekiq', require: nil
 gem 'timecop'
 gem 'webmock'
 gem 'yard'
--- a/lib/elastic_apm/agent.rb
+++ b/lib/elastic_apm/agent.rb
@@ -103,7 +103,6 @@ module ElasticAPM
     def enqueue_transaction(transaction)
       boot_worker unless worker_running?
 
-      puts "Thread:#{Thread.current.object_id}"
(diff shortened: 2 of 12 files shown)


mikker:
@simitt I'm having trouble again – in the forked version, Spans aren't added to the Transaction. Can you see what the trouble might be?

mikker:
Thinking more about I think the problem might be that the spans are added in the forked version of the transaction and therefore not brought back after the job has run. Can't see a solution to this right now other than going back to some kind of inter-process comm. but will think some more.

simitt:
Yes I am pretty sure the issue is that the transaction is started in another process. I didn't have time to look into it again yet, and probably will be busy with something else a view days. 
We probably could look into starting the transaction+spans inside the child process but yielding it back to the parent process before the child is stopped.

simitt:
As just discussed over zoom with @mikker, I did a little POC for using `IO.pipe` for inter-process communication, see https://github.com/simitt/apm-agent-ruby/commit/96bcea5b0c9dada30f523b500311bd1f9139d24c

Using pipes seems relatively straight forward code wise. 
However, it is pretty easy to block the reading thread by not sending anything to the pipe or by not using a delimiter `\n` as the reader will then wait forever. 
As a `transaction` cannot simply be dumped with `Marshal.dump`, manual (de-)serialization is necessary. In the POC I use the existing serialization code to enable dumping the transaction. However, for loading the transaction additional code for manual deserialization would be necessary. Or, as an alternative the serialized transaction could be added as is to the transactions array that gets sent to the Server.

To summarize, using `IO.pipe` looks like a good concept for handling interprocess communication, but as neither @mikker nor I have experience with possible pitfalls of using `IO.pipe` and quite some additional tests would be required to make the code resilient, it might be a better option to implement a synchronous call to the APM Server per Resque job for now and revisit after going GA if needed.

PR #122, opened 2018-05-16
Remove ActiveSupport dependency

### LINKED ISSUE #71: Consider removing ActiveSupport dep
Every `Instrumenter` has a `Subscriber` now. The `Subscriber` is only used when tracing ActiveSupport instrumentation. To remove AS we need to make the `Subscriber` optional.

mikker:
Closes elastic/apm-agent-ruby#71

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,6 +38,3 @@ when /.+/
 else
   gem framework
 end
-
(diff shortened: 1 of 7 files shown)


mikker:
Looks like CI is happy. Going to run this in an example app to get _real_ confirmation.

PR #125, opened 2018-05-28
Add updated documentation

### LINKED ISSUE #113: Update Docs for GA

mikker:
Closes #113

### DIFF
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -1,7 +1,8 @@
 [[api]]
 == Public API
 
-Although most usage is covered automatically when using the APM Agent it also has a public API that allows custom usage.
+Although most usage is covered automatically, Elastic APM also has a public API that
+allows custom usage.
 
 [float]
 [[agent-life-cycle]]
@@ -10,7 +11,7 @@ Although most usage is covered automatically when using the APM Agent it also ha
 Controlling when the agent starts and stops.
 
 [float]
-[[api-start]]
+[[api-agent-start]]
 ==== `ElasticAPM.start`
 
 To create and start an ElasticAPM agent use `ElasticAPM.start`:
@@ -20,26 +21,27 @@ To create and start an ElasticAPM agent use `ElasticAPM.start`:
 ElasticAPM.start(server_url: 'http://localhost:8200')
 ----
 
-  * `config`: A hash or `ElasticAPM::Config` instance with configuration options. See <<configuration,Configuration>>.
+  * `config`: An optional hash or `ElasticAPM::Config` instance with configuration
+  options.  See <<configuration,Configuration>>.
 
-If you are using Rails this is done automatically for you.
-
-**ElasticAPM expects a single instance of the Agent.**
+If you are using <<getting-started-rails,Ruby on Rails>> this is done automatically for you.
+If not see <<getting-started-rack,Getting started with Rack>>.
 
 [float]
-[[api-stop]]
+[[api-agent-stop]]
 ==== `ElasticAPM.stop`
 
-Stop the currently running agent. Use this inside `at_exit` in your <<getting-started-rack,Rack app>> to gracefully shut down.
+Stop the currently running agent. Use this inside `at_exit` in your
+<<getting-started-rack,Rack app>> to gracefully shut down.
 
 [float]
-[[api-running]]
+[[api-agent-running]]
 ==== `ElasticAPM.running?`
 
 Returns whether the ElasticAPM Agent is currently running.
 
 [float]
-[[api-agent]]
+[[api-agent-agent]]
 ==== `ElasticAPM.agent`
 
 Returns the currently running agent or nil.
@@ -48,45 +50,57 @@ Returns the currently running agent or nil.
 === Instrumentation
 
 [float]
-[[api-current-transaction]]
+[[api-agent-current-transaction]]
 ==== `ElasticAPM.current_transaction`
 
 Returns the current `ElasticAPM::Transaction` or nil.
 
 [float]
-[[api-transaction]]
+[[api-agent-transaction]]
 ==== `ElasticAPM.transaction`
 
 Start a _transaction_ eg. a web request or background job.
 
-If called without a block you are expected to end and submit the transaction yourself.
-If given a block, the code inside will be wrapped in a transaction. You still need to submit it yourself with `transaction.submit(status)`.
+If called without a block you are expected to end the transaction yourself.
+If given a block, the code inside will be wrapped in a transaction.
+
+You need to submit it yourself with `transaction.submit(status)`.
 
 [source,ruby]
 ----
+# call with block
 transaction = ElasticAPM.transaction('Do things') do
-  # ...
-end
+  do_work # ...
+end # .done is called when block ends
 
 transaction.submit(200)
+
+# without block
+transaction = ElasticAPM.transaction('Do things')
+do_work
+transaction.submit(200) # .submit(result) will also .done(result)
 ----
 
 Arguments:
 
   * `name`: A name for your transaction. Transactions are grouped by name. **Required**.
   * `type`: A `type` for your transaction eg. `db.postgresql.sql`.
-  * `rack_env: env`: An optional Rack `env` used to enrich the transaction with information about the current request.
-  * `&block`: An optional block to wrap with the transaction. The block is passed the transaction as an optional argument.
+  * `context:`: An optional <<api-context,Context>> used to enrich the transaction with
+  information about the current request.
+  * `&block`: An optional block to wrap with the transaction. The block is passed the
+  transaction as an optional argument.
 
 Returns the transaction.
 
 [float]
-[[api-span]]
+[[api-agent-span]]
 ==== `ElasticAPM.span`
 
-Most transactions have nested spans signifying work of a specific sort eg. database queries or external web requests.
+Most transactions have nested spans signifying work of a specific sort eg. database
+queries or external web requests.
 
-If given a block, wrap the code inside in a span. If not you are expected to close the span yourself with `span.done(result)`.
+If given a block, wrap the code inside in a span.
+If not you are expected to close the span yourself with `span.done(result)`.
 
 [source,ruby]
 ----
@@ -102,12 +116,14 @@ Arguments:
   * `name`: A name for your span. **Required**.
   * `type`: The type of work eg. `db.postgresql.query`.
   * `context`: An instance of `Span::Context`.
-  * `&block`: An optional block to wrap with the span. The block is passed the span as an optional argument.
+  * `include_stacktrace`: Whether or not to collect a Stacktrace.
+  * `&block`: An optional block to wrap with the span.
+  The block is passed the span as an optional argument.
 
 Return the span or the return value of the given block.
 
 [float]
-[[api-build-context]]
+[[api-agent-build-context]]
 ==== `ElasticAPM.build_context`
 
 Build a new _Context_ from a Rack `env`.
@@ -124,7 +140,7 @@ Returns the built context.
 === Errors
 
 [float]
-[[api-report]]
+[[api-agent-report]]
 ==== `ElasticAPM.report`
 
 Send an `Exception` to Elastic APM.
@@ -143,12 +159,13 @@ end
 Arguments:
 
   * `exception`: An instance of `Exception`. **Required**.
-  * `handled`: Whether the error was _handled_ eg. wasn't rescued and was represented to the user. Default: `true`.
+  * `handled`: Whether the error was _handled_ eg. wasn't rescued and was represented
+  to the user. Default: `true`.
 
 Returns `[ElasticAPM::Error]`.
 
 [float]
-[[api-report-message]]
+[[api-agent-report-message]]
 ==== `ElasticAPM.report_message`
 
 Send a custom message to Elastic APM.
@@ -163,7 +180,6 @@ ElasticAPM.report_message('This should probably never happen?!')
 Arguments:
 
   * `message`: A custom error string. **Required**.
-  * `handled`: Whether the error was _handled_ eg. wasn't rescued and was represented to the user. Default: `true`.
 
 Returns `[ElasticAPM::Error]`.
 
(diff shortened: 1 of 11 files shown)


mikker:
(Thank you @beniwohli for many of the descriptions)

mikker:
@simitt Most of this is old news but a light skim would be great if you have the time?

docs/api.asciidoc
```diff
@@ -20,7 +21,8 @@ To create and start an ElasticAPM agent use `ElasticAPM.start`:
 ElasticAPM.start(server_url: 'http://localhost:8200')
 ----
 
-  * `config`: A hash or `ElasticAPM::Config` instance with configuration options. See <<configuration,Configuration>>.
+  * `config`: An optional hash or `ElasticAPM::Config` instance with configuration
+  options.  See <<configuration,Configuration>>.
 
 If you are using Rails this is done automatically for you.
```
simitt: Could you add a reference to the page where you show how to instrument a rack app here?

docs/api.asciidoc
```diff
@@ -30,7 +32,8 @@ If you are using Rails this is done automatically for you.
 [[api-stop]]
 ==== `ElasticAPM.stop`
 
-Stop the currently running agent. Use this inside `at_exit` in your <<getting-started-rack,Rack app>> to gracefully shut down.
+Stop the currently running agent. Use this inside `at_exit` in your
+<<getting-started-rack,Rack app>> to gracefully shut down.
```
simitt: I cannot add a comment on line 29 so I add it here:
**ElasticAPM expects a single instance of the Agent.**

Can you add a little bit more description here please. Does this mean I cannot have multiple agents in multiple threads/processes?

docs/api.asciidoc
```diff
@@ -59,8 +62,10 @@ Returns the current `ElasticAPM::Transaction` or nil.
 
 Start a _transaction_ eg. a web request or background job.
 
-If called without a block you are expected to end and submit the transaction yourself.
-If given a block, the code inside will be wrapped in a transaction. You still need to submit it yourself with `transaction.submit(status)`.
+If called without a block you are expected to end the transaction yourself.
```
simitt: There is no code snippet that shows how to end a transaction.

docs/api.asciidoc
```diff
@@ -75,18 +80,22 @@ Arguments:
 
   * `name`: A name for your transaction. Transactions are grouped by name. **Required**.
   * `type`: A `type` for your transaction eg. `db.postgresql.sql`.
-  * `rack_env: env`: An optional Rack `env` used to enrich the transaction with information about the current request.
-  * `&block`: An optional block to wrap with the transaction. The block is passed the transaction as an optional argument.
+  * `context:`: An optional Context used to enrich the transaction with
```
simitt: How should the context look like? (Is it a hash, is it described somewhere you can reference to?)

docs/api.asciidoc
```diff
@@ -75,18 +80,22 @@ Arguments:
...
+  information about the current request.
+  * `&block`: An optional block to wrap with the transaction. The block is passed the
+  transaction as an optional argument.
```
simitt: The formulation `to wrap with the transaction`  is confusing me a bit. I think simply saying `An optional block that gets passed to the transaction` should be sufficient.

docs/api.asciidoc
```diff
@@ -75,18 +80,22 @@ Arguments:
...
 [float]
 [[api-span]]
 ==== `ElasticAPM.span`
 
-Most transactions have nested spans signifying work of a specific sort eg. database queries or external web requests.
+Most transactions have nested spans signifying work of a specific sort eg. database
+queries or external web requests.
 
-If given a block, wrap the code inside in a span. If not you are expected to close the span yourself with `span.done(result)`.
+If given a block, wrap the code inside in a span.
```
simitt: Could you align the formulation with the transaction:

> If given a block, the code inside will be wrapped in a transaction.

Could you generally add a hint which spans are automatically collected, and also relate this to manually trace transactions.

docs/api.asciidoc
```diff
@@ -102,7 +111,9 @@ Arguments:
   * `name`: A name for your span. **Required**.
   * `type`: The type of work eg. `db.postgresql.query`.
   * `context`: An instance of `Span::Context`.
-  * `&block`: An optional block to wrap with the span. The block is passed the span as an optional argument.
+  * `include_stacktrace`: Whether or not to collect a Stacktrace.
```
simitt: Do you describe `Span::Context` somewhere?

There is a `ElasticAPM.build_context` description below, but you don't motivate what this context can be used for. Could you please clarify which context to use where and how to create?

docs/api.asciidoc
```diff
@@ -102,7 +111,9 @@ Arguments:
...
+  * `&block`: An optional block to wrap with the span.
+  The block is passed the span as an optional argument.
```
simitt: Same comment as for `transaction`

docs/api.asciidoc
```diff
@@ -174,12 +185,14 @@ Returns `[ElasticAPM::Error]`.
 [[api-set-tag]]
 ==== `ElasticAPM.set_tag`
 
-Add a tag to the current transaction. Tags are basic key-value pairs that are indexed in your Elasticsearch database and therefore searchable.
+Add a tag to the current transaction.
```
simitt: Could you add a description related to the header `Context`.
Why is this not part of the section `Transaction` as it seems to only refer to Transactions and not to Errors?

(Actually is the API method `ElasticAPM.set_tag` a convenience method over ElasticAPM.current_transaction.set_tag`?)

docs/api.asciidoc
```diff
@@ -223,3 +239,29 @@ Arguments:
...
+
+Example:
+
+[source,ruby]
+----
+ElasticAPM.add_filter(:filter_pings) do |payload|
+  payload['transactions']&.reject! do |t|
+    t['name'] == 'PingsController#index'
+  end
+end
```
simitt: You need to use symbols over strings here ;)

docs/configuration.asciidoc
```diff
@@ -15,20 +13,319 @@ server_url: 'http://localhost:8200'
...
+This string is used to ensure that only your agents can send data to your APM server.
+Both the agents and the APM server have to be configured with the same secret token.
+One example to generate a secure secret token is:
+
+[source,bash]
+----
+ruby -r securerandom -e 'print SecureRandom.uuid'
+----
+
+WARNING: Secret tokens only provide any real security if your APM server use TLS.
```
simitt: _uses_  instead of _use_

docs/configuration.asciidoc
```diff
@@ -15,20 +13,319 @@ server_url: 'http://localhost:8200'
...
+| Environment                        | `Config` key           | Default
+| `ELASTIC_APM_ENABLED_ENVIRONMENTS` | `enabled_environments` | `['production']`
+|============
+
+Specify which environments to enable APM in.
+
+If the current envrironment isn't included, the agent will effectively be a _noop_
+and do nothing.
+So you can keep in your custom instrumentation code without the agent sending
+anything nor complain.
```
simitt: Will it only not send anything or will it not do anything?

docs/configuration.asciidoc
```diff
@@ -15,20 +13,319 @@ server_url: 'http://localhost:8200'
...
+==== `source_lines_error_app_frames`
+[float]
+[[config-source-lines-error-library-frames]]
+==== `source_lines_error_library_frames`
+[float]
+[[config-source-lines-span-app-frames]]
+==== `source_lines_span_app_frames`
+[float]
+[[config-source-lines-span-library-frames]]
+==== `source_lines_span_library_frames`
```
simitt: How about `Number of stacktrace frames to collect` or something similar instead of having 4 headers here?

docs/configuration.asciidoc
```diff
@@ -15,20 +13,319 @@ server_url: 'http://localhost:8200'
...
+[float]
+[[config-max-queue-size]]
+==== `max_queue_size`
+
+|============
+| Environment                  | `Config` key     | Default
+| `ELASTIC_APM_MAX_QUEUE_SIZE` | `max_queue_size` | `500`
+|============
+
+Maximum queue length of transactions before sending transactions to the APM server.
```
simitt: I'd prefer maximum `number` of collected transactions instead of `queue length`

docs/configuration.asciidoc
```diff
@@ -15,20 +13,319 @@ server_url: 'http://localhost:8200'
...
+A lower value will increase the load on your APM server,
```
simitt: I would not state that, as it would lead to smaller request bodies and a more consistent load on the Server, so I'd actually expect this to have a good impact on the server.
It might increase the load on the agent for sending more http requests than otherwise.

docs/configuration.asciidoc
```diff
@@ -15,20 +13,319 @@ server_url: 'http://localhost:8200'
...
+[[config-flush-interval]]
+==== `flush_interval`
+
+|============
+| Environment                  | `Config` key     | Default
+| `ELASTIC_APM_FLUSH_INTERVAL` | `flush_interval` | `10`
+|============
+
+Interval with which transactions should be sent to the APM server, in seconds.
+A lower value will increase the load on your APM server,
```
simitt: same as above..

docs/configuration.asciidoc
```diff
@@ -15,20 +13,319 @@ server_url: 'http://localhost:8200'
...
+while a higher value can increase the memory pressure on your app.
+A higher value also impacts the time until transactions are indexed and searchable
+in Elasticsearch.
+
```
simitt: Could you add a sentence that when both values are configured, transactions are sent whenever one of the two criterias are reached.

docs/configuration.asciidoc
```diff
@@ -15,20 +13,319 @@ server_url: 'http://localhost:8200'
...
+|============
+| Environment                           | `Config` key              | Default
+| `ELASTIC_APM_TRANSACTION_SAMPLE_RATE` | `transaction_sample_rate` | `1.0`
+|============
+
+By default, the agent will sample every transaction (e.g. request to your service).
+To reduce overhead and storage requirements, you can set the sample rate to a value
+between `0.0` and `1.0`.
+We still record overall time and the result for unsampled transactions, but no
+context information, tags, or spans.
```
simitt: Did you align with the other agent owners on the default value? I am surprised that you start with 1.0.

docs/configuration.asciidoc
```diff
@@ -15,20 +13,319 @@ server_url: 'http://localhost:8200'
...
+|============
+| Environment                         | `Config` key            | Default
+| `ELASTIC_APM_TRANSACTION_MAX_SPANS` | `transaction_max_spans` | `500`
+|============
+
+Limits the amount of spans that are recorded per transaction.
+This is helpful in cases where a transaction creates a very high amount of spans
+(e.g. thousands of SQL queries).
+Setting an upper limit will prevent overloading the agent and the APM server with
+too much work for such edge cases.
```
simitt: and the ES storage..

docs/configuration.asciidoc
```diff
@@ -15,20 +13,319 @@ server_url: 'http://localhost:8200'
...
+[[config-verify-server-cert]]
+==== `verify_server_cert`
+|============
+| Environment                       | `Config` key         | Default
+| `ELASTIC_APM_VERIFY_SERVER_CERT`  | `verify_server_cert` | `true`
+|============
+
+By default, the agent verifies the SSL certificate if you use an HTTPS connection to
+the APM server.
+Verification can be disabled by changing this setting to `false`.
```
simitt: I'd add a  note, that this is generally not recommended (but might be helpful for testing)

docs/configuration.asciidoc
```diff
@@ -15,20 +13,319 @@ server_url: 'http://localhost:8200'
...
+[[config-disabled-spies]]
+==== `disabled_spies`
+
+[options="header"]
+|============
+| Environment                  | `Config` key     | Default
+| `ELASTIC_APM_DISABLED_SPIES` | `disabled_spies` | `['json']`
+|============
+
+Elastic APM automatically instruments select third party libraries.
```
simitt: _selected_ third party libraries

Use this option to _disable_ any of these.

Could you again link to a list of supported third party libraries?

docs/custom-instrumentation.asciidoc
```diff
@@ -1,14 +1,85 @@
...
+  def do_hard_work
+    # ...
+  end
+
+  span_method :do_hard_work # takes optional `name` and `type`
+end
+----
+
+[float]
+==== Custom span example
```
simitt: What's the difference between this example and the one right above? If I am not mistaken both show how a custom span can be collected?

docs/custom-instrumentation.asciidoc
```diff
@@ -1,14 +1,85 @@
...
+
+    begin
+      Sequel[:users] # many third party libs will be automatically instrumented
+      transaction.submit('success') if transaction
+    rescue Exception => e
+      ElasticAPM.report(e)
+      transaction.submit('error') if transaction
+      raise
+    ensure
+      transaction.release
```
simitt: You haven't described that a transaction also needs to be `released` when describing the Public API.

docs/custom-instrumentation.asciidoc
```diff
@@ -1,14 +1,85 @@
 [[custom-instrumentation]]
 === Custom instrumentation
 
-When installed ElasticAPM will automatically wrap your app's request/responses in transactions and report its errors.
+When installed ElasticAPM will automatically wrap your app's request/responses in
+transactions and report its errors.
+It also wraps each background job if you use Sidekiq or DelayedJob.
```
simitt: Generally it is not really clear to me from the description which Transactions, Spans and Errors are captured automatically for which Frameworks and Libraries, and which ones does a user need to manually collect.

You say 

> When installed ElasticAPM will automatically wrap your app's request/responses... 

I'd change this to `When installed and properly configured (link to how to set it up) ... `

docs/getting-started-rack.asciidoc
```diff
@@ -1,9 +1,26 @@
...
+server_url: http://localhost:8100
+secret_token: ''
+----
+
+Include the middleware, start (and stop) Elastic APM when booting your app:
+
 [source,ruby]
 ----
-# config.ru
+# Example config.ru
```
simitt: Could you also add a pure Rack example, as this section is generally called `Getting started with Rack`..

simitt:
I forgot one thing, could you add a sentence about the order of config options (ENV variables, config files, config options in code blocks)?

PR #127, opened 2018-05-29
Allow to set custom key filters

Bugagazavr:
This feature adds ability to set own custom_key_filters


mikker:
Hi @Bugagazavr! Thank you for this. This is definitely something we'll want 👍 
I'll take a closer look and ponder a bit about the API and return to this soon.

mikker:
In the meantime – can I make you sign our CLA at http://www.elasticsearch.org/contributor-agreement/?

Bugagazavr:
@mikker it's already signed, but don't know why status of check is not updated.

PR #131, opened 2018-06-01
Add JRuby 9.2 to test matrix

mikker:


mikker:
Failing due to this https://github.com/jruby/jruby/issues/5200.

mikker:
Not really sure what to do here. @simitt any suggestions? Only thing I see is write conditional tests around it. Not really a real solution.

simitt:
Looking at the test, what is wrong (logically) with having `actual_exception` as function name? It really is the function that causes the error. Could you explain why it would be better to have `/` here as function name?

mikker:
I don't want the _right_ answer. I just want consistency between the ruby runtimes/versions. This is JRuby 9.1 having one thing as the top-most line in the backtrace and JRuby 9.2 having another. I don't feel like I can do anything about that fact so what I'm wondering is how do I make our specs work around that. 😄

mikker:
jenkins, retest this please

mikker:
jenkins, retest this please

mikker:
jenkins, retest this please

PR #132, opened 2018-06-03
Don't fail if context tries reading too far back

reist:
Negative context start will cause Ruby to return nil for the range, after which APM just crashes.


mikker:
Wow, good catch! Thank you for this.

mikker:
Jenkins, please test this

mikker:
jenkins, please test this

mikker:
jenkins test this please

mikker:
💙💚💛💜❤️

PR #140, opened 2018-06-07
Remove newline on hostname

franee:
There's a newline due to the ruby backtick call on hostname.

### DIFF
--- a/spec/elastic_apm/system_info_spec.rb
+++ b/spec/elastic_apm/system_info_spec.rb
@@ -12,6 +12,12 @@ module ElasticAPM
           expect(subject[key]).to_not be_nil
         end
       end
+
+      context 'hostname' do
+        it 'has no newline at the end' do
+          expect(subject[:hostname]).not_to match(/\n\z/)
+        end
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


franee:
Yes

karmi:
Hi @franee, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/apm-agent-ruby/pull/140.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

franee:
@karmi done. Thanks!

spec/elastic_apm/system_info_spec.rb
```diff
@@ -12,6 +12,12 @@ module ElasticAPM
           expect(subject[key]).to_not be_nil
         end
       end
+
+      context 'hostname' do
+        it 'has no newline at the end' do
+          expect(subject[:hostname]).to eq(`hostname`.chomp)
```
mikker: Makes it a bit weird to have the same approach both places. Maybe `.to_not match(\n\z)` ?

franee: Yeah you’re right. Will push another one.

franee:
@mikker, I just pushed a new commit.

mikker:
Looks good!
jenkins test this please

mikker:
💙💚💛💜❤️

Issue #142, opened 2018-06-07
Allow filters to abort sending the entire payload


mikker:
My thinking is they'll do it by returning `nil`.

PR #148, opened 2018-06-08
Abort posting payload when filter returns nil

mikker:
Closes elastic/apm-agent-ruby#142

### DIFF
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -260,6 +260,8 @@ Arguments:
 
 Return the altered payload.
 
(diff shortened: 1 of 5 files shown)


docs/api.asciidoc
```diff
@@ -260,6 +260,8 @@ Arguments:
 
 Return the altered payload.
 
+If `nil` is returned all subsequent filters will be skipped and the psot request cancelled.
```
alvarolobato: Typo 'psot'

PR #150, opened 2018-06-12
Start mongodb for specs.

simitt:
If no mongodb server is running, start a docker
container before running the mongodb specs.

Locally `rspec spec` was failing when I haven't had started an external mongodb server. This attempts to fix the unit tests, by firing up a mongodb docker container.

### DIFF
--- a/spec/elastic_apm/spies/mongo_spec.rb
+++ b/spec/elastic_apm/spies/mongo_spec.rb
@@ -4,8 +4,18 @@ require 'mongo'
 
 module ElasticAPM
   RSpec.describe 'Spy: MongoDB' do
-    before do
+    around do |ex|
+      start_mongodb
+      ex.run
+      stop_mongodb
+    end
+
+    def stop_mongodb
       `docker-compose -f spec/docker-compose.yml down -v 2>&1`
+    end
+
+    def start_mongodb
+      stop_mongodb
       `docker-compose -f spec/docker-compose.yml up -d mongodb 2>&1`
     end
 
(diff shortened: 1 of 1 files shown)


mikker:
This is smart 👍🏻
\`\` outputs to `stdout` right? So it's visible to the user that this is happening?

simitt:
Yes it does output in case the docker container is started. I'd say that this is fine to tell the user, but I can look into suppressing it, if you prefer not to have it printed.

spec/elastic_apm/spies/mongo_spec.rb
```diff
@@ -4,14 +4,30 @@
 
 module ElasticAPM
   RSpec.describe 'Spy: MongoDB' do
+    around do |ex|
+      start_mongodb
+      ex.run
```
mikker: This is where I would wrap in `begin` ... `ensure` to make sure we shut it down whatever happens

spec/elastic_apm/spies/mongo_spec.rb
```diff
@@ -4,14 +4,30 @@
...
+      stop_mongodb
+    end
+
+    def stop_mongodb
+      `docker-compose -f spec/docker-compose.yml down -v 2>&1`
```
mikker: This is piping `stderr` to `stdout`, right? 👍🏻

spec/elastic_apm/spies/mongo_spec.rb
simitt: Since it is in an around hook the `stop_mongodb` will also be called if an exception happens, 
see [around-hook-with-exception](https://relishapp.com/rspec/rspec-core/docs/hooks/around-hooks#an-around-hook-continues-to-run-even-if-the-example-throws-an-exception)

PR #152, opened 2018-06-12
Update filter example to return payload.

simitt:


mikker:
Sorry for the merge confusion, thanks!

PR #170, opened 2018-07-06
Support multiple HTTP adapters and http-keepalive

cheald:
This PR refactors the HttpAdapter used in `ElasticAPM::Http` into an abstract base class, which is then implemented by subclasses. I refactored the call sites a bit so that the adapter-specific logic is encapsulated in each adapter.

This was done primarily to add support for HTTP keepalives. Net::HTTP doesn't provide them, and the cost of constantly opening connections to the APM server is not insignificant. I wanted keepalives to the APM server, and here we are.

It adds a soft dependency on the `http` gem - if it's present, the HttpRbAdapter is loaded, and is available to be used via the configuration object (`config.http_adapter = :HttpRbAdapter`). If it's not present, then it's not loaded. I don't love this pattern, but it's a low-touch way to get what we want. A better solution might be to rewrite the Http class to use Faraday, and then to permit the user to pass in a Faraday configuration using the client of their choice.

The suite passes and I've done basic testing in development mode - stuff gets through to the APM server - but I haven't banged on it super hard yet.

### DIFF
--- a/lib/elastic_apm/http_adapters/http_rb_adapter.rb
+++ b/lib/elastic_apm/http_adapters/http_rb_adapter.rb
@@ -1,40 +1,42 @@
 require 'elastic_apm/http_adapters/abstract_adapter'
+enable = false
 begin
   require 'http'
+  enable = true
+rescue LoadError
+end
 
-  module ElasticAPM
-    module HttpAdapters
-      # @api private
-      class HttpRbAdapter < AbstractHttpAdapter
-        def perform(uri, data, headers)
-          return DISABLED if @config.disable_send?
+module ElasticAPM
+  module HttpAdapters
+    # @api private
+    class HttpRbAdapter < AbstractHttpAdapter
+      def perform(uri, data, headers)
+        return DISABLED if @config.disable_send?
 
-          req = client(uri).headers(headers)
-          if @config.http_compression &&
-            data.bytesize > @config.compression_minimum_size
-            req = req.use(auto_deflate: { method: :deflate })
-          end
-          Response.new req.post(uri, body: data, ssl_context: get_context)
+        req = client(uri).headers(headers)
+        if @config.http_compression &&
+          data.bytesize > @config.compression_minimum_size
+          req = req.use(auto_deflate: { method: :deflate })
         end
+        Response.new req.post(uri, body: data, ssl_context: get_context)
+      end
 
-        private
+      private
 
-        def client(uri)
-          @client ||= begin
-            keepalive = @config.http_adapter_options[:keepalive]
-            keepalive ? ::HTTP.persistent(uri) : ::HTTP
-          end
+      def client(uri)
+        @client ||= begin
+          keepalive = @config.http_adapter_options[:keepalive]
+          keepalive ? ::HTTP.persistent(uri) : ::HTTP
         end
+      end
 
-        def get_context
-          ctx = OpenSSL::SSL::SSLContext.new
-          if @config.use_ssl? && @config.verify_server_cert?
-            ctx.verify_mode = OpenSSL::SSL::VERIFY_NONE
-          end
-          ctx
+      def get_context
+        ctx = OpenSSL::SSL::SSLContext.new
+        if @config.use_ssl? && @config.verify_server_cert?
+          ctx.verify_mode = OpenSSL::SSL::VERIFY_NONE
         end
+        ctx
       end
     end
   end
-rescue LoadError
-end
+end if enable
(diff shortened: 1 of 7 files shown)


mikker:
jenkins test this please

cheald:
I've tried it in my primary project in dev mode; I haven't rolled it out to a production environment yet. I can do that in the next day or two if you want to wait on a real-world test to do the merge.

mikker:
Not necessary to do production unless you're curious yourself 😄

1. I think I'd like an API that moves the keep-alive option to the adapter itself. Otherwise we'll end up with all options for all adapters in the same config. This is hypothetical of course but might be interesting to see if we can come up with something.
1. Why do you prefer `require_relative`? I usually stick to `require` so I don't end up requiring anything twice. But I'm curious if I'm missing something?

cheald:
1. I'm not sure I understand; are you thinking something like:

```ruby
config.http_adapter_options = {
  keepalive: true
}
```
which the adapter then optionally consumes? That would let us just pass an arbitrary hash of options at setup time.

2. `require_relative` is just a habit I've picked up - it's generally most useful in the context of one-off tests or places where the load path isn't otherwise properly set up. I don't think it makes a difference in this case - `require` should be fine.

mikker:
Yes, probably something like that. Were kind of restricted by the Yaml file to do it in a way that Yaml supports.

We could also restrict this to an option you need to set in Ruby eg,

```ruby
config.elastic_apm.http_adapter = ElasticAPM::Http::HttpRbAdapter.new do |a|
  a.keep_alive = true
end
```

But that's another paradigm than with the existing `Config` object. We should probably use the same in both.

cheald:
I think I'd suggest that we make a best-effort attempt to do it via YAML, but it's not a bad idea to document how to set up the adapter manually, which gives adapters the ability to do more complex setup. I'll push a proposal shortly.

lib/elastic_apm/http_adapters/http_rb_adapter.rb
```diff
@@ -1,37 +1,40 @@
-require_relative './abstract_adapter'
+require 'elastic_apm/http_adapters/abstract_adapter'
 begin
```
mikker: I like this approach: https://github.com/elastic/apm-agent-ruby/blob/695ff27a9a12c69e22b3c89af0a2ba055912a709/spec/elastic_apm/subscriber_spec.rb#L3-L11

lib/elastic_apm/http_adapters/http_rb_adapter.rb
```diff
@@ -1,37 +1,40 @@
...
+            ctx.verify_mode = OpenSSL::SSL::VERIFY_NONE
+          end
+          ctx
         end
-        ctx
       end
     end
   end
-end if Kernel.const_defined?(:HTTP) && ::HTTP.const_defined?(:Chainable)
+rescue LoadError
```
mikker: (contd.) Also puts this close to the `begin` so we see why we're doing the whole dance in the first place.

lib/elastic_apm/http_adapters/http_rb_adapter.rb
cheald: Yeah, that's fair. I'd skip the `puts` for this, since we don't want to see a message on every startup if we don't want to use http.rb.

mikker:
jenkins test this please

mikker:
jenkins test this please

mikker:
Hi @cheald! I'm on parental leave the next 4 weeks. Didn't want to leave you hanging – this is good work and I'm looking forward to getting it into the project when I'm back. Talk to you then! 👋

cheald:
Congratulations, and enjoy your time with your family! It'll be here when you get back.

mikker:
Hi again @cheald – I'm back!

I'm just about to start working on supporting v2 of our intake API which is going to be using persistent, long-running connections by default. Without committing to anything yet, I'm thinking that I'll use http.rb for it. That'll add it as a hard dependency which of course puts this PR in sort of an awkward spot.

I'm not sure I want to add the whole concept of optional HTTP adapters only to deprecate it so soon after. I hope you understand. I'm very happy for your work on this though 👍

cheald:
Even better! As long as we get persistent connections, I'm not picky about how we get there. Http.rb is great!

Welcome back!

mikker:
`v2.0` has a hard dependency on Http.rb and brings persistent connections and chunked requests.

Bringing this change into v1 when it's already on the way to be overrun seems unnecessary.

Thanks very much for the proposal and initiative @cheald!

mikker:
If you feel like it, comments and criticisms on the `v2.0` branch or current PR (#189) are very welcome!

PR #174, opened 2018-07-25
Add rake instrumenter

steve21168:


mikker:
Hi @steve21168 ! Thanks for this. What's your use case exactly? I suppose you are running some application specific rake tasks in production? Maybe through cron or something like it?

Your implementation is pretty straight forward 👍 I worry though that Rake is used for many things, so adding this to _all_ tasks might mean it shows up in tasks that aren't really relevant for APM.

Could we make it opt in for the individual task instead? Maybe use something like Rake's dependent tasks eg with `Rake::Task#enhance` (https://ruby-doc.org/stdlib-2.0.0/libdoc/rake/rdoc/Rake/Task.html#method-i-enhance)

```ruby
namespace :app do
  task :do_the_thing do
    # We need to provide task name ourselves
    if (transaction = ElasticAPM.current_transaction)
      transaction.name = 'do_the_thing'
    end
  end
end

namespace :elastic_apm do
  task :start do
    # start the agent and a new transaction
    ElasticAPM.start
    ElasticAPM.transaction
  end
end

Rake::Task['app:do_the_thing'].enhance(['elastic_apm:start']) do
  # end transaction and stop the agent
  if (transaction = ElaticAPM.current_transaction)
    transaction.submit 'success'
  end

  ElasticAPM.stop
end
```

This is of course not much different than adding all of it manually `¯\_(ツ)_/¯`

steve21168:
@mikker You're correct, we run scheduled tasks via Rake.  So it's useful to gauge the timing of each transaction, but even more importantly I want the APM to track and report any errors.  

I agree that Rake gets used for many things, some of which may not be desirable to report.  That's the reason I added it to disabled spies by default.

I think running scheduled tasks via Rake is pretty common in the Ruby community.  If you look at error reporting services like Sentry, Airbrake, or a comparable APM (NewRelic's APM) they all support tracking of Rake tasks.  Point being I think there should be some kind of support for this within the library.

Per your solution, I like it.  To be clear the "In Lib" portion would reside in this APMs code?  I think that would be fine but can we have the APM name the transaction itself based on the config.instrument_rake_tasks name that we pass in? That way in the application code you would only need to add a single line like:

ElasticAPM.agent.config.instrument_rake_tasks = ['app:do_the_thing']

To add instrumentation to a task.

mikker:
You’re completely right - totally missed that we *did* have the task name anyway.

This wouldn’t catch errors though. Any idea how we could do that with this approach? I’ll take a look tomorrow.

steve21168:
Hmm, yeah that's a problem.  Maybe we stick with monkey patching rake but come up with a different way of enabling/disabling? I agree that this is something that should mostly be off.  

Having it strictly enabled by an environment variable would be suitable. For example when we run our scheduled jobs we could just set that environment variable on the run command: 

`
ELASTIC_APM_RAKE=true rake app:do_thing
`

mikker:
I like the ENV approach. Makes it easy to turn it on for both a single run, a single task or for everything.

So, that and patching `Rake::Task` is probably the way to go. Do you want to take a go at it in this PR? _No_ is a perfectly fine answer 😄 although I suppose I'll have to do my job myself then.

steve21168:
@mikker Happy to try and help.  I just pushed a commit.  This change allows for additional "electable_spies" and will still be overridden by something explicitly listed under disabled_spies.  Let me know what you think.

steve21168:
@mikker Yeah so I understand having the ElasticAPM.agent.config.enable_rake_task_instrumentation method. Which I would think should default to false right?  

Where are you suggesting the perform method should live?

steve21168:
@mikker Hey I'd love to get this in the APM.  I've added the environment option for enable_rake_task_instrumentation.  Still unsure where the best place for the control of enabling/disabling the spy should be. 

The spy has no access to the config so it can't make that decision on the install.

PR #178, opened 2018-07-30
Make Lru Cache threadsafe

simitt:
* Synchronizes the calls to the LRU Cache. 
* Simplifies the call for reading a value. The block is passed into `Hash.new(&block)`. According to the [ruby-docs](https://ruby-doc.org/core-2.2.0/Hash.html#method-c-new) the block used to init a Hash has to ensure that it modifies the hash accordingly: 

  > If a block is specified, it will be called with the hash object and the key, and should return the default value. It is the block's responsibility to store the value in the hash if required.

  Thus, if the same spec is applied to `LRU Cache`, then the block doesn't need to be called explicitly, as it is triggered by fetching the value from the hash, if the value is not set.

### DIFF
--- a/lib/elastic_apm/util/lru_cache.rb
+++ b/lib/elastic_apm/util/lru_cache.rb
@@ -7,27 +7,22 @@ module ElasticAPM
       def initialize(max_size = 512, &block)
         @max_size = max_size
         @data = Hash.new(&block)
-        @missing_key_block = block
+        @mutex = Mutex.new
       end
 
       def [](key)
-        found = true
-        value = @data.delete(key) { found = false }
-
-        if found
-          @data[key] = value
-        else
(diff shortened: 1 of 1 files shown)


simitt:
jenkins, retest this please

simitt:
jenkins, retest this please

simitt:
jenkins, retest this please

simitt:
jenkins, retest this please

simitt:
jenkins, retest this please.

simitt:
@mikker I do see quite some variation for jruby-9.1 (listed as ruby-2.3.3), but also between runs in `master`. The benchmarks for the PR never got significantly worse than for master.
All the other ruby versions are actually about the same. 

I let this fix run locally for a sidekiq instance with 3 workers for some time now and didn't see any issues for ruby 2.3.

Issue #164, opened 2018-06-28
Configurable default context & tags for every transaction


brandonwestcott:
Hello, our team has recently setup apm and we are really impressed. @mikker thanks for making the ruby agent a reality!

Wondering if there has been any discussion on setting default context and tags that would be added onto every transaction. In our typical usage of other agent beats, we end up utilizing that "fields" configuration option in order to add global metadata. For example, one common use case for us is to always add the "team/department" shipping the log because we shard our ELK indices by that field for retention / security.

Our first attempt to solve this was to just set those things inside rack middleware, but then once we added in sidekiq, we of course realized that wasn't sufficient for other spies.

Is this something you would be open to adding to the feature set? If so and you are short on time, I would be happy to submit a PR for review.

brandonwestcott:
Yes, these are distinct teams that are "managing" the different apps. And thats just one of many examples too, others include propagating client name, blue/green identifier, and even things like aws_az, etc. 

Something inside config would be perfect. In fact, the middleware I attempted above was actually driven from an `config/elastic_apm_custom.yml`.

mikker:
Kept it to tags in the PR. Tags differ from `custom` in that they are indexed by ES. To your use case I think this makes the most sense. Also lets the ENV-version be simpler in that tags are simple `string=string` pairs.

Hope it's sufficient? 😄

brandonwestcott:
This is great, thank you!

lrpauley:
was support for this feature [removed here](https://github.com/elastic/apm-agent-ruby/commit/13e2a024c3adbd330a2ce3f7685d3d3e5a39e055#diff-3ff0f7dfc745f9ac0268a337d9caf3e0), or maybe it was re-implemented another way?

mikker:
It seems to have been removed by mistake. Total blunder, sorry. I'll re-add it.

lrpauley:
no worries, thanks for getting back to me so quickly!

PR #183, opened 2018-08-10
Add default_tags option

mikker:
Allows the user to add config-level tags that'll be included in every transaction.

See elastic/apm-agent-ruby#164 

Also includes some spec suite clean-up.

### DIFF
--- a/bench/report.rb
+++ b/bench/report.rb
@@ -50,7 +50,7 @@ payloads = titles.zip(averages, counts).map do |(title, avg, count)|
     transaction_count: count,
     executed_at: Time.new.iso8601,
     'git.commit' => git_sha,
-    'git.date' => git_date && Time.parse(git_date).iso8601,
+    'git.date' => String(git_date).strip != '' && Time.parse(git_date).iso8601,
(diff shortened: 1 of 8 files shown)


bench/report.rb
```diff
@@ -50,7 +50,7 @@
     transaction_count: count,
     executed_at: Time.new.iso8601,
     'git.commit' => git_sha,
-    'git.date' => git_date && Time.parse(git_date).iso8601,
+    'git.date' => String(git_date).strip != '' && Time.parse(git_date).iso8601,
```
simitt: Could you move this to a separate commit/PR please?

mikker: Too late 😬

lib/elastic_apm/transaction.rb
```diff
@@ -29,6 +29,7 @@ def initialize(
       @notifications = [] # for AS::Notifications
 
       @context = context || Context.new
+      @context.tags.merge!(instrumenter.config.default_tags)
```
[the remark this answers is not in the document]
mikker: It probably should, yes 👍

PR #186, opened 2018-08-13
check for Rails::Application instead of just ::Rails

loglesby:
Fix flawed check that would assume Rails::Application is defined if ::Rails is, which is not always the case. In a Sinatra app which uses some Rails libraries, this results in failure to boot.


mikker:
Hi @loglesby! Nice catch 👍 thank you.

mikker:
jenkins, test this please

loglesby:
Yep, all done, no worries! :)

PR #187, opened 2018-08-17
🚧 Support Intake v2 (very early, still WIP)

mikker:
Stil todo:

- [x] Cycle connection when max time reached
- [ ] Cycle connection when max bytes
- [ ] Add `metadata`
- [ ] gzip
- [ ] Handle errors?
- [ ] …


mikker:
Closing, as this is somehow based on the wrong branch

PR #192, opened 2018-08-31
Add rake task instrumentation

mikker:
Continued from #174.

Brings a new setting `instrument_rake` that defaults to `false`.

Some people use rake for running common tasks in production. In this case APM could help a lot with debugging and optimizing. But rake is used for a lot of other things as well, so we don't want to just go ahead and enable this on every task everytime.

This way a user can choose to

- completely omit the rake spy (`disabled_spies += ['rake']`)
- load the spy but don't instrument (default)
- load the spy and instrument everywhere (`instrumented_rake_tasks: ['my_task']` in config file)
- load everywhere but only instrument individual tasks (`$ ELASTIC_APM_INSTRUMENTED_RAKE_TASKS=some_task rake some_task`)

---

- [x] Spec
- [x] Docs

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -390,6 +390,18 @@ Use this option to disable any of these.
 Get an array of enabled spies with `ElasticAPM.agent.config.enabled_spies`.
 
 [float]
+[[config-instrumented-rake-tasks]]
+==== `instrumented_rake_tasks`
+
+[options="header"]
+|============
+| Environment                           | `Config` key              | Default | Example
+| `ELASTIC_APM_INSTRUMENTED_RAKE_TASKS` | `instrumented_rake_tasks` | `[]`    | `['my_task']`
+|============
+
+Elastic APM can instrument your Rake tasks but given that they are used for a multitude of sometimes very different and not always relevant things, this is opt in.
+
+[float]
 [[config-default-tags]]
 ==== `default_tags`
 
--- a/lib/elastic_apm/config.rb
+++ b/lib/elastic_apm/config.rb
@@ -42,6 +42,7 @@ module ElasticAPM
       span_frames_min_duration: 5,
 
       disabled_spies: %w[json],
+      instrument_rake: false,
 
       default_tags: {},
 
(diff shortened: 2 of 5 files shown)


lib/elastic_apm/config.rb
```diff
@@ -172,8 +175,9 @@ def initialize(options = {})
     attr_reader   :ignore_url_patterns
 
     alias :disable_environment_warning? :disable_environment_warning
-    alias :verify_server_cert? :verify_server_cert
     alias :disable_send? :disable_send
+    alias :instrument_rake? :instrument_rake
+    alias :verify_server_cert? :verify_server_cert
```
simitt: What does `verify_server_cert` have to do with the rake spy?

mikker: Just noticed it was missing an alias with `?` at the end

mikker:
The reason for including by default but not enabling is that it's somewhat cumbersome to enable spies for single tasks. Eg you'd have to `$ ELASTIC_APM_DISABLED_SPIES=json rake thing` because `json` is the only other spy disabled by default. That seems very indirect.

mikker:
I kind of like turning the option into an array of enabled task names but I don't think there's much difference between that and this?

simitt:
The difference would be in how you can use it. If I understand you correct then atm you cannot use the configuration file for defining which rake tasks you want to be monitored, but you need to call the specific rake task with the ENV variable that apm spy is enabled. 
By using this rake task name array, you could add it to your configuration and you're all set, but you could also pass it along as ENV var.

mikker:
jenkins, retest this please

mikker:
@steve21168 Still need you to sign the CLA. Can't merge your commits without it 🙃

steve21168:
@mikker Should be signed now.

docs/configuration.asciidoc
```diff
@@ -389,6 +389,18 @@ Use this option to disable any of these.
...
 Get an array of enabled spies with `ElasticAPM.agent.config.enabled_spies`.
 
+[float]
+[[config-instrumented-rake-tasks]]
+==== `instrumented_rake_tasks`
+
+[options="header"]
+|============
+| Environment                           | `Config` key              | Default
+| `ELASTIC_APM_INSTRUMENTED_RAKE_TASKS` | `instrumented_rake_tasks` | `['app:custom:task', 'other_task']`
```
simitt: Isn't the default value an empty array?

mikker: You are right! I think I confused it with an example column.

PR #209, opened 2018-09-14
Add new instrumentation APIs (WIP)

mikker:
- Move responsibility of submitting from Transactions themselves to the Instrumenter.
- Decouples spans from Transactions to prepare for Distributed Tracing etc.
- Streamlines API with other agents (`start_transaction`/`end_transaction`)
- Adds deprecation helper module to warn about method removals

---

- [x] Implementation
- [x] Specs
- [x] Deprecate (but keep old apis until major bump)
- [x] ~Docs~ pushing docs until DT is implemented


lib/elastic_apm.rb
```diff
@@ -16,142 +17,237 @@
...
+    end
+
+    ### Errors
+
+    # Report and exception to APM
+    #
+    # @param exception [Exception] The exception
+    # @param handled [Boolean] Whether the exception was rescued
+    # @return [Error] The generated [Error]
+    def report(exception, handled: true)
```
simitt: Did you think about also aligning this with `span` and `transaction`, e.g. use `error`, and `error_message`?

mikker: I probably should, yeah. Even though I like my names better 😀

PR #213, opened 2018-09-25
🔀 Support Distributed Tracing

### LINKED ISSUE #193: 🔀 Distributed Tracing
- [ ] Intake v2
- [ ] [test that responses from the server during chunked writes](https://github.com/elastic/apm-server/issues/1298#issuecomment-415704054) work with HTTP.rb
- [ ] `traceparent` header extract/inject functionality
- [ ] update opbeans-ruby to include a distributed tracing call
- [ ] `transaction.span_count.total` (https://github.com/elastic/apm-server/issues/1241)
- [ ] auto-instrument http module(s) to use extact/inject functionality

mikker:
Closes elastic/apm-agent-ruby#193

### DIFF
--- a/lib/elastic_apm/error_builder.rb
+++ b/lib/elastic_apm/error_builder.rb
@@ -7,6 +7,7 @@ module ElasticAPM
       @agent = agent
     end
 
+    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
     def build_exception(exception, handled: true)
       error = Error.new
       error.exception = Error::Exception.new(exception, handled: handled)
@@ -19,10 +20,14 @@ module ElasticAPM
 
       if (transaction = ElasticAPM.current_transaction)
         error.context = transaction.context.dup
+        error.trace_id = transaction.trace_id
+        error.transaction_id = transaction.id
+        error.parent_id = ElasticAPM.current_span&.id || transaction.id
       end
 
       error
     end
+    # rubocop:enable Metrics/MethodLength, Metrics/AbcSize
 
     def build_log(message, backtrace: nil, **attrs)
       error = Error.new
--- a/lib/elastic_apm/middleware.rb
+++ b/lib/elastic_apm/middleware.rb
@@ -1,8 +1,13 @@
+#
 # frozen_string_literal: true
+
+require 'elastic_apm/traceparent'
 
 module ElasticAPM
   # @api private
   class Middleware
+    include Logging
+
     def initialize(app)
       @app = app
     end
@@ -47,7 +52,16 @@ module ElasticAPM
 
     def start_transaction(env)
       ElasticAPM.start_transaction 'Rack', 'request',
-        context: ElasticAPM.build_context(env)
+        context: ElasticAPM.build_context(env),
+        traceparent: traceparent(env)
+    end
+
+    def traceparent(env)
+      return unless (header = env['HTTP_ELASTIC_APM_TRACEPARENT'])
+      Traceparent.parse(header)
+    rescue Traceparent::InvalidTraceparentHeader
+      warn "Couldn't parse invalid traceparent header: #{header.inspect}"
+      nil
     end
 
     def running?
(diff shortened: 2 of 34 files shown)


mikker:
Another big one for when you have time, @simitt 😊

lib/elastic_apm/error_builder.rb
```diff
@@ -19,10 +20,13 @@ def build_exception(exception, handled: true)
 
       if (transaction = ElasticAPM.current_transaction)
         error.context = transaction.context.dup
+        error.trace_id = transaction.trace_id
+        error.parent_id = transaction.id
```
simitt: The parent of an `error` could also be a `span`.  There's a dedicated `transaction_id` defined for an error, that should be set if available. See [json spec](https://github.com/elastic/apm-server/blob/v2/docs/spec/errors/v2_error.json#L21).

lib/elastic_apm/middleware.rb
```diff
@@ -47,7 +50,15 @@ def path_ignored?(env)
...
       ElasticAPM.start_transaction 'Rack', 'request',
-        context: ElasticAPM.build_context(env)
+        context: ElasticAPM.build_context(env),
+        traceparent: traceparent(env)
+    end
+
+    def traceparent(env)
+      return unless (header = env['HTTP_ELASTIC_APM_TRACEPARENT'])
+      Traceparent.parse(header)
+    rescue Traceparent::InvalidTraceparentHeader
```
simitt: Should you maybe log something here?

lib/elastic_apm/span.rb
```diff
@@ -9,67 +9,75 @@ module ElasticAPM
...
-      @parent = parent
-      @context = context
 
       @id = SecureRandom.hex(8)
-      @trace_id = transaction.trace_id
-      @transaction_id = transaction.id
 
-      @timestamp = @transaction.timestamp
+      self.transaction = transaction
+      self.parent = parent
```
simitt: From a consistency point of view - why using `self` here over `@`?

lib/elastic_apm/traceparent.rb
```diff
@@ -0,0 +1,65 @@
...
+    HEX_REGEX = /[^[:xdigit:]]/
+
+    def initialize
+      @version = VERSION
+    end
+
+    def self.from_transaction(transaction)
+      new.tap do |t|
+        t.trace_id = SecureRandom.hex(16)
+        t.span_id = transaction.id
```
simitt: Why do you set the `span_id` to `transaction.id`? From what I understand it should be set to the `span.id` that initiated the outgoing call to another service.

lib/elastic_apm/util.rb
```diff
@@ -15,5 +15,13 @@ def self.git_sha
       sha = `git rev-parse --verify HEAD 2>&1`.chomp
       $? && $?.success? ? sha : nil # rubocop:disable Style/SpecialGlobalVars
     end
```
simitt: I suggest to add the `.utc` inside the `def self.micros` method, to ensure all timestamps are used in utc.

spec/elastic_apm/middleware_spec.rb
```diff
@@ -47,5 +47,77 @@ class MiddlewareTestError < StandardError; end
...
+
+      context 'with a blank header' do
+        it 'skips traceparent, makes new', :intercept do
+          app.call(
+            Rack::MockRequest.env_for('/', 'HTTP_ELASTIC_APM_TRACEPARENT' => '')
+          )
+
+          traceparent = @intercepted.transactions.first.traceparent
+          expect(traceparent.trace_id)
+            .to_not eq '0af7651916cd43dd8448eb211c80319c'
```
simitt: You could also check here that `traceparent.span_id` is empty.

lib/elastic_apm/span.rb
simitt: There is no custom setter for `parent`.

PR #223, opened 2018-10-23
Add Transport::Worker

mikker:
This adds a worker running in a separate thread that has 2 responsibilites:

1. Serialize objects off the main thread
2. Serve the resulting payloads to the Connection

**NB**: Haven't tested this in a real app yet. Doing that now.

---

- [x] Re-add filters
- [x] Pull synchronize as much as possible from `Connection`
- [ ] Docs

### DIFF
--- a/lib/elastic_apm/agent.rb
+++ b/lib/elastic_apm/agent.rb
@@ -60,22 +60,23 @@ module ElasticAPM
 
     def initialize(config)
       @config = config
+
       @transport = Transport::Base.new(config)
+      @instrumenter = Instrumenter.new(config) { |event| enqueue event }
 
-      @instrumenter = Instrumenter.new(self)
-
+      @stacktrace_builder = StacktraceBuilder.new(config)
       @context_builder = ContextBuilder.new(self)
       @error_builder = ErrorBuilder.new(self)
-      @stacktrace_builder = StacktraceBuilder.new(config)
     end
 
-    attr_reader :config, :transport, :messages, :pending_transactions,
-      :instrumenter, :context_builder, :stacktrace_builder, :error_builder
+    attr_reader :config, :transport, :instrumenter,
+      :stacktrace_builder, :context_builder, :error_builder
 
     def start
       debug '[%s] Starting agent, reporting to %s', VERSION, config.server_url
 
-      @instrumenter.start
+      transport.start
+      instrumenter.start
 
       config.enabled_spies.each do |lib|
         require "elastic_apm/spies/#{lib}"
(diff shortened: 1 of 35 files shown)


lib/elastic_apm/agent.rb
```diff
@@ -186,7 +188,7 @@ def set_user(user)
     end
 
     def add_filter(key, callback)
-      @transport.filters.add(key, callback)
+      transport.worker.filters.add(key, callback)
```
mikker: Not really a fan of this. Not thread safe either. A user could theoretically add a new filter while the agent is applying the existing filters somewhere else. Probably not in _real life projects_ but in theory it isn't safe as is.

lib/elastic_apm/transport/base.rb
```diff
@@ -1,60 +1,52 @@
...
+        @worker.stop 5
       end
 
-      def flush
-        @connection.close!
+      def ensure_worker
+        return if @worker.alive?
+
+        debug 'Starting worker'
+        @worker.work_forever
```
simitt: This doesn't look thread safe and could start multiple workers if called by more than one thread.

mikker:
The `Agent` itself is a singleton and almost everything else is a child of it – so in a sense everything is already a singleton 😄 

But you're right – I usually keep away from singletons as long as there's a way around one. Regular old, boring, ruby classes and objects are just easier to test and reason about.

lib/elastic_apm/transport/base.rb
```diff
@@ -13,40 +13,53 @@ class Base
...
+        end
       end
 
-      attr_reader :config, :filters, :connection, :worker
+      attr_reader :config, :filters, :queue, :messages
 
       def submit(resource)
-        ensure_worker
-        worker.push(resource)
+        queue.push(resource)
```
simitt: If the queue is full this is blocking until space is available, which means the thread could potentially block for a really long time if the server connection is dead. WDYT about using the non-blocking push and catching the exception that is caught?

lib/elastic_apm/agent.rb
```diff
@@ -188,7 +188,7 @@ def set_user(user)
     end
 
     def add_filter(key, callback)
-      transport.worker.filters.add(key, callback)
+      # transport.worker.filters.add(key, callback)
```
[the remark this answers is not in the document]
simitt: Could you hold a `filters` object in `Transport::Base` and pass that into every worker?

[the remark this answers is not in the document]
mikker: I suppose I could. As long as `add_filter` is called before the workers boot. As soon as they are going, there's no looking back 😁

mikker:
jenkins, retest this please

mikker:
jenkins, retest this please

mikker:
jenkins, retest this please

mikker:
jenkins, retest this please

mikker:
jenkins, retest this please

PR #227, opened 2018-11-12
Loosen concurrent-ruby version dependency

ferrous26:
Since the release of concurrent-ruby 1.1.x last week, bundler has been insisting on downgrading elastic-apm to 0.7.4 in a rails project I am working. 

Bundler seems to prefer a newer concurrent-ruby to a newer elastic-apm unless I force an elastic-apm version constraint. Loosening elastic-apm's constraint to 1.x should fix this and allow the latest of both gems.

Since elastic-apm only seems to have a very light dependency on concurrent-ruby for the timer task, I think loosening the requirement should be safe. Alternatively, the version constraint could be updated to 1.1.x.

### DIFF
--- a/elastic-apm.gemspec
+++ b/elastic-apm.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |spec|
     f.match(%r{^(test|spec|features)/})
   end
 
-  spec.add_dependency('concurrent-ruby', '~> 1.0.0')
+  spec.add_dependency('concurrent-ruby', '~> 1.0')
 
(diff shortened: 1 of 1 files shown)


mikker:
Thank you for reporting and fixing this! Happy to merge and release a new version. Can I make you sign our CLA? http://www.elasticsearch.org/contributor-agreement/

mikker:
Manually changed this in the recent 2.0 release. Sorry, but I need the CLA signed to be able to merge PRs. Thanks for bringing this to my attention!

Issue #237, opened 2018-11-19
Worker thread dies with exception in 2.0.1


OpakAlex:
[ElasticAPM] Queue is full (256 items), skipping…
What this means? Why it happens? 

0 new records into index since update version.

---

Update by @mikker: See [this comment](https://github.com/elastic/apm-agent-ruby/issues/237#issuecomment-481738046) for possible settings to tweak. I'm working on a docs page describing possible remedies to full queues.

mikker:
Hi @OpakAlex!

It means the agent has trouble sending its recorded transactions to APM Server.

Your logs will

OpakAlex:
@mikker trying to find something, only in production mode.
Also i have some warnings:
```
[8] ! WARNING: Detected 2 Thread(s) started in app boot:
[8] ! #&lt;Thread:0x000055787bfba218@/bundle/ruby/2.5.0/gems/activerecord-5.2.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:297 sleep&gt; - /bundle/ruby/2.5.0/gems/activerecord-5.2.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:299:in `sleep'
[8] ! #&lt;Thread:0x000055787c139cd8@/bundle/ruby/2.5.0/gems/concurrent-ruby-1.0.5/lib/concurrent/executor/ruby_thread_pool_executor.rb:317 sleep_forever&gt; - /bundle/ruby/2.5.0/gems/elastic-apm-2.0.1/lib/elastic_apm/transport/worker.rb:44:in `pop'
```

OpakAlex:
@mikker do i need on apm server turn on `v2`?

mikker:
Yes, version 2+ of the agent requires APM Server to be at least version 6.5.

Looking into the warnings.

OpakAlex:
@mikker i want to understand this.

mikker:
Yes, APM Server 6.5 should work with ES all the way back to 6.2, see https://www.elastic.co/support/matrix#matrix_compatibility under _Product Compatibility_

roncohen:
@OpakAlex please be aware that even though APM Server 6.5 can work with Elasticsearch 6.2, you will need to use Kibana 6.5 to visualize trace data from APM Server 6.5

OpakAlex:
So, APM can write correct to ES 6.2, but if i need to see this results correct, i need kibana 6.5, right?
[...]

PR #239, opened 2018-11-19
Lazy boot transport workers

mikker:
Puma with `preload_app!` enabled will boot workers before booting its threads leaving the the threads out of reach of each other.

Fixes elastic/apm-agent-ruby#237


mikker:
jenkines, retest this please

PR #240, opened 2018-11-19
Truncate keyword fields to maximum 1024 chars

### LINKED ISSUE #224: Problem validating JSON document against schema
I have 2 rails application with apm gem, and i have this issue:
```
RROR	[request]	beater/handlers.go:384	error handling request	{"request_id": "fbcc0e8f-e8db-43ac-b0ff-802803c689b2", "error": "Problem validating JSON document against schema: I[#] S[#] doesn't validate with \"transaction#\"\n  I[#/transactions/1/context/request/url/full] S[#/properties/transactions/items/properties/context/properties/request/properties/url/properties/full/maxLength] length must be &lt;= 1024, but got 1463"}
ERROR	[request]	beater/handlers.go:384	error handling request	{"request_id": "cdd5600e-11f4-4854-a15a-c841b7d78283", "error": "Problem validating JSON document against schema: I[#] S[#] doesn't validate with \"transaction#\"\n  I[#/transactions/1/context/request/url/search] S[#/properties/transactions/items/properties/context/properties/request/properties/url/properties/search/maxLength] length must be &lt;= 1024, but got 1447"}
```

APM -> 6.2.4

mikker:
Fixes elastic/apm-agent-ruby#224

### DIFF
--- a/lib/elastic_apm/transport/serializers/span_serializer.rb
+++ b/lib/elastic_apm/transport/serializers/span_serializer.rb
@@ -5,24 +5,68 @@ module ElasticAPM
     module Serializers
       # @api private
       class SpanSerializer < Serializer
-        # rubocop:disable Metrics/MethodLength
+        def initialize(config)
+          super
+
+          @context_serializer = ContextSerializer.new(config)
+        end
+
+        attr_reader :context_serializer
+
+        # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
         def build(span)
           {
             span: {
               id: span.id,
               transaction_id: span.transaction_id,
               parent_id: span.parent_id,
-              name: span.name,
-              type: span.type,
+              name: keyword_field(span.name),
+              type: keyword_field(span.type),
               duration: ms(span.duration),
-              context: span.context&.to_h,
+              context: context_serializer.build(span.context),
               stacktrace: span.stacktrace.to_a,
               timestamp: span.timestamp,
               trace_id: span.trace_id
             }
           }
         end
-        # rubocop:enable Metrics/MethodLength
+        # rubocop:enable Metrics/MethodLength, Metrics/AbcSize
+
+        # @api private
+        class ContextSerializer < Serializer
+          def build(context)
+            return unless context
+
+            {
+              sync: context.sync,
+              db: build_db(context.db),
+              http: build_http(context.http)
+            }
+          end
+
+          private
+
+          def build_db(db)
+            return unless db
+
+            {
+              instance: db.instance,
+              statement: db.statement,
+              type: db.type,
+              user: db.user
+            }
+          end
+
+          def build_http(http)
+            return unless http
+
+            {
+              url: http.url,
+              status_code: http.status_code,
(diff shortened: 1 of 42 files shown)


lib/elastic_apm/util.rb
```diff
@@ -24,5 +24,12 @@ def self.hex_to_bits(str)
     def self.reverse_merge!(first, second)
       first.merge!(second) { |_, old, _| old }
     end
+
+    def self.truncate(value)
+      return unless value
+      return value if value.length <= 1024
+
+      format('%s…', value[0...1024])
```
simitt: This won't work as you send 1024 + 3 chars for the dots, so the server will reject that. 

`format` creates a new copy here which I don't think is necessary. You could return `value[0,len]` or if you are keen on adding the dots something like `TRUNCATED_CONST="..."; value[0,len] << TRUNCATED_CONST`

lib/elastic_apm/util.rb
```diff
@@ -24,5 +24,12 @@ def self.hex_to_bits(str)
     def self.reverse_merge!(first, second)
       first.merge!(second) { |_, old, _| old }
     end
+
+    def self.truncate(value)
```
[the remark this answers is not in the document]
mikker: 👍

lib/elastic_apm/util.rb
mikker: That’s a one-char utf ellipsis and a three-dot range operator. Means excluding the end. So actually 1024 chars. See the spec as proof 😊

mikker: Your comment about allocating a copy is a valid point though 👍

mikker: Ooh, I _was_ off by one – `value[0...(max_length - 1)] + '…'` checks out

simitt: nitpick: the `s1 + s2` does also create a new string though..

mikker:
I think you are right about some parts of the serializers being a bit too naive right now.

I also felt a bit weird about the `Metadata` object not going through a serializer. But it's another kind of beast, I suppose.

mikker:
@simitt This makes parts of the serialization a bit more explicit. Some parts are still happily _Naively Hashable_. This also introduces a few unnecessary `null` objects that could just as well be omitted. Even though they probably don't cost much after gzipping.

lib/elastic_apm/transport/serializers/span_serializer.rb
```diff
@@ -12,17 +12,25 @@ def build(span)
               id: span.id,
               transaction_id: span.transaction_id,
               parent_id: span.parent_id,
-              name: span.name,
-              type: span.type,
+              name: keyword_field(span.name),
+              type: keyword_field(span.type),
               duration: ms(span.duration),
-              context: span.context&.to_h,
+              context: build_context(span.context),
```
simitt: Truncation of `span.context.http.method` is missing.

mikker:
@simitt Ready again when you are

mikker:
jenkins, retest this please

PR #242, opened 2018-11-19
Fix code typo transaction->with_transaction

harry-wood:
This looks like a typo to me.
(But please double-check this because I'm not sure)


mikker:
Definitely a typo! Thanks for fixing it 🙏

mikker:
Can I make you change the base of this PR to target `master` instead?

PR #247, opened 2018-11-21
Run worker in run loop

OpakAlex:
For sidekiq we have one issue, log term work with some load, we got message:
```
2018-11-20T13:13:09.995Z 8 TID-grgn7ph10 WARN: [ElasticAPM] Worker died with exception: #<Zlib::GzipFile::Error: closed gzip stream>
 WARN: [ElasticAPM] Queue is full (256 items), skipping…
```

This PR fixed issues with sidekiq.

### DIFF
--- a/lib/elastic_apm/spies/http.rb
+++ b/lib/elastic_apm/spies/http.rb
@@ -22,8 +22,10 @@ module ElasticAPM
             type = "ext.http_rb.#{method}"
 
             ElasticAPM.with_span name, type do |span|
-              req['Elastic-Apm-Traceparent'] =
-                transaction.traceparent.to_header(span_id: span.id)
+              if span && span.id
+                req['Elastic-Apm-Traceparent'] =
+                  transaction.traceparent.to_header(span_id: span.id)
+              end
 
               perform_without_apm(req, options)
             end
--- a/lib/elastic_apm/spies/net_http.rb
+++ b/lib/elastic_apm/spies/net_http.rb
@@ -24,8 +24,10 @@ module ElasticAPM
             type = "ext.net_http.#{method}"
 
             ElasticAPM.with_span name, type do |span|
-              req['Elastic-Apm-Traceparent'] =
-                transaction.traceparent.to_header(span_id: span.id)
+              if span && span.id
+                req['Elastic-Apm-Traceparent'] =
+                  transaction.traceparent.to_header(span_id: span.id)
+              end
 
               request_without_apm(req, body, &block)
             end
--- a/lib/elastic_apm/transport/base.rb
+++ b/lib/elastic_apm/transport/base.rb
@@ -25,7 +25,6 @@ module ElasticAPM
       attr_reader :config, :queue, :workers, :filters
 
       def start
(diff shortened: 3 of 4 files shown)


mikker:
Thank you for providing this @OpakAlex. For anyone else coming to this' sake: We're investigating this issue but haven't concluded anything yet. If you experience anything like this, please let us know.

mikker:
Why would the span be `nil` there? I'd only expect it to be if the agent was nil or the current transaction was nil but there are guard clauses that check for both of those.

Are you by any chance using S3 in your Sidekiq jobs and that could be why they keep dying?

OpakAlex:
But it's not depends on issue with died apm. This another one ;) Because in my case i download s3 data once in 24 hours.

OpakAlex:
One more issue @mikker , i detected memory leaks for my jobs. This happens after update apm gem. Any ideas why?

mikker:
@OpakAlex Could it be due to all the worker restarts from your eternal loop-change?

OpakAlex:
@mikker haha 100% YES, because without this fix it works onl +-1hour ;) For this time i didn't have memory usage data..... it works more, my limits 1070mb for sidekiq container. So it just working more time and reach this limit, than kubernetes restart container.

OpakAlex:
I think the best idea it's just port to v1.1..... because we didn't have any issues (only big payload)

OpakAlex:
So i think the best idea for me, just disable apm for sidekiq, untill you will do port to v1.1. Because i really don't have time to play this game. APM itself so good tool, but now it's a pain for me.

mikker:
I'm sorry to hear. I am, like I noted before, running this exact combination, Puma and Sidekiq, myself with no trouble.

Before you go, what's your version of Ruby, Rails and Sidekiq? Could you share your Gemfile? You know my email if you don't want to share it here. Would help a lot with the debugging.

OpakAlex:
done @mikker ,i am trying one more fix. will tell you result.

OpakAlex:
@mikker looks like `run_forever` make leaks for memory. i will try to remove this option. and check what happens.

OpakAlex:
@mikker so memory leaks fixed. But events stop sending after +-1h... We have same what we had 2 days ago ;)

mikker:
Yes, and that was the gzip streams suddenly being closed, right? That was the reported error that shut down the workers?

mikker:
Got it 👍 

Sampling can bring down the overhead of the agent when used with high-traffic apps. `config.transaction_sample_rate = 0.5` for example would mean _include spans with this transaction if `rand >= 0.5`_

OpakAlex:
ok, i will try this in config and will tell result.

mikker:
Sure, you can – but it wasn't a suggestions to fix your problem 😄

OpakAlex:
yes @mikker. 
So i have some statistic, i have 2 apps in production (with sidekiq workers).
One it's a OSS app: huginn. All works perfect. (sidikeq too)
And another one (Gemfile was send to you) where issue with sidekiq reports.
I think issue maybe with some gems... or...
what do you think @mikker ?

PR #249, opened 2018-11-21
Add support for Faraday

mikker:
Adds instrumentation and Distributed Tracing support to external HTTP requests by Faraday.

Also adds a `disable_in` method to the  `Net::HTTP` spy. This is so we don't get several spans for the same request as Faraday uses `Net::HTTP`under the covers by default.

### DIFF
--- a/lib/elastic_apm/spies/net_http.rb
+++ b/lib/elastic_apm/spies/net_http.rb
@@ -6,12 +6,31 @@ module ElasticAPM
     # @api private
     class NetHTTPSpy
       # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+      class << self
+        def disable_in
+          @disabled = true
+
+          begin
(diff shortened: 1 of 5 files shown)


mikker:
jenkins, retest this please

lib/elastic_apm/spies/net_http.rb
```diff
@@ -6,6 +6,22 @@ module Spies
...
+            yield
+          ensure
+            @disabled = false
+          end
+        end
+
+        def disabled?
+          @disabled ||= false
+        end
+      end
```
simitt: What if a `NetHTTP` and a `Faraday` request happen at the same time? The `disabled` is set as a class instance variable.

mikker: You are right. I'll make it thread local.

PR #254, opened 2018-11-23
Add extra logger and message when failing to configure

### LINKED ISSUE #251: Ruby agent silently fails to deliver data when invalid config option is set.
When using an unknown config option in the `elastic_apm.yml` the ruby agent silently fails to deliver data to the APM Server. 
Behaviour exists in versions `1.x` and `2.x`. 

Log entry in a test rails app showed:
```
W, [2018-11-23T16:44:26.719814 #13640]  WARN -- : [f443a45b-1459-47c2-a8b5-42b5ade29815] DEPRECATION WARNING: Leaving `ActiveRecord::ConnectionAdapters::SQLite3Adapter.represent_boolean_as_integer`
set to false is deprecated. SQLite databases have used 't' and 'f' to serialize
boolean values and must have old data converted to 1 and 0 (its native boolean
serialization) before setting this flag to true. Conversion can be accomplished
by setting up a rake task which runs

  ExampleModel.where("boolean_column = 't'").update_all(boolean_column: 1)
  ExampleModel.where("boolean_column = 'f'").update_all(boolean_column: 0)

for all models and all boolean columns, after which the flag must be set to
true by adding the following to your application.rb file:

  Rails.application.config.active_record.sqlite3.represent_boolean_as_integer = true
 (called from require at /Users/simitt/.rvm/gems/ruby-2.5.0@rails-apm-test/bundler/gems/apm-agent-ruby-97a08dfe860c/lib/elastic_apm/spies.rb:74)
```

Expected behaviour would be to gracefully ignore such config options or log an error and stop the ruby agent.

mikker:
`Rails.logger` isn't writing to `stdout` at the point of starting and configuring the agent in the railtie. Any exceptions come across while doing so might very well go by unseen because the developer doesn't see anything in their terminal.

This works around this by adding another `Logger` instance in the railtie that's sure to write to `stdout`. 

Consider: Should we write to both?

Fixes #251

### DIFF
--- a/lib/elastic_apm/config.rb
+++ b/lib/elastic_apm/config.rb
@@ -3,6 +3,7 @@
 require 'logger'
 require 'yaml'
 
+require 'elastic_apm/util/prefixed_logger'
 require 'elastic_apm/config/duration'
 require 'elastic_apm/config/size'
 
@@ -107,8 +108,6 @@ module ElasticAPM
       yield self if block_given?
 
       build_logger if logger.nil?
-    rescue NoMethodError => e
-      raise ConfigError, "No such option `#{e.name.to_s.delete('=')}'"
     end
 
     attr_accessor :config_file
@@ -148,10 +147,10 @@ module ElasticAPM
     attr_accessor :transaction_sample_rate
     attr_accessor :verify_server_cert
 
-    attr_reader   :custom_key_filters
-    attr_reader   :ignore_url_patterns
-    attr_reader   :span_frames_min_duration
-    attr_reader   :span_frames_min_duration_us
+    attr_reader :custom_key_filters
+    attr_reader :ignore_url_patterns
+    attr_reader :span_frames_min_duration
+    attr_reader :span_frames_min_duration_us
 
     attr_accessor :view_paths
     attr_accessor :root_path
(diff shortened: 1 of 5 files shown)


mikker:
jenkins, retest this please

mikker:
jenkins, retest this please

mikker:
You’re right 👍

mikker:
@simitt How do you find this approach?

lib/elastic_apm/config.rb
```diff
@@ -281,11 +298,21 @@ def set_from_env
...
     end
 
     def set_from_config_file
       return unless File.exist?(config_file)
       assign(YAML.load_file(config_file) || {})
+    rescue ConfigError => e
+      alert_logger.warn format(
+        'Failed to configure from config file: %s',
+        e.message
+      )
```
simitt: In the Sinatra app this leads to e.g.
```
W, [2018-12-04T12:59:51.383087 #97702]  WARN -- The option `debug_transactions=' has been removed.: [ElasticAPM]
W, [2018-12-04T12:59:51.383260 #97702]  WARN -- The option `enabled_environments=' has been removed.: [ElasticAPM]
W, [2018-12-04T12:59:51.383342 #97702]  WARN -- Failed to configure from arguments: No such option `nonexist': [ElasticAPM]
```

There is no `error` logged but APM doesn't start. I suggest to log this as an `error` instead of a `warning`.

simitt: Right it did start, I checked for documents in ES, but there were non so I thought it didnt. But it actually seems I caught an unrelated issue.

Issue #184, opened 2018-08-10
Faraday and net-http spy issue


Bugagazavr:
Hello, when i use faraday ( which uses `net-http` as default adapter ), i got duplicated spans. But i can't reproduce this with pure `net-http`. It looks like spy catches same request twice when Faraday is used.

How to reproduce:

```rb
require 'elastic-apm'
require 'faraday'

ElasticAPM.start
ElasticAPM.transaction 'Rack', 'request'

conn = Faraday.new(:url => 'https://www.google.com')
response = conn.get '/'

puts ElasticAPM.current_transaction.spans.count
puts ElasticAPM.current_transaction.spans.map(&:inspect)

ElasticAPM.current_transaction.release

ElasticAPM.transaction 'Rack', 'request'

Net::HTTP.get(URI('https://www.google.com'))

puts ElasticAPM.current_transaction.spans.count
puts ElasticAPM.current_transaction.spans.map(&:inspect)
```

Result:

```sh
RACK_ENV=production ruby 1.rb
# 2
# <ElasticAPM::Span id:0 name:"GET www.google.com" type:"ext.net_http.GET">
# <ElasticAPM::Span id:1 name:"GET www.google.com" type:"ext.net_http.GET">
# 1
# <ElasticAPM::Span id:0 name:"GET www.google.com" type:"ext.net_http.GET">
```

mikker:
Thank you for reporting this, @Bugagazavr. I'll see if I can reproduce. The thumb reactions tell me you aren't the only one 🙄

mikker:
This should be solved in the `2.1.1` release from yesterday. Can you confirm?

Bugagazavr:
@mikker Well i need some time to upgrade, thank you for reminder

Bugagazavr:
Looks like APM server 6.5 required?

mikker:
Yes, `gem 'elastic-apm', '~> 2'` is APM Server 6.5+ required.

Bugagazavr:
Confirmed.

PR #263, opened 2018-12-05
Fix faraday call with url_prefix

Bugagazavr:
This is fix for spy, when used relative faraday calls.

Related: https://github.com/elastic/apm-agent-ruby/issues/184

Case 1 ( url is nil ):
```rb
conn = Faraday::Connection.new('http://api.example.com/v1/some_resource/1')
response = conn.get
#          ArgumentError:
#            bad argument (expected URI object or URI string)
```

In this case url is nil, this means we calling `GET http://api.example.com/v1/some_resource/1` directly. Spy got crashed when try to parse url.

Case 2 ( url is a path ):
```rb
conn = Faraday::Connection.new('http://api.example.com')
response = conn.get('/v1/some_resource/1')
```

That's not raises an error, but we got a nil instead of host name

```
[1] pry(main)> URI('/v1/some_resource/1')
=> #<URI::Generic /v1/some_resource/1>
[2] pry(main)> URI('/v1/some_resource/1').host
=> nil
```

Workaround: Try to use host from `url_prefix`

When url filled:

```
[1] pry(main)> Faraday::Connection.new('http://api.example.com').url_prefix.host
=> "some.api.com"
[2] pry(main)> Faraday::Connection.new(url: 'http://api.example.com').url_prefix.host
=> "some.api.com"
```

When url is not filled:

```
[1] pry(main)> Faraday::Connection.new.url_prefix
=> #<URI::HTTP http:/>
[2] pry(main)> Faraday::Connection.new.url_prefix.host
=> nil
```

### DIFF
--- a/lib/elastic_apm/spies/faraday.rb
+++ b/lib/elastic_apm/spies/faraday.rb
@@ -15,7 +15,11 @@ module ElasticAPM
               return run_request_without_apm(method, url, body, headers, &block)
             end
 
-            host = URI(url).host
+            host = if url_prefix.is_a?(URI) && url_prefix.host
+                     url_prefix.host
+                   else
+                     URI(url).host
+                   end
 
             name = "#{method.upcase} #{host}"
             type = "ext.faraday.#{method}"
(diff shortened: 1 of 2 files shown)


mikker:
Great explanation! Thank you 😊

mikker:
Jenkins, test this please

kylejginavan:
Glad to see this fix is in master. ETA on a new gem version?

PR #283, opened 2019-01-08
JRuby performance

davishmcclurg:
The agent added more overhead than I expected to our JRuby app. I believe
the main issue is `caller` is just slow on JRuby 😞 I couldn't find any better
fix than skipping the call if the span isn't sampled. It would be nice to have a
separate setting for stacktrace sample rate, but I didn't want to go down that
path before bringing up the issue.

The second commit moves stacktrace building to the worker thread. Not as
important but I thought it made sense.

### DIFF
--- a/lib/elastic_apm.rb
+++ b/lib/elastic_apm.rb
@@ -230,7 +230,7 @@ module ElasticAPM # rubocop:disable Metrics/ModuleLength
         break unless span && include_stacktrace
         break unless agent.config.span_frames_min_duration?
 
-        span.original_backtrace = caller
+        span.original_backtrace ||= caller
       end
     end
 
--- a/lib/elastic_apm.rb
+++ b/lib/elastic_apm.rb
@@ -225,9 +225,13 @@ module ElasticAPM # rubocop:disable Metrics/ModuleLength
         name,
         type,
         context: context,
-        backtrace: include_stacktrace ? caller : nil,
         trace_context: trace_context
-      )
+      ).tap do |span|
+        break unless span && include_stacktrace
+        break unless agent.config.span_frames_min_duration?
+
+        span.original_backtrace = caller
+      end
     end
 
     # Ends the current span
--- a/lib/elastic_apm/span.rb
+++ b/lib/elastic_apm/span.rb
@@ -33,8 +33,7 @@ module ElasticAPM
     # rubocop:enable Metrics/ParameterLists
 
     attr_accessor :name, :type, :original_backtrace, :parent_id, :trace_context
-    attr_reader :context, :stacktrace, :duration,
-      :timestamp, :transaction_id
+    attr_reader :context, :stacktrace, :duration, :timestamp, :transaction_id
 
     def id
       trace_context&.span_id
@@ -59,10 +58,6 @@ module ElasticAPM
     def done(end_time: Util.micros)
       stop end_time
 
-      if should_build_stacktrace?
-        build_stacktrace
-      end
-
       self
     end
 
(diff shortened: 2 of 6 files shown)


lib/elastic_apm/agent.rb
```diff
@@ -122,13 +122,8 @@ def end_transaction(result = nil)
...
 
-    def start_span(name = nil, type = nil, backtrace: nil, context: nil)
-      instrumenter.start_span(
-        name,
-        type,
-        backtrace: backtrace,
-        context: context
-      )
+    def start_span(name = nil, type = nil, context: nil)
+      instrumenter.start_span(name, type, context: context)
```
mikker: Let’s keep the backtrack argument so the Apis stay the same as far as possible. Makes the multi-link delegation easier to grok. I don’t think this will change anything around your fix as it’ll just be nil and do nothing.

Do you agree?

lib/elastic_apm/span.rb
```diff
@@ -75,6 +70,15 @@ def running?
       started? && !stopped?
     end
 
+    def stacktrace
+      return @stacktrace if defined? @stacktrace
```
mikker: We can skip `defined?`?

lib/elastic_apm/span.rb
```diff
@@ -75,6 +70,15 @@ def running?
       started? && !stopped?
     end
 
+    def stacktrace
```
mikker: Why lazy load the prop instead? Perf or preference? 😊

lib/elastic_apm.rb
```diff
@@ -210,12 +210,11 @@ def span(name, type = nil, context: nil, include_stacktrace: true, &block)
...
     def start_span(name, type = nil, context: nil, include_stacktrace: true)
-      agent&.start_span(
-        name,
-        type,
-        context: context,
-        backtrace: include_stacktrace ? caller : nil
-      )
+      agent&.start_span(name, type, context: context)&.tap do |span|
+        if include_stacktrace && agent.config.span_frames_min_duration?
+          span.original_backtrace = caller
```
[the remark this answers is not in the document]
davishmcclurg: The safe navigation operator should handle that:

```
>> nil.tap { puts '!' }
!
=> nil
>> nil&.tap { puts '!' }
=> nil
```

Might be confusing though 🤷‍♂️

lib/elastic_apm/agent.rb
davishmcclurg: And assign it to `span.original_backtrace` in the instrumenter? I think that would be fine. Might want to switch to `span.original_backtrace ||= caller` above then

lib/elastic_apm/span.rb
davishmcclurg: The goal was to move the `stacktrace_builder.build` call to a worker thread. I didn't test the performance difference though, just thought it made sense to do less work on the main thread.

Now that I'm looking at it again, it's probably better to keep an explicit `build_stacktrace` step rather than lazy load. That way it's easier to determine where the work actually happens.

davishmcclurg:
For the perf testing, I added the gem to our jruby/rails package and benchmarked a bunch of search queries. The transactions had ~12 spans (from faraday) each. The agent added ~15% overhead. I didn't do much investigation after I discovered how slow JRuby's `caller` is. :(

lib/elastic_apm/span.rb
davishmcclurg: I think so. This was meant to memoize `stacktrace`

mikker:
15% is really quite a lot. Does most of that go away without `caller`?

davishmcclurg:
Here are the average response times (ms) from my tests:

```
no apm      39.675422
no caller   43.189033
caller      56.885760
```

mikker:
Thanks. This is collection stack traces and source lines?

When I profiled APM half a year-ish ago this was by far the most expensive thing to do.

davishmcclurg:
Just stack traces, no source lines. Source lines don't work for us because we compile are ruby files to java classes during the package building process.

mikker:
I corrected the few things we discussed and rebased on master. Might need to reset your local branch if you want to add more to this.

I'm going to merge this as is and create an issue to move stacktrace building off the main thread. Thanks again!

mikker:
Spec failure is unrelated to this

Issue #212, opened 2018-09-25
Add support for Rails.credentials in elastic_apm.yml


mkemme:
Configuration file `elastic_apm.yml `should support ERB expansion so that credentials could be stored in credential store (like in `database.yml`). For example:

`secret_token: <%= Rails.application.credentials.dig(:elastic_apm, :secret_token) %>`

mikker:
Hi @mkemme! I've thought about this. I think you're right as this seems to be a pretty common pattern in yml-based configs.

If you're feeling adventurous PRs are of course welcome 😄

mikker:
Closed by https://github.com/elastic/apm-agent-ruby/pull/288.

PR #286, opened 2019-01-14
Support for erb style on config file

PedroSeda:
This PR adds support to erb styled variables on config file

[Issue 212](https://github.com/elastic/apm-agent-ruby/issues/212)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 - Support for [OpenTracing](https://opentracing.io) ([#273](https://github.com/elastic/apm-agent-ruby/pull/273))
 - Add capture_* options ([#279](https://github.com/elastic/apm-agent-ruby/pull/279))
+- Support ERB tags in config file ([#286](https://github.com/elastic/apm-agent-ruby/pull/286))
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


PedroSeda:
> Hi @PedroSeda! This is wonderful – I see you found the issue too 😃 Thank you.
> 
> I like how Rails does this, only evaluating as ERB if the file is called `.yml.erb`. That'll also work better with peoples' editors' syntax highlighting (guessing). If file is just `.yml` we only load.
> 
> I think the config option stays `config_file: config/elastic_apm.yml` but we check programmaticaly if an ERB version exists.
> 
> Do you agree?

I might be wrong, but I think `credentials.yml`, `secrets.yml` and `database.yml` all support ERB tags without the `.erb` at the end. I think it's default to Rails, but it might not be to other frameworks.

PedroSeda:
> We also need
> 
> * [ ]  Update the docs to mention this
> * [ ]  `require 'erb'` somewhere to be sure(?)
> * [ ]  Update the changelog
> 
> Would you like to do this? Then you are very welcome to get the credit. Otherwise let me know and I'll sail this safely to shore 🚢

Will do, I'll tackle that this afternoon

PedroSeda:
I added an example using an ERB tag to the docs. Let me know if that's enough :smile:

mikker:
Thanks! I added yet another note to the configuration section in the PR referenced.

mikker:
Merged via https://github.com/elastic/apm-agent-ruby/pull/288.

PR #302, opened 2019-01-29
Add Compatibility for Optional CA Cert

### LINKED ISSUE #301: Allow Setting SSL CA Certificate
Without the ability to provide a CA certificate file, the APM agent cannot verify a self-signed server certificate. Without verification, there isn't much point in using TLS at all.

HTTP.rb supports using CA certificates through OpenSSL::SSL::SSLContext, so all the agent needs is to take an `ssl_context` in as an optional parameter and pass it in to HTTP.rb.

marothstein:
# Overview

Providing a CA cert allows the use of a
self-signed server certificate without disabling
certificate verification.

This adds an optional `ssl_ca_cert` parameter and
conditionally adds it to the default_options
hash of HTTP.rb. That includes it automatically
in the connection and subsequent requests.

## Changes

- Adds a new optional config parameter, `ssl_ca_cert`
- Adds a new environment variable config parameter, `ELASTIC_APM_SSL_CA_CERT`
- Adds the `ssl_ca_cert` path value to the HTTP.rb default options in an OpenSSL::SSL::SSLContext variable to allow OpenSSL to verify the APM Server certificate against the provided CA cert

## Usage

The path to the CA file can be specified using the parameter name explicitly, through the file, or by environment variable. 

### Config Variable

```ruby
ElasticAPM::Config.new(ssl_ca_cert: "/path/to/ca.pem")
```

### Config File

```yaml
---
service_name: "MyApp"
pool_size: 2 
ssl_ca_cert: "/path/to/ca.pem"
```

### Environment Variable

```shell
# .env
ELASTIC_APM_SSL_CA_CERT=/path/to/ca.pem
```

## Related Issues

Closes #301.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -129,7 +129,7 @@ WARNING: Secret tokens only provide any real security if your APM server use TLS
 
 Maximum amount of objects kept in queue, before sending to APM Server.
 
-If you hit this limit you either have to increase the agent's 
+If you hit this limit you either have to increase the agent's
 <<config-pool-size,worker pool size>> or it could mean the agent has trouble
 connecting to APM Server. The <<config-log-path,logs>> should tell you what
 went wrong.
@@ -511,6 +511,20 @@ Set it to `-1` to collect stack traces for all spans.
 Set it to `0` to disable stack trace collection for all spans.
 
 It has to be provided in *<<config-format-duration, duration format>>*.
+
+[float]
+[[config-ssl-ca-cert]]
+==== `ssl_ca_cert`
+
+[options="header"]
+|============
+| Environment                | `Config` key    | Default | Example
+| `ELASTIC_APM_SSL_CA_CERT`  | `ssl_ca_cert`   | `nil`   | /path/to/ca.pem
+|============
+
+This parameter should contain the path to a .pem formatted CA certificate. This
+allows verification of self-signed certificates instead of disabling server
+certificate verification.
 
 [float]
 [[config-transaction-max-spans]]
--- a/lib/elastic_apm/transport/connection.rb
+++ b/lib/elastic_apm/transport/connection.rb
@@ -41,6 +41,16 @@ module ElasticAPM
           headers['Authorization'] = "Bearer #{token}"
         end
 
+        # Set SSL CA Cert
+        if config.use_ssl? && config.ssl_ca_cert
+          ssl_ctx = OpenSSL::SSL::SSLContext.new
+          ssl_ctx.ca_file = config.ssl_ca_cert
+
+          # include SSL context in default options
+          client_options = HTTP.default_options.merge(ssl_context: ssl_ctx)
+          HTTP.default_options = client_options
+        end
+
         @client = HTTP.headers(headers).persistent(@url)
 
         @mutex = Mutex.new
--- a/lib/elastic_apm/config.rb
+++ b/lib/elastic_apm/config.rb
@@ -47,6 +47,7 @@ module ElasticAPM
       source_lines_span_app_frames: 5,
       source_lines_span_library_frames: 0,
       span_frames_min_duration: '5ms',
+      ssl_ca_cert: nil,
       transaction_max_spans: 500,
       transaction_sample_rate: 1.0,
       verify_server_cert: true,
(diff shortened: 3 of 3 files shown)


lib/elastic_apm/transport/connection.rb
```diff
@@ -41,6 +41,16 @@ def initialize(config, metadata)
           headers['Authorization'] = "Bearer #{token}"
         end
 
+        # Set SSL CA Cert
+        if (config.use_ssl? && config.ssl_ca_cert)
```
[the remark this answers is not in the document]
marothstein: Fied.

mikker:
This is wonderful, @marothstein! Thanks a bunch for doing this. It was already planned in #248 and is of course a feature we should provide.

To make sure we streamline with the other languages' agents can I make you rename the option to `ELASTIC_APM_SERVER_CERT` and equivalent for the non-ENV versions?

mikker:
jenkins, test this please

docs/configuration.asciidoc
```diff
@@ -512,6 +512,23 @@ Set it to `0` to disable stack trace collection for all spans.
...
+|============
+| Environment                | `Config` key    | Default | Example
+| `ELASTIC_APM_SSL_CA_CERT`  | `ssl_ca_cert`   | `nil`   | /path/to/ca.pem
+|============
+
+This parameter should contain the path to a .pem formatted CA certificate. This
+allows verification of self-signed certificates instead of disabling server
+certificate verification.
+
+The path provided must be accessible to the agent for HTTP.rb to load the
```
mikker: I'd just leave out the last part, `  for HTTP.rb to load the certificate.` Implementation detail 😉

lib/elastic_apm/transport/connection.rb
```diff
@@ -41,6 +41,16 @@ def initialize(config, metadata)
...
         end
 
+        # Set SSL CA Cert
+        if config.use_ssl? && config.ssl_ca_cert
+          ssl_ctx = OpenSSL::SSL::SSLContext.new
+          ssl_ctx.ca_file = config.ssl_ca_cert
+
+          # include SSL context in default options
+          client_options = HTTP.default_options.merge(ssl_context: ssl_ctx)
+          HTTP.default_options = client_options
```
mikker: Won't this leak and lead to weird bugs if the user's app also uses `HTTP.rb`? Can we pass it when we initialize the `@client` below instead?

mikker:
jenkins, retest this please

docs/configuration.asciidoc
marothstein: One of the things I found frustrating about the Elastic documentation is that there are a lot of details that are hinted at but not defined. I added this as a helpful note for those implementing this. In general, I’m in favor of making things easier for people.

I agree with you on this case, though, and am happy to remove.

lib/elastic_apm/transport/connection.rb
marothstein: That's a good callout. I went this route because I was having a hard time finding how to pass the  `ssl_context` to the client with chained methods. I'll try again.

docs/configuration.asciidoc
mikker: I'm sorry to hear. I want the docs of the agent to be as good as possible but it's never easy to walk the line between too much and too little. If you remember or find any places you think are lacking please let me know where it is and I'll see what I can do!

marothstein: Looking again at this, I totally agree with you on the HTTP.rb point. It should not be in the docs and I will remove it.

lib/elastic_apm/transport/connection.rb
mikker: Thank you!

marothstein:
Actually, @mikker, I don't think that `ELASTIC_APM_SERVER_CERT` is the correct variable name. The server cert is _not_ what we are providing. I know it's sounds like a parallel with  `ELASTIC_APM_VERIFY_SERVER_CERT`, but we are actually providing the CA certificate and should say so explicitly.

docs/configuration.asciidoc
marothstein: Done.

mikker:
You are right about the naming. I was mistaking the two. For this case, your original name is better so let's stick with that 👍

Can I make you add a test case too? I don't know how easy it is to test this sort of stuff _end-to-end_ (probably not easy?) but let's at least have a a mock then that ensures we hit this code path once.

mikker:
Had a talk with a colleague and we landed on the option name `ELASTIC_APM_SERVER_CA_CERT`.

mikker:
Continued in #315

mikker:
Thank you for your work, @marothstein!
Hope you don't mind I do the last bit to get this merged.

PR #310, opened 2019-02-08
Add contributing.md

bmorelli25:
Here's the shell of the CONTRIBUTING.md file. There are two sections you'll need to address - both at the bottom: Testing and Releasing. Let me know if I can help in any way. 

Closes elastic/apm-agent-ruby#307

PR #313, opened 2019-02-12
Update contributing.md

mikker:
Cont. from elastic/apm-agent-ruby#310


mikker:
Test failures are only in `rails#master` so we'll ignore

PR #326, opened 2019-02-21
Store parent_id and id in trace_context

### LINKED ISSUE #322: Investigate infinite recursion relation when using DT

simitt:
The `TraceContext` did only define a `span_id` which was used as `id` and as `parent_id` at the same time in a transaction. This PR intends to clearly separate `id` used in `transaction` and `span` events from the `parent_id` parsed from the request headers. 

fixes #322

### DIFF
--- a/lib/elastic_apm/trace_context.rb
+++ b/lib/elastic_apm/trace_context.rb
@@ -8,8 +8,8 @@ module ElasticAPM
     VERSION = '00'
     HEX_REGEX = /[^[:xdigit:]]/.freeze
 
-    TRACE_ID_N = 16
-    ID_N = 8
+    TRACE_ID_LENGTH = 16
+    ID_LENGTH = 8
 
     def initialize(
       version: VERSION,
@@ -19,19 +19,16 @@ module ElasticAPM
       recorded: true
     )
       @version = version
-      @trace_id = trace_id || hex_id(TRACE_ID_N)
-      @parent_id = span_id # rename to parent_id with next major version bump
-      @id = id || hex_id(ID_N)
+      @trace_id = trace_id || hex(TRACE_ID_LENGTH)
+      # TODO: rename to parent_id with next major version bump
+      @parent_id = span_id
+      @id = id || hex(ID_LENGTH)
       @recorded = recorded
     end
 
     attr_accessor :version, :id, :trace_id, :parent_id, :recorded
 
     alias :recorded? :recorded
-
-    def self.for_transaction(sampled: true)
-      new(recorded: sampled)
-    end
 
     # rubocop:disable Metrics/AbcSize
     def self.parse(header)
@@ -65,14 +62,13 @@ module ElasticAPM
     end
 
     def ensure_parent_id
-      @parent_id ||= hex_id(ID_N)
-      @parent_id
+      @parent_id ||= hex(ID_LENGTH)
     end
 
     def child
       dup.tap do |tc|
         tc.parent_id = tc.id
-        tc.id = hex_id(ID_N)
+        tc.id = hex(ID_LENGTH)
       end
     end
 
@@ -92,7 +88,7 @@ module ElasticAPM
 
     private
 
-    def hex_id(len)
+    def hex(len)
       SecureRandom.hex(len)
     end
   end
--- a/lib/elastic_apm/trace_context.rb
+++ b/lib/elastic_apm/trace_context.rb
@@ -8,36 +8,29 @@ module ElasticAPM
     VERSION = '00'
     HEX_REGEX = /[^[:xdigit:]]/.freeze
 
+    TRACE_ID_N = 16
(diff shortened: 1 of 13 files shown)


lib/elastic_apm/trace_context.rb
```diff
@@ -8,36 +8,29 @@ class InvalidTraceparentHeader < StandardError; end
     VERSION = '00'
     HEX_REGEX = /[^[:xdigit:]]/.freeze
 
+    TRACE_ID_N = 16
```
mikker: `TRACE_ID_LENGTH` maybe?

lib/elastic_apm/trace_context.rb
```diff
@@ -8,36 +8,29 @@ class InvalidTraceparentHeader < StandardError; end
...
       trace_id: nil,
       span_id: nil,
+      id: nil,
       recorded: true
     )
       @version = version
-      @trace_id = trace_id
-      @span_id = span_id
+      @trace_id = trace_id || hex_id(TRACE_ID_N)
+      @parent_id = span_id # rename to parent_id with next major version bump
```
mikker: Should we add something grep'able like `TODO` to this comment?

lib/elastic_apm/trace_context.rb
```diff
@@ -71,12 +64,36 @@ def hex_flags
       format('%02x', flags.to_i(2))
     end
 
+    def ensure_parent_id
+      @parent_id ||= hex_id(ID_N)
+      @parent_id
```
mikker: I think the above line already returns this

lib/elastic_apm/trace_context.rb
```diff
@@ -71,12 +64,36 @@ def hex_flags
...
+        tc.id = hex_id(ID_N)
+      end
     end
 
     def to_header
-      format('%s-%s-%s-%s', version, trace_id, span_id, hex_flags)
+      format('%s-%s-%s-%s', version, trace_id, id, hex_flags)
+    end
+
+    # @deprecated Use parent_id instead
```
mikker: I actually have a `Deprecations` module. But that outputs logs which might not be what we want as this is more of an internally pointed deprecation?

lib/elastic_apm/trace_context.rb
```diff
@@ -71,12 +64,36 @@ def hex_flags
...
+    def span_id=(span_id)
+      @parent_id = span_id
+    end
+
+    private
+
+    def hex_id(len)
```
mikker: Your time in Go is making its mark on your variable names – saving 3 characters here 😀

lib/elastic_apm/transaction.rb
```diff
@@ -26,28 +30,20 @@ def initialize(
...
       Util.reverse_merge!(@context.tags, tags) if tags
 
-      if trace_context
-        @parent_id = trace_context.span_id
-        @trace_context = trace_context
-      else
-        @trace_context = TraceContext.for_transaction(sampled: sampled)
-      end
+      @trace_context = trace_context ||
+                       TraceContext.for_transaction(sampled: sampled)
```
mikker: we almost don't need the `for_transaction()` now? Could be just `new(sampled: sampled)` and the result would be the same?

lib/elastic_apm/transaction.rb
```diff
@@ -63,8 +59,8 @@ def done?
 
     deprecate :done?, :stopped?
 
-    def trace_id
-      trace_context&.trace_id
+    def ensure_parent_id
+      trace_context.ensure_parent_id
```
mikker: Could use the delegation too? Or do you feel like this is different?

spec/elastic_apm/trace_context_spec.rb
```diff
@@ -24,7 +25,9 @@ module ElasticAPM
         it { expect { subject }.to_not raise_error }
         its(:version) { should eq '00' }
         its(:trace_id) { should eq '0af7651916cd43dd8448eb211c80319c' }
-        its(:span_id) { should eq 'b7ad6b7169203331' }
+        its(:parent_id) { should eq 'b7ad6b7169203331' }
+        its(:id) { should_not be_nil }
+        its(:id) { should_not eq 'b7ad6b7169203331' }
```
mikker: I feel attacked 😆

lib/elastic_apm/trace_context.rb
simitt: Internally, this is not used any more. The only reason for keeping this accessors in is for being backwards compatible in case someone used them, since `TraceContext` is part of the public API. Having this in mind, I think it would be fine if the deprecated msg would be logged. WDYT?

mikker: I agree. Let's used the `Deprecations` module then 👍

mikker:
Can you add a short entry to the changelog as well?

PR #327, opened 2019-02-22
Keep request.body as object when form data

mikker:
This changes request.body back to the object we get from Rack, making it way easier to sanitize. If not form data we leave the body untouched.


mikker:
Test failures unrelated.

PR #330, opened 2019-02-25
Add deprecation warning to TraceContext span_id

simitt:


mikker:
Nice 👍 why do we need to differentiate between the getters and setters?

simitt:
The getters couldn't be called with the flexible number of args and a block, and the alias needs to add the `=` at the end while removing it from the name. I didn't spend too much time on this, maybe there's a more elegant way of solving it.

mikker:
I see. We could maybe get around this by using `send(method_name)` instead of `#{method_name}` but it's not that important. You can try if you want – otherwise LGTM 😊

mikker:
A few of the tests output the deprecation methods – we should probably migrate whatever they do to the new methods?

simitt:
You're absolutely right, changed that.

PR #335, opened 2019-02-26
Generate dedicated Context for errors / Align capture_body config values

mikker:
Right now, errors get their `Context` copied over from `current_transaction` which isn't necessarily the correct dat, and config options might differ for the two different event types.

This PR makes sure we generate separate, new contexts for errors only copying over relevant data such as `tags` and `custom`.

It started out as aligning on the values for the option `capture_body` so that's in there too.


lib/elastic_apm.rb
```diff
@@ -285,27 +285,43 @@ def with_span(
...
     # @param rack_env [Rack::Env] A Rack env
     # @return [Context] The built context
-    def build_context(rack_env)
-      agent&.build_context(rack_env)
+    def build_context(
+      deprecated_env = nil,
+      rack_env: nil,
+      for_type: :transaction
+    )
+      if !rack_env && (rack_env = deprecated_env)
```
simitt: Why did you introduce `deprecated_env` here and make `rack_env` a keyword arg but then require it?
How about `def build_context(rack_env:, for_type:)` instead?

lib/elastic_apm/config.rb
```diff
@@ -296,6 +296,31 @@ def collect_metrics?
...
+
+      case value
+      when true
+        alert_logger.warn "Boolean value for option `capture_body' has " \
+          "been deprecated. Setting to 'all'"
+        @capture_body = 'all'
+      when false
+        alert_logger.warn "Boolean value for option `capture_body' has " \
+          "been deprecated. Setting to 'off'"
+        @capture_body = 'off'
```
simitt: 👍

lib/elastic_apm/context.rb
```diff
@@ -9,13 +9,24 @@
...
+    attr_accessor :user
+    attr_reader :custom
+    attr_reader :tags
+
+    def empty?
+      return false if tags.any?
+      return false if custom.any?
+      return false unless user.empty?
+
+      true
```
simitt: You can sum up the last check to have:
```
def empty?
      return false if tags.any?
      return false if custom.any?
      user.empty?
end
```

lib/elastic_apm/context_builder.rb
```diff
@@ -42,6 +42,16 @@ def apply_to_request(context, rack_env)
...
     # rubocop:enable Metrics/MethodLength, Metrics/AbcSize
 
+    def should_capture_body?(for_type)
+      option = config.capture_body
+
+      return true if option == 'all'
+      return true if option == 'transactions' && for_type == :transaction
+      return true if option == 'errors' && for_type == :error
+
+      false
```
simitt: The configuration doesn't change between requests, so you could store it as a class variable and then create another class var holding a hash with mappings for the type. Something along the lines of:
```
m = { transaction: true, error: true }
case config.capture_body
when 'off'
  m[:transaction] = false
  m[:error] = false
when 'transaction':
  m[:error] = false
when 'error':
  m[:transaction] = false
```

And then access it within `should_capture_body?(for_type)` with `m[for_type]`

lib/elastic_apm/error_builder.rb
```diff
@@ -59,10 +50,20 @@ def add_stacktrace(error, kind, backtrace)
...
+
       error.transaction_id = transaction.id
       error.transaction = { sampled: transaction.sampled? }
+      error.trace_id = transaction.trace_id
+      error.parent_id = ElasticAPM.current_span&.id || transaction.id
+
+      return unless transaction.context
+
+      Util.reverse_merge!(error.context.tags, transaction.context.tags)
+      Util.reverse_merge!(error.context.custom, transaction.context.custom)
```
simitt: Why are you merging `tags` and `custom`, is this aligned with how other agents behave? I'd expect to find only the configured tags and customs if specifically set for an error.

lib/elastic_apm.rb
mikker: Backwards compatibility. Used to be 1 non-keyword arg. To keep it working with a deprecation warning, I have both.

lib/elastic_apm/context.rb
mikker: Right, I thought about that but liked the explicitness of this approach :)

lib/elastic_apm/context_builder.rb
mikker: Oh, so memoize the method basically?

lib/elastic_apm/error_builder.rb
mikker: Some do but it differs so I started https://github.com/elastic/apm/issues/62

lib/elastic_apm/context.rb
simitt: ok, that's fine with me.

lib/elastic_apm/context_builder.rb
mikker: Skipping because the lookups aren't expensive in the first place 😊

mikker:
Hmm, I think I'm missing `context.request` on errors using this branch. Will add a test and investigate.

PR #343, opened 2019-03-05
Don't auto-start in Railtie if inside Rails' console

### LINKED ISSUE #333: ELASTIC_APM_ENABLED_ENVIRONMENTS is not working
I am exporting `ELASTIC_APM_ENABLED_ENVIRONMENTS=production` and running `rails console` still I can see messages related to APM booting and logs polluting my console

mikker:
A bit hard to test other than manually. Suggestions welcome.

---

Closes #333

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Both APIs are backwards compatible with fallbacks and deprecation warnings, sche
 ### Changed
 
 - Errors get their own contexts, perhaps leading to slightly different (but more correct) results. ([#335](https://github.com/elastic/apm-agent-ruby/pull/335))
+- The agent no longer starts automatically inside Rails' console ([#343](https://github.com/elastic/apm-agent-ruby/pull/343))
 
 ## 2.5.0 (2019-03-01)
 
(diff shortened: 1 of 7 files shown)


mikker:
I wanted to add a test but because the Rails test is set up as it is, we'd need a second, completely separate test context just to test this so I skipped it.

mikker:
Okay, I added a test and found a bug – are you happy now, @simitt ? 🙈

reist:
Was there a point to adding the INFO line about APM being disabled? It's pretty annoying seeing it on every console start.

mikker:
Used to print a line about starting so I imagined it would make sense to have one about NOT starting. I can see how some would want their boot-ups silent. Would a config option to silence such warnings suffice?

I suppose there already is one if you change the log level. But I was thinking about one specifically for muting any boot up message(s)

reist:
That'd be nice, yes.
About the log level, I couldn't find where it needs to be changed. Setting it in `config/application.rb` or `config/elastic_apm.yml` applies too late, I think.

mikker:
I think it's actually using what I call the `alert_logger` because it is so soon that the regular `Rails.log` wouldn't output to `stdout`. A setting is a much more sensible solution.

PR #394, opened 2019-05-01
Revert "net-http spy: use Module.prepend instead of alias"

mikker:
Reverts elastic/apm-agent-ruby#382

Turns out when other code tries to `class_eval` on top of `Module#prepend`ed code it turns into an infinitely recursive loop, ultimately raising a `stack level too deep` error.

See for example https://docs.newrelic.com/docs/agents/ruby-agent/troubleshooting/systemstackerror-stack-level-too-deep

While `Module#prepend` is the more sensible and perhaps modern solution, I'm reverting this for now.

### DIFF
--- a/lib/elastic_apm/spies/net_http.rb
+++ b/lib/elastic_apm/spies/net_http.rb
@@ -28,33 +28,32 @@ module ElasticAPM
         end
       end
 
-      module Overrides
-        # @api private
-        def request(req, body = nil)
-          return super if ElasticAPM::Spies::NetHTTPSpy.disabled?
-
-          transaction = ElasticAPM.current_transaction
-          return super unless transaction
-
-          host, = req['host'] && req['host'].split(':')
-          method = req.method
-
-          host ||= address
-
-          name = "#{method} #{host}"
-          type = "ext.net_http.#{method}"
-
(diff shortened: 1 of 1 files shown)


mikker:
cc @timogoebel – was this the issue that you ran into yourself or what was it you were trying to get away from?

mikker:
On a second look I see this was the exact problem, you were trying to get away from. Hmm, what to do? We can either break one or the other way.

timogoebel:
> Hmm, what to do? We can either break one or the other way.

Yeah, reverting the patch is unfortunate. In the past we have solved this by using `alias_method` for projects that use Rails < 5 and `Module.prepend` for projects that use Rails >= 5. I'm wondering if we can use metaprogramming to offer both options via a config flag or if we should release a new major version that switches to `Module.prepend`.

Did the Module.prepend patch cause problems with code from this gem or with other code?

mikker:
I have an app that has Skylight installed alongside Elastic APM and they use `alias` to extend Net::HTTP.

I guess we could potentially have this problem everywhere we use `alias` in the spies. Providing both solutions seems a bit much, yet deciding to break either one or the other isn't nice either. Some kind of either detecting the approach or having an abstraction that can do both depending on a config flag were my ideas as well.

Issue #287, opened 2019-01-14
[APM-CI][Integration test][Ruby] Improve testing framework for JRuby to include different JDKs


alvarolobato:
After an issue found with JRuby 9.x and JDK >= 10 we should improve how we are testing the different version combination.

@kuisathaverat this is something to keep in mind when migrating to pipelines.

alvarolobato:
@kuisathaverat we need to prioritize this when you have some time.

kuisathaverat:
Which JDKs do we want to test? JDK 7+ to the latest?

mikker:
7 or 8 is probably fine. If you ask me, we don't need every possible combination.

kuisathaverat:
let's start with 7,8, latest-1 (11), latest (12)

v1v:
I managed to create a rough implementation of the JRuby-JDK versions and also asked the owner of the JRuby docker images whether they are interested to use that approach:
- https://github.com/cpuguy83/docker-jruby/issues/50

Running some tests locally to ensure we could potentially create a Pipeline to generate and push those docker images to our internal docker registry.

v1v:
Supported docker images

```
sh run.sh -e 1.7 -a build -r "docker.elastic.co/observability-ci"
 docker images
	Image docker.elastic.co/observability-ci/jruby:9.1-7-jdk           GENERATED  
	Image docker.elastic.co/observability-ci/jruby:9.2-11-jdk          GENERATED  
	Image docker.elastic.co/observability-ci/jruby:9.2-8-jdk           GENERATED  
	Image docker.elastic.co/observability-ci/jruby:9.1-7-jdk           GENERATED  
	Image docker.elastic.co/observability-ci/jruby:9.2-11-jdk          GENERATED  
	Image docker.elastic.co/observability-ci/jruby:9.2-12-jdk          GENERATED  
	Image docker.elastic.co/observability-ci/jruby:9.2-13-jdk          GENERATED  
	Image docker.elastic.co/observability-ci/jruby:9.2-8-jdk           GENERATED  
	Image docker.elastic.co/observability-ci/jruby:9.1-onbuild         GENERATED  
	Image docker.elastic.co/observability-ci/jruby:9.2-8-jre-alpine    GENERATED  
	Image docker.elastic.co/observability-ci/jruby:9.1-7-jre-alpine    GENERATED  
	Image docker.elastic.co/observability-ci/jruby:9.1-7-jdk-alpine    GENERATED  
[...]

PR #403, opened 2019-05-03
[APM-CI][Integration test][Ruby] Support JRuby with different JDK versions

v1v:
Caused by https://github.com/elastic/apm-agent-ruby/issues/287

## Highlights
- Support jruby with different jdk versions
- Enable docker login with a retry + sleep in case any infra issues.
- Ruby versions moved from `-` split to `:` as it is the way we can support our internal docker registry.
- Fix pipeline to bypass a weird script approval misleading message.

## Tasks
- [x] Test locally for the current list of versions in `spec/.jenkins_ruby.yml`
- [x] Validate whether the Jenkinsfile works as expected. [here](https://apm-ci.elastic.co/job/apm-agent-ruby/job/apm-agent-ruby-mbp/job/PR-403/14/)
- [x] --Script approval is required-- False! It's solved with https://github.com/elastic/apm-agent-ruby/pull/403/commits/db5ab60768dd3cf15f2e3321fd280af7bd6dedfe
- [x] Benchmark is also required. Unfortunately, the PRs are not validating it and I've just found it

## Todo
- I'd rather prefer to handle the stage naming in the UI Pipeline view in another PR as it does require to change the shared library too. What do I mean? See the below screenshot:
 
image

The name of the stage is based on the name of the docker images plus the framework, as the name of the new docker images are hosted in our internal docker registry, their names are quite long. 

What do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ Gemfile.lock
 spec/elastic_apm.log
 spec/ruby-agent-junit.xml
 .gem
+*.bulk
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,10 +74,10 @@ should "Squash and merge".
 
 To do a full test run, use either `bundle exec rspec` or `rake spec`. Individual specs should also run as expected. The Mongo test needs a Mongo instance running, but will start one itself if Docker is installed.
 
-To test other platform, use the Docker setup and scripts like `spec.sh RUBY FRAMEWORK`.
+To test other platform, use the Docker setup and scripts like `spec.sh RUBY_DOCKER_IMAGE FRAMEWORK`.
 
 ```sh
-$ spec/scripts/spec.sh ruby-2.6 rails-5.2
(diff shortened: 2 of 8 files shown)


v1v:
jenkins run the tests please

.gitignore
```diff
@@ -15,3 +15,4 @@ Gemfile.lock
 spec/elastic_apm.log
 spec/ruby-agent-junit.xml
 .gem
+*.bulk
```
mikker: What's this?

v1v: That's a file auto-generated when running the benchmarking locally:
- https://github.com/elastic/apm-agent-ruby/pull/403/files#diff-964990e4e95f38885151096be29638e5R39

v1v:
For the record, we will work on providing a nicer UI in the Jenkins pipeline to simplify the name of the stages, but that's a follow-up.

mdelapenya:
Shaun approves 😎 

image

PR #407, opened 2019-05-06
[APM-CI][All] Make the trigger by comment case-insensitive

v1v:


mikker:
Perhaps we should add the `.ci` dir to ignored directories for Codecov?

mikker:
Also there's something weird going on with it. It lists many ruby files as changed with this PR although they aren't.

Issue #350, opened 2019-03-18
Add "Performance tuning" docs


mikker:
https://github.com/elastic/apm/issues/66

cc @bmorelli25 -- Is it correct that this is something I should write and not something you will do?

bmorelli25:
Yes, you'll need to create the first draft for this doc. Even bulleted ideas would work, I can handle the formatting if you'd like.

saviogl:
:+1:

estolfo:
Some tips to add: 
- Try disabling stacktraces with span_frames_min_duration: 0
- Stacktraces are the most expensive thing the agent does, performance wise
- If you're doing custom span, you can also disable it with the argument include_stacktrace: false to keep them on for errors, auto-instrumented things, etc

PR #413, opened 2019-05-08
Add debugging docs page

mikker:
The purpose of this page is to provide a few helpful hints at how to debug a non-working agent as well as have a canonical place to point people to, if they experience their event queues filling up.

As soon as the page is up I want to actually link to it in the warning message to point folks in the right direction instead of (unrelated perhaps) issues. We can use the new url shortener to make a nice, short link.

@bmorelli25 I'm assigning you as you probably have a few ideas on how to improve the copy. See this as a first draft and do whatever you want to it 😉

### DIFF
--- a/docs/debugging.asciidoc
+++ b/docs/debugging.asciidoc
@@ -26,5 +26,5 @@ Things to consider:
 
   - Is `server_url` misconfigured or APM Server down? If the agent fails to connect you will also see log messages containing `Connection error` or `Couldn't establish connection to APM Server`.
   - Experiencing high load? The agent can spawn multiple instances of its Workers that pick off the queue by changing the option `pool_size` (default is `1`).
-  - If you have high load you could also consider setting `transaction_sample_rate` to something smaller than `1.0`. This determines whether to include _spans_ for every _transaction_. If you have enough traffic, skipping some (probably) identical spans won't have a noticeable effect on your data.
+  - If you have high load you may also consider setting `transaction_sample_rate` to something smaller than `1.0`. This determines whether to include _spans_ for every _transaction_. If you have enough traffic, skipping some (probably) identical spans won't have a noticeable effect on your data.
 
--- a/docs/debugging.asciidoc
+++ b/docs/debugging.asciidoc
@@ -1,0 +1,23 @@
+[[debugging]]
+== Debugging the agent itself
+
+Hopefully the agent Just Works™ for you but depending on your situation the agent could need some tuning.
+
+Firstly, to know more about what's going on inside the agent, you can increase the amount of log messages it writes by setting the log level with the option `log_level = 0` -- `0` being the level of most messages, `DEBUG`.
+
+[float]
+[[debugging-errors]]
+=== Errors
+
+[float]
+[[debugging-errors-queue-full]]
+==== `Queue is full (256 items), skipping…`
+
+The agent has an internal queue that holds events after they are done, before they are safely serialized and sent to APM Server. To avoid using up all of your memory, the queue has a fixed size. Taking from the queue happens faster than adding to it but it can happen that the queue fills, hence the warning.
+
+Things to consider:
+
+  - Is `server_url` misconfigured or APM Server down?
+  - Experiencing high load? The agent can spawn multiple instances of its Workers that pick off the queue by changing the option `pool_size` (default is `1`).
+  - If you have high load you could also consider setting `transaction_sample_rate` to something smaller than `1.0`. This determines whether to include _spans_ for every _transaction_. If you have enough traffic, skipping some (probably) identical spans won't have a noticeable effect on your data.
+
--- a/docs/debugging.asciidoc
+++ b/docs/debugging.asciidoc
@@ -3,7 +3,7 @@
 
 Hopefully the agent Just Works™, but depending on your situation the agent might need some tuning.
 
-Firstly, to know more about what's going on inside the agent, you can increase the amount of log messages it writes by setting the log level with the option `log_level = 0` -- `0` being the level of most messages, `DEBUG`.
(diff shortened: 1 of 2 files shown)


mikker:
Perhaps related: https://github.com/elastic/apm-agent-ruby/issues/350

Not sure I still think we need the perf tuning doc if we have this. So the question is which is the better framing for the (sort of) same info?

docs/debugging.asciidoc
```diff
@@ -0,0 +1,23 @@
+[[debugging]]
+== Debugging the agent itself
+
+Hopefully the agent Just Works™ for you but depending on your situation the agent could need some tuning.
+
+Firstly, to know more about what's going on inside the agent, you can increase the amount of log messages it writes by setting the log level with the option `log_level = 0` -- `0` being the level of most messages, `DEBUG`.
```
axw: Not directly related to docs, but do you think it would be be helpful to enable users to use log level names, like "DEBUG" or "INFO"? What's typical in Ruby apps?

docs/debugging.asciidoc
```diff
@@ -0,0 +1,23 @@
...
+
+[float]
+[[debugging-errors]]
+=== Errors
+
+[float]
+[[debugging-errors-queue-full]]
+==== `Queue is full (256 items), skipping…`
+
+The agent has an internal queue that holds events after they are done, before they are safely serialized and sent to APM Server. To avoid using up all of your memory, the queue has a fixed size. Taking from the queue happens faster than adding to it but it can happen that the queue fills, hence the warning.
```
axw: > Taking from the queue happens faster than adding to it but it can happen that the queue fills

Aren't those two statements contradictory? If taking always happens faster, then the queue cannot fill. Whether it's typical or not really comes down to the application, and what level of load is normal.

Perhaps restate like "Depending on the load of your application, events may be added to the queue faster than they can be consumed ..."

docs/debugging.asciidoc
```diff
@@ -0,0 +1,23 @@
...
+
+Things to consider:
+
+  - Is `server_url` misconfigured or APM Server down?
```
axw: If the URL is misconfigured or the server is down, will the user see other errors too? Perhaps list them out and refer to them here, so the user can manually correlate?

docs/debugging.asciidoc
mikker: It's common to use the constants `Logger::DEBUG` (`== 0`) for example. I just thought this was easier than explaining how to get to those depending on your situation. Via ENV you cant, in the config file you need the template tags `<%= Logger::DEBUG %>`, in normal ruby they're just there. The Getting Started docs assume you configure with a config file, so I could just assume the same here and let people figure out themselves how it might be different in their situation?

docs/debugging.asciidoc
mikker: I think I just missed a "_usually faster_"

docs/debugging.asciidoc
mikker: To your question: Could be a _nice to have_ to convert `:debug` to `Logger::DEBUG` I suppose. But not common or expected, in my experience.

docs/debugging.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+In your `config/elastic_apm.yml`:
+
+[source,yaml]
+----
+log_level: <%= Logger::DEBUG %>
+----
+
+[float]
+[[debugging-errors]]
+=== Errors
```
axw: (Thanks for updating the wording below.)

Following the same line of reasoning, I think using the title "Errors" gives the impression that this is definitely exceptional behaviour. Not sure what a better title is though. Maybe "Log Messages"?

bmorelli25:
Sorry I haven't had a chance to look at this yet. It's on my radar, will get to it ASAP

mikker:
No worries! Whenever you can is soon enough 😊 see you soon!

docs/debugging.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+
+[float]
+[[debugging-log-messages]]
+=== Log messages
+
+[float]
+[[debugging-errors-queue-full]]
+==== `Queue is full (256 items), skipping…`
+
+The agent has an internal queue that holds events after they are done, before they are safely serialized and sent to APM Server. To avoid using up all of your memory, the queue has a fixed size. Depending on your load and server setup, events may be added to the queue faster than they are consumed, hence the warning.
```
bmorelli25:
```suggestion
The agent has an internal queue that holds events after they are done, and before they are safely serialized and sent to APM Server. To avoid using up all of your memory, this queue has a fixed size. Depending on your load and server setup, events may be added to the queue faster than they are consumed, hence the warning.
```

docs/debugging.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+
+Things to consider:
+
+  - Is `server_url` misconfigured or APM Server down? If the agent fails to connect you will also see log messages containing `Connection error` or `Couldn't establish connection to APM Server`.
+  - Experiencing high load? The agent can spawn multiple instances of its Workers that pick off the queue by changing the option `pool_size` (default is `1`).
+  - If you have high load you could also consider setting `transaction_sample_rate` to something smaller than `1.0`. This determines whether to include _spans_ for every _transaction_. If you have enough traffic, skipping some (probably) identical spans won't have a noticeable effect on your data.
```
bmorelli25:
```suggestion
  - If you have high load you may also consider setting `transaction_sample_rate` to something smaller than `1.0`. This determines whether to include _spans_ for every _transaction_. If you have enough traffic, skipping some (probably) identical spans won't have a noticeable effect on your data.
```

mikker:
Good suggestions, @bmorelli25 
💙💚💛💜❤️

bmorelli25:
retest this please

PR #414, opened 2019-05-08
Update benchmark script

mikker:
The benchmark spins up a tiny Rails app with one endpoint that does one db query, then sends 10.000 requests through it _with_ the agent and then _without_ the agent.

### DIFF
--- a/bench/app.rb
+++ b/bench/app.rb
@@ -1,38 +1,53 @@
+#!/usr/bin/env ruby
 # frozen_string_literal: true
 
 $LOAD_PATH.unshift(File.expand_path('../lib', __dir__))
+require 'active_record'
+require 'action_controller/railtie'
 require 'elastic-apm'
+require 'elastic_apm/railtie'
 
-class App
-  module Helpers
-    def with_app(config = {})
-      app = App.new(config)
-      app.start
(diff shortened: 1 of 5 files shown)


axw:
@mikker I think it would be more useful to see something like the Java agent's benchmarks:

https://github.com/elastic/apm-agent-java/blob/62acdb1f349474ade1fe037e10bfbe803e7ca580/apm-agent-benchmarks/src/main/java/co/elastic/apm/agent/benchmark/ElasticApmContinuousBenchmark.java#L60

This is a simplistic servlet (HTTP request handler) which ones a single SQL query. A servlet is constructed, and then requests are passed through it and statistics gathered to measure the overhead with and without the agent enabled.

The Go agent does something similar, but I measure the HTTP and SQL instrumentation independently.

kuisathaverat:
When it is finished ping us to add the logic to send the bulk files to an Elasticsearch.

mikker:
Ready now, @kuisathaverat 🙏

kuisathaverat:
we did it now https://github.com/elastic/apm-agent-ruby/blob/master/Jenkinsfile#L269-L270 so you can check the benchmarks at kibana

PR #417, opened 2019-05-08
Reintroduce skipping events by returning nil in filter

mikker:
This was somehow removed.

It gets a little bit weird as the filters are destructive and alter the original payload (to save copying every payload ever) but are still expected to have a return value. Seems not so ruby-like in a way.

An alternative implementation could use exceptions to make skipping a bit more obvious. Like, require a filter to `raise Filters::Skip, 'status ping'` to skip. But that would mean changing the api and be a breaking change (unless we do both but not sure it's worth it – wdyt @simitt ?

### DIFF
--- a/lib/elastic_apm/transport/worker.rb
+++ b/lib/elastic_apm/transport/worker.rb
@@ -58,7 +58,10 @@ module ElasticAPM
 
       def serialize_and_filter(resource)
         serialized = serializers.serialize(resource)
-        @filters.apply!(serialized)
+
+        # if a filter returns nil, it means skip the event
+        return nil if @filters.apply!(serialized) == Filters::SKIP
+
         JSON.fast_generate(serialized)
       rescue Exception
         error format('Failed converting event to JSON: %s', resource.inspect)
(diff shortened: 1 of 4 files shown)


lib/elastic_apm/transport/worker.rb
```diff
@@ -58,7 +58,10 @@ def process(resource)
 
       def serialize_and_filter(resource)
         serialized = serializers.serialize(resource)
-        @filters.apply!(serialized)
+
+        # if a filter returns nil, it means skip the event
+        return nil if @filters.apply!(serialized) == Filters::SKIP
```
simitt: Reading `filters.apply!(serialized)` I would expect the `filters` to be changed in place and not `serialized`. (I know that you didn't introduce this right now).  Would it help if you derive the serialized events from a common base class offering a method for filtering, e.g. `serialized_event.apply!(filters)`. Afterwards, you could check if `serialized_event.body` is nil and move on or not.

mikker: Could you write a short (but slightly longer) example?

mikker: Ah, now I get it. Renaming the method `apply_to` would solve it too 😉

Issue #387, opened 2019-04-16
Capture Ruby runtime metrics


nehaduggal-zz:
Ruby agent to capture runtime metrics to help visualize the impact of underlying infrastructure in context of the application.

Metrics to be captured by the Ruby agent: 
- Time in Garbage collection 
- Frequency of GC
- Memory usage
- Thread count
	

APM Meta issue: https://github.com/elastic/apm/issues/73

There should be a dashboard for all these metrics in apm-contrib repo.

mikker:
Closed by #487

PR #437, opened 2019-06-06
[Draft] Add VM metrics

mikker:
Closes elastic/apm-agent-ruby#387

### DIFF
--- a/docs/metrics.asciidoc
+++ b/docs/metrics.asciidoc
@@ -12,11 +12,15 @@ You can adjust the interval by setting <<config-metrics-interval,`metrics_interv
 
 The metrics will be stored in the `apm-*` index and have the `processor.event` property set to `metric`.
 
+[float]
+[[metrics-system]]
+=== System metrics
+
 **Note:** Metrics from the Ruby agent are Linux only for now.
 
 [float]
 [[metric-system.cpu.total.norm.pct]]
-=== `system.cpu.total.norm.pct`
+==== `system.cpu.total.norm.pct`
 
 * *Type:* Float
 * *Format:* Percent
@@ -26,7 +30,7 @@ normalised by the number of cores.
 
 [float]
 [[metric-system.memory.total]]
-=== `system.memory.total`
+==== `system.memory.total`
 
 * *Type:* Long
 * *Format:* Bytes
@@ -35,7 +39,7 @@ The total memory of the system in bytes.
 
 [float]
 [[metric-system.memory.actual.free]]
-=== `system.memory.actual.free`
+==== `system.memory.actual.free`
 
 * *Type:* Long
 * *Format:* Bytes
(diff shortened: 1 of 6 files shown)


lib/elastic_apm/metrics/vm.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      def sample
+        @sampler.sample
+      end
+
+      def collect
+        return unless sampler
+
+        sample.each_with_object({}) do |(key, value), snap|
+          snap[:"#{SCOPE}#{key}"] = value
```
axw: I don't think we need to be capturing _all_ of the GC statistics, and we should be a bit more explicit about the names we choose to report. I suggest these names as a start:

 - ruby.gc.count
 - ruby.gc.time
 - ruby.thread.count (or ruby.threads? Java is using `.count`, but I'm not sure if that's the ideal convention)

Do you intend to add heap usage too?

lib/elastic_apm/metrics/vm.rb
```diff
@@ -0,0 +1,30 @@
...
+    # @api private
+    class VM
+      def initialize(_config)
+        @total_time = 0
+      end
+
+      # rubocop:disable Metrics/MethodLength
+      def collect
+        stat = GC.stat
+        @total_time += GC::Profiler.total_time
```
axw: I think we should only report `ruby.gc.time` if the GC profiler is enabled.

lib/elastic_apm/metrics/vm.rb
```diff
@@ -0,0 +1,30 @@
...
+        GC::Profiler.clear
+        thread_count = Thread.list.count
+
+        {
+          'ruby.gc.count': stat[:count],
+          'ruby.gc.time': @total_time,
+          'ruby.heap.live': stat[:heap_live_slots],
```
axw: Repeating my message from Slack, for posterity: let's call this `ruby.heap.slots.live`, and `ruby.heap.slots.free` below.

lib/elastic_apm/metrics/vm.rb
```diff
@@ -0,0 +1,30 @@
...
+          'ruby.heap.free': stat[:heap_free_slots],
+          'ruby.allocations.total': stat[:total_allocated_objects],
```
axw: this should probably have a ".heap" in it right? how about `ruby.heap.objects.allocated`?

mikker:
I'll include the docs and changelog entry in this PR. Just wanted to get the impl. right first.

mikker:
Thanks once again for helping out, @axw!

mikker:
@bmorelli25 Docs update spotted!

docs/metrics.asciidoc
```diff
@@ -12,11 +12,15 @@ You can adjust the interval by setting <<config-metrics-interval,`metrics_interv
 
 The metrics will be stored in the `apm-*` index and have the `processor.event` property set to `metric`.
 
+[float]
+[[metrics-system]]
+=== System metrics
+
 **Note:** Metrics from the Ruby agent are Linux only for now.
```
bmorelli25:
```suggestion
NOTE: Metrics from the Ruby agent are Linux only for now.
```

Let's turn this into a real "note"

mikker:
Continued in #487 because Jenkins

PR #448, opened 2019-06-18
Save request cookies on context, filter cookies in headers

### LINKED ISSUE #447: Filter session id stored in cookies
The apm agent currently does not filter the session id from the request `Cookie` header or the response `Set-Cookie` header. For security reasons, this information should be filtered by default.

Please find below some sample data:

Request:
```
timezone=Europe%2FBerlin; _session_id=2e50ed79c9edfefddaae4ec28041462c
```

Response:
```
_session_id=2e50ed79c9edfefddaae4ec28041462c; path=/; expires=Tue, 18 Jun 2019 14:28:01 -0000; HttpOnly; secure; SameSite=Lax
```

mikker:
Fixes elastic/apm-agent-ruby#447

### DIFF
--- a/lib/elastic_apm/transport/filters/secrets_filter.rb
+++ b/lib/elastic_apm/transport/filters/secrets_filter.rb
@@ -14,7 +14,8 @@ module ElasticAPM
           /secret/i,
           /token/i,
           /api[-._]?key/i,
-          /session[-._]?id/i
+          /session[-._]?id/i,
+          /(set[-_])?cookie/i
         ].freeze
 
         VALUE_FILTERS = [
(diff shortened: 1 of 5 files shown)


lib/elastic_apm/transport/filters/secrets_filter.rb
```diff
@@ -30,6 +31,7 @@ def initialize(config)
         def call(payload)
           strip_from! payload.dig(:transaction, :context, :request, :headers)
           strip_from! payload.dig(:transaction, :context, :request, :env)
+          strip_from! payload.dig(:transaction, :context, :request, :cookies)
```
axw: Just trying to understand what's going on: is this going to strip secrets from `request.cookies`? I can't see a test for this.

mikker: Yes, that's correct. There's not a specific test for this field but there are for the others and the implementation is the same:
https://github.com/elastic/apm-agent-ruby/blob/master/spec/elastic_apm/transport/filters/secrets_filter_spec.rb

mikker:
jenkins, retest this please

lib/elastic_apm/transport/filters/secrets_filter.rb
axw: OK. Generally I'd like to see full coverage for sensitive things like this, but I think we should address this with a functional test for all agents.

PR #449, opened 2019-06-18
Fix disabling verifying server certificate

### LINKED ISSUE #443: Agent ignores verify_server_cert, cannot verify self-signed certs
Originally reported by @simitt:

Option `verify_server_cert: false` in `config/elastic_apm.yml` is ignored. Setting `ELASTIC_APM_VERIFY_SERVER_CERT=false` in ENV is also ignored. The ruby agent still tries to verify and fails with
```
[ElasticAPM] Couldn't establish connection to APM Server:
"#<OpenSSL::SSL::SSLError: SSL_connect returned=1 errno=0 state=error: certificate verify failed (self signed certificate in certificate chain)>"
```

mikker:
Fixes elastic/apm-agent-ruby#443

### DIFF
--- a/lib/elastic_apm/transport/connection.rb
+++ b/lib/elastic_apm/transport/connection.rb
@@ -130,11 +130,20 @@ module ElasticAPM
         ].join(' ')
       end
 
-      def build_ssl_context
-        return unless @config.use_ssl? && @config.server_ca_cert
+      def build_ssl_context # rubocop:disable Metrics/MethodLength
+        return unless @config.use_ssl?
 
         OpenSSL::SSL::SSLContext.new.tap do |context|
-          context.ca_file = @config.server_ca_cert
+          if @config.server_ca_cert
+            context.ca_file = @config.server_ca_cert
+          end
+
+          context.verify_mode =
+            if @config.verify_server_cert
+              OpenSSL::SSL::VERIFY_PEER
+            else
+              OpenSSL::SSL::VERIFY_NONE
+            end
         end
       end
     end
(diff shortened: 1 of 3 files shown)


lib/elastic_apm/transport/connection.rb
```diff
@@ -130,11 +130,20 @@ def build_user_agent(metadata)
...
-          context.ca_file = @config.server_ca_cert
+          if @config.server_ca_cert
+            context.ca_file = @config.server_ca_cert
+          end
+
+          context.verify_mode =
+            if @config.verify_server_cert
+              OpenSSL::SSL::VERIFY_PEER
+            else
+              OpenSSL::SSL::VERIFY_NONE
```
simitt: If I correctly understand this fix, then if both configs are set `server_ca_cert` and `verify_server_cert: false` then the server certificate is _not_ verified. I think the python agent handles this differently - as soon as a certificate is provided it is verified, no matter what. Might be a bit more secure. 
However you decide to do this, please add a note in the docs clarifying it.

mikker: I see how it would be weird to supply a CA cert and also disable it, but I don't necessarily feel like we should prohibit that weird behaviour. With this, the agent does exactly like you tell it to?

simitt: Fine with me, just thought it might make sense to document it.

axw: FWIW: the Python agent, as well as others, don't currently support `SERVER_CA_CERT`. If a specific server (not _CA_) cert is specified with `SERVER_CERT`, then `VERIFY_SERVER_CERT` is ignored.

I think this is OK.

mikker:
jenkins, retest this please

mikker:
jenkins, retest this please

mikker:
Test failures unrelated

PR #464, opened 2019-07-03
Support Central Config

### LINKED ISSUE #446: Support remote configuration
https://github.com/elastic/apm/issues/76

mikker:
Closes elastic/apm-agent-ruby#446 
Depends on elastic/apm-agent-ruby#463

### DIFF
--- a/lib/elastic_apm/central_config.rb
+++ b/lib/elastic_apm/central_config.rb
@@ -84,14 +84,16 @@ module ElasticAPM
 
     private
 
+    # rubocop:disable Metrics/MethodLength
     def handle_success(resp)
-      # 304 Not Modified
-      unless resp.status == 304
+      if resp.status == 304
+        info 'Received 304 Not Modified'
+      else
         update = JSON.parse(resp.body.to_s)
         assign(update)
-      end
 
-      info 'Updated config from APM Server'
+        info 'Updated config from Kibana'
+      end
 
       schedule_next_fetch(resp)
 
@@ -100,6 +102,7 @@ module ElasticAPM
       error 'Failed to apply remote config, %s', e.inspect
       debug { e.backtrace.join('\n') }
     end
+    # rubocop:enable Metrics/MethodLength
 
     def handle_error(error)
       error(
--- a/lib/elastic_apm/central_config.rb
+++ b/lib/elastic_apm/central_config.rb
@@ -38,9 +38,7 @@ module ElasticAPM
     end
 
     def fetch_and_apply_config
-      @task&.cancel
-      @task =
-        Concurrent::Promise
+      Concurrent::Promise
         .execute(&method(:fetch_config))
         .on_success(&method(:handle_success))
         .rescue(&method(:handle_error))
@@ -68,10 +66,14 @@ module ElasticAPM
     # rubocop:enable Metrics/MethodLength
 
     def assign(update)
+      # For each updated option, store the original value,
+      # unless already stored
       update.each_key do |key|
         @modified_options[key] ||= config.get(key.to_sym)&.value
       end
 
+      # If the new update doesn't set a previously modified option,
+      # revert it to the original
       @modified_options.each_key do |key|
         next if update.key?(key)
         update[key] = @modified_options.delete(key)
@@ -82,16 +84,12 @@ module ElasticAPM
 
     private
 
-    # rubocop:disable Metrics/MethodLength
     def handle_success(resp)
-      update =
-        if !resp.body.empty?
-          JSON.parse(resp.body.to_s)
-        else
-          {}
-        end
-
-      assign(update)
+      # 304 Not Modified
+      unless resp.status == 304
+        update = JSON.parse(resp.body.to_s)
+        assign(update)
+      end
 
       info 'Updated config from APM Server'
 
@@ -102,7 +100,6 @@ module ElasticAPM
       error 'Failed to apply remote config, %s', e.inspect
       debug { e.backtrace.join('\n') }
     end
-    # rubocop:enable Metrics/MethodLength
 
     def handle_error(error)
       error(
--- a/lib/elastic_apm/central_config.rb
+++ b/lib/elastic_apm/central_config.rb
@@ -1,0 +1,139 @@
+# frozen_string_literal: true
+
+require 'elastic_apm/central_config/cache_control'
+
+module ElasticAPM
+  # @api private
+  class CentralConfig
+    include Logging
+
+    # @api private
+    class ResponseError < InternalError
+      def initialize(response)
+        @response = response
+      end
+
+      attr_reader :response
+    end
+    class ClientError < ResponseError; end
+    class ServerError < ResponseError; end
+
+    DEFAULT_MAX_AGE = 300
+
+    def initialize(config)
+      @config = config
+      @modified_options = {}
+      @service_info = {
+        'service.name': config.service_name,
+        'service.environment': config.environment
+      }.to_json
+    end
+
+    attr_reader :config, :task
+
+    def start
+      return unless config.central_config?
+
+      fetch_and_apply_config
+    end
+
+    def fetch_and_apply_config
+      @task&.cancel
+      @task =
+        Concurrent::Promise
+        .execute(&method(:fetch_config))
+        .on_success(&method(:handle_success))
+        .rescue(&method(:handle_error))
+    end
+
+    def stop
+      @task&.cancel
+    end
+
+    # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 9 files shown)


lib/elastic_apm/central_config.rb
```diff
@@ -0,0 +1,107 @@
...
+        raise NotFoundError, resp
+      when 500
+        raise ServerError, resp
+      end
+    end
+
+    private
+
+    def handle_success(resp)
+      update = JSON.parse(resp.body)
```
simitt: There's not necessarily a body sent, as we work with `etags`. 
From a server's perspective: if the `etag` the agent sends up equals the `etag` of the newly fetched resource from Kibana, the server responds with a `nil` body, http status `NotModified`, the `etag`, and an according `Cache-Control: max-age:30s` header. 
see [server side response](https://github.com/elastic/apm-server/blob/master/beater/agent_config_handler.go#L73)

lib/elastic_apm/central_config.rb
```diff
@@ -0,0 +1,107 @@
...
+      )
+
+      schedule_next_fetch(error.response)
+    end
+
+    def perform_request
+      Http.post(
+        config.server_url + '/agent/v1/config/',
+        body: @service_info,
+        headers: { content_type: 'application/json' }
```
simitt: you should set the `etag` header in the request.

lib/elastic_apm/central_config.rb
```diff
@@ -0,0 +1,107 @@
...
+
+      schedule_next_fetch(resp)
+
+      true
+    end
+
+    def handle_error(error)
+      error(
+        'Failed fetching config: %s, trying again in %d seconds',
+        error.response.body, DEFAULT_MAX_AGE
```
simitt: `DEFAULT_MAX_AGE` is not correct here, as usually you will follow the `max-age` from the response cache-control header.

lib/elastic_apm/central_config.rb
```diff
@@ -0,0 +1,107 @@
...
+      end
+
+      attr_reader :response
+    end
+    class NotFoundError < ResponseError; end
+    class ServerError < ResponseError; end
+
+    DEFAULT_MAX_AGE = 300
+
+    def initialize(config, default_polling_interval: 300)
```
simitt: please reuse `DEFAULT_MAX_AGE` here

I am not sure `default_polling_interval` is really needed, as not used anywhere.

lib/elastic_apm/central_config.rb
```diff
@@ -0,0 +1,107 @@
...
+      when 500
+        raise ServerError, resp
+      end
+    end
+
+    private
+
+    def handle_success(resp)
+      update = JSON.parse(resp.body)
+      config.assign(update)
```
simitt: What happens if the response contains config options that the agent doesn't have specified? Does it raise an error or log a warning? I think there is no explicit deal around this atm, and afaik free text configs will be allowed, so you should account for it.

mikker: It'll warn and ignore. _Free text configs_ as in how?

simitt: ok that seems sane. 

> Free text configs as in how?

I think it was discussed to allow adding unstructured config in Kibana allowing a more dynamic support for beta configs.

bmorelli25:
👋Mikker. I'll be out of the office when you're back, but here's some boilerplate documentation you can hopefully adapt to match your config style: elastic/apm#86. Also, don't forget to update the ruby [precedence](https://www.elastic.co/guide/en/apm/agent/ruby/current/configuration.html). Would be great if it included a link to the new configuration variable.

lib/elastic_apm/central_config.rb
mikker: I think I'll add it if and when it becomes a real posibility.

mikker:
jenkins, retest this please

lib/elastic_apm/central_config.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+    end
+    # rubocop:enable Metrics/MethodLength
+
+    private
+
+    def handle_success(resp)
+      unless resp.body.empty?
+        update = JSON.parse(resp.body)
+        config.assign(update)
```
axw: It looks like you're just splatting the new config over the top, is that right? What happens when the config is removed in Kibana? We should fall back to locally defined config (https://github.com/elastic/apm/issues/76#issuecomment-510019654). If you're deferring that, please document the limitation and create a TODO/issue.

mikker: Got it. Nice catch.

mikker:
@simitt, @axw Will you give this a second, quick review?

mikker:
Squashed all commits to hopefully make the tests run ....

lib/elastic_apm/central_config.rb
```diff
@@ -0,0 +1,139 @@
...
+      when 400..499
+        raise ClientError, resp
+      when 500..599
+        raise ServerError, resp
+      end
+    end
+    # rubocop:enable Metrics/MethodLength
+
+    def assign(update)
+      update.each_key do |key|
```
axw: I think this could do with some comments. Let me see if I'm getting it right:

- when a config attribute wasn't previously defined remotely, but now is, we'll record the current local value in `@modified_options`
- when a config attribute was previously defined remotely, but no longer is, we'll remove the value from `@modified_options` and revert to it

Is it possible for the local config to be updated after the agent is started?

lib/elastic_apm/central_config.rb
```diff
@@ -0,0 +1,139 @@
...
+
+    # rubocop:disable Metrics/MethodLength
+    def fetch_config
+      resp = perform_request
+
+      case resp.status
+      when 200..299
+        resp
+      when 300..399
+        resp
```
axw: does that mean we'll call `handle_success` for 3xx? in which case, won't that revert config on 304?

spec/elastic_apm/central_config_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+
+          subject.fetch_and_apply_config
+          sleep 0.1
+
+          expect(subject.task).to be_pending
+          expect(subject.task.initial_delay).to eq 123
+        end
+      end
+
+      context 'when server responds 304' do
```
axw: probably a more realistic test would be 200 followed by 304 - then you can also check that the 304 response doesn't cause the config to be reverted

lib/elastic_apm/central_config.rb
mikker: That's correct. It is technically possible to update some if not most of them on the go but I don't suspect anyone does it nor is it described nor encouraged anywhere.

Ruby being Ruby; it is possible.

lib/elastic_apm/central_config.rb
mikker: As far as I understand it, the 304s will have empty bodies and so `handle_success` will handle it.

spec/elastic_apm/central_config_spec.rb
mikker: That's sort of what these test: https://github.com/mikker/apm-agent-ruby/blob/906b3b62616bfe272b1e8e043530ea23f6cdcb98/spec/elastic_apm/central_config_spec.rb#L157-L168

lib/elastic_apm/central_config.rb
mikker: Oh no, you're right. I'll fix.

axw: I don't think we should rely on the 304 body being empty or non-JSON in any case.

spec/elastic_apm/central_config_spec.rb
mikker: I modified the 304 test for this

lib/elastic_apm/central_config.rb
mikker: Agreed. I modified `handle_success` to explicitly handle 304s

lib/elastic_apm/central_config.rb
```diff
@@ -0,0 +1,136 @@
...
+    private
+
+    def handle_success(resp)
+      # 304 Not Modified
+      unless resp.status == 304
+        update = JSON.parse(resp.body.to_s)
+        assign(update)
+      end
+
+      info 'Updated config from APM Server'
```
axw: move this up into the unless, since there's no update? and maybe a debug log message on 304?

v1v:
jenkins run the tests please

mikker:
Continued in #484 because Jenkins

PR #479, opened 2019-08-06
MongoSpy: Change span name format and add command to the context's db.statement field

estolfo:
### Change 1

 I've added the collection and command names to the `Span::Context::DB` in the `MongoSpy`.
The only tricky thing is, there are some MongoDB commands (mostly db admin commands) that aren't executed in the context of a collection. In this case, the format of the command is:

`{ command_name: 1 }`

with the integer 1 being in place of a collection name. For this case, I left the `db.statement` value as `nil`, as it was before.

For a command executed in the context of a collection, the mongodb command is in the format:

`{ command_name: collection_name }`

For this case, I made the `db.statement` value `collection_name.command_name`.

Let me know if that's ok or if you'd like to fill in the `db.statement` value somehow for the db admin commands (the first case).

### Change 2

I changed the `mongo_spec` to start/stop the `ElasticAPM` agent and mongodb server in `before(:all)`/`after(:all)` hooks, as starting mongodb in particular takes a few seconds. This way, the two specs can run without having to each start mongodb. 
Nevertheless, looked at the other spec files and I noticed that often the agent is started directly in the spec and not in a `before(:all)` hook. Is there a reason for that?

### Change 1

I've changed the span name to be in the format `db.collection.operation`.
When the command is at the db level, the collection name is the integer 1. For this case, the span name is in the format `db.operation`

### Change 2

I've added the entire command to the `db.statement` value. This is consistent with other spies, most notably the `SequelSpy`.

### Change 3

I changed the `mongo_spec` to start/stop the mongodb server in `before(:context)`/`after(:context)` hooks, as starting mongodb in particular takes a few seconds. This way, the two specs can run without having to each start mongodb.

### DIFF
--- a/lib/elastic_apm/spies/mongo.rb
+++ b/lib/elastic_apm/spies/mongo.rb
@@ -55,10 +55,13 @@ module ElasticAPM
         end
 
         def build_context(event)
+          # Some MongoDB commands are not on collections but rather db admin commands.
+          # For these commands, the value at the `command_name` key is the integer 1.
+          collection = event.command[event.command_name] unless event.command[event.command_name] == 1
           Span::Context.new(
             db: {
               instance: event.database_name,
-              statement: nil,
+              statement: ("#{collection}.#{event.command_name}" if collection),
               type: 'mongodb',
               user: nil
             }
--- a/lib/elastic_apm/spies/mongo.rb
+++ b/lib/elastic_apm/spies/mongo.rb
@@ -63,7 +63,7 @@ module ElasticAPM
           Span::Context.new(
             db: {
               instance: event.database_name,
-              statement: event.command.inspect,
+              statement: event.command.to_s,
               type: 'mongodb',
               user: nil
             }
--- a/lib/elastic_apm/spies/mongo.rb
+++ b/lib/elastic_apm/spies/mongo.rb
@@ -36,6 +36,8 @@ module ElasticAPM
 
         def push_event(event)
           return unless ElasticAPM.current_transaction
+          # Some MongoDB commands are not on collections but rather db admin commands.
+          # For these commands, the value at the `command_name` key is the integer 1.
           collection = event.command[event.command_name] unless event.command[event.command_name] == 1
           name = [event.database_name, collection, event.command_name].compact.join('.')
 
@@ -57,9 +59,6 @@ module ElasticAPM
         end
 
         def build_context(event)
-          # Some MongoDB commands are not on collections but rather db admin commands.
-          # For these commands, the value at the `command_name` key is the integer 1.
-          collection = event.command[event.command_name] unless event.command[event.command_name] == 1
           Span::Context.new(
             db: {
               instance: event.database_name,
--- a/lib/elastic_apm/spies/mongo.rb
+++ b/lib/elastic_apm/spies/mongo.rb
@@ -36,10 +36,12 @@ module ElasticAPM
 
         def push_event(event)
           return unless ElasticAPM.current_transaction
+          collection = event.command[event.command_name] != 1 && event.command[event.command_name]
+          name = [event.database_name, collection, event.command_name].compact.join('.')
 
           span =
             ElasticAPM.start_span(
-              event.command_name.to_s,
+              name,
               TYPE,
               context: build_context(event)
             )
@@ -61,7 +63,7 @@ module ElasticAPM
           Span::Context.new(
             db: {
               instance: event.database_name,
-              statement: (collection && "#{collection}.#{event.command_name}"),
+              statement: event.command.inspect,
               type: 'mongodb',
               user: nil
             }
--- a/lib/elastic_apm/spies/mongo.rb
+++ b/lib/elastic_apm/spies/mongo.rb
@@ -36,7 +36,7 @@ module ElasticAPM
 
         def push_event(event)
           return unless ElasticAPM.current_transaction
-          collection = event.command[event.command_name] != 1 && event.command[event.command_name]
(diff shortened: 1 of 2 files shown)


lib/elastic_apm/spies/mongo.rb
```diff
@@ -55,10 +55,13 @@ def pop_event(event)
...
 
         def build_context(event)
+          # Some MongoDB commands are not on collections but rather db admin commands.
+          # For these commands, the value at the `command_name` key is the integer 1.
+          collection = event.command[event.command_name] unless event.command[event.command_name] == 1
           Span::Context.new(
             db: {
               instance: event.database_name,
-              statement: nil,
+              statement: ("#{collection}.#{event.command_name}" if collection),
```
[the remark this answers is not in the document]
estolfo: No problem, that makes sense as well.

spec/elastic_apm/spies/mongo_spec.rb
```diff
@@ -4,9 +4,13 @@
 
 module ElasticAPM
   RSpec.describe 'Spy: MongoDB' do
-    around do |ex|
+    before(:all) do
```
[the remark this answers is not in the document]
estolfo: done

spec/elastic_apm/spies/mongo_spec.rb
```diff
@@ -48,8 +54,35 @@ def start_mongodb
       expect(db.user).to be nil
 
       client.close
+    end
 
-      ElasticAPM.stop
+    it 'instruments commands on collections', :intercept do
+
```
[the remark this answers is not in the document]
estolfo: done

axw:
We have a long-running initiative to align the agents on span names: https://docs.google.com/spreadsheets/u/1/d/1SmWeX5AeqUcayrArUauS_CxGgsjwRgMYH4ZY8yQsMhQ/edit?usp=asogm&ouid=116323289420264606750. Part of this is proposal is to include the db/collection in the span name.

In the Go agent I'm capturing the full command body in MongoDB Extended JSON format: https://github.com/elastic/apm-agent-go/blob/bfee6541ec2013567d21692a0e3088c3a50a92fa/module/apmmongo/monitor_integration_test.go#L109-L110.

Do you see any issues with going that route instead? The statement bit isn't strictly necessary right now if you update the span names instead.

estolfo:
@mikker Would changing the span name require a major release? It sounds like it would be a backwards-breaking change.

@axw I like the idea of using the command definition as the db statement, as it align nicely with the `SequelSpy`. There was `nil` in the db statement field before for the `MongoSpy` so we have full freedom to decide what to put there now. I will also need to check that we have character limits on the db statements.

mikker:
I suppose it would be, yes.

Db statements are truncated at 10k chars – https://github.com/elastic/apm-agent-ruby/blob/master/lib/elastic_apm/transport/serializers/span_serializer.rb#L53

lib/elastic_apm/spies/mongo.rb
```diff
@@ -61,7 +63,7 @@ def build_context(event)
           Span::Context.new(
             db: {
               instance: event.database_name,
-              statement: (collection && "#{collection}.#{event.command_name}"),
+              statement: event.command.inspect,
```
mikker: Why `.inspect`?

spec/elastic_apm/spies/mongo_spec.rb
```diff
@@ -66,20 +66,54 @@ def start_mongodb
...
+      expect(span.name).to eq 'elastic-apm-test.testing.parallelCollectionScan'
       expect(span.type).to eq 'db.mongodb.query'
       expect(span.duration).to_not be_nil
 
       db = span.context.db
       expect(db.instance).to eq 'elastic-apm-test'
       expect(db.type).to eq 'mongodb'
-      expect(db.statement).to eq 'testing.find'
+      # ParallelCollectionScan doesn't send 'lsid' in the command so we can validate the entire command document.
+      expect(db.statement).to eq "#{BSON::Document.new('parallelCollectionScan' => 'testing', 'numCursors' => 1)}"
```
mikker: Let's just have the final string in the spec?

lib/elastic_apm/spies/mongo.rb
estolfo: Otherwise it's a `Hash`. If I'm not mistaken, the `SpanSerializer` assumes the `db.statement` is a `String` [here](https://github.com/elastic/apm-agent-ruby/blob/master/lib/elastic_apm/transport/serializers/span_serializer.rb#L53).
Calling `#inspect` changes the `BSON::Document` to Extended JSON format.

spec/elastic_apm/spies/mongo_spec.rb
estolfo: Sure. I originally put this in case `BSON::Document` serialization to a String did something special.

lib/elastic_apm/spies/mongo.rb
estolfo: I should probably just call `#to_s` though...

spec/elastic_apm/spies/mongo_spec.rb
mikker: I guess we'd want to know if it did (or started doing)

estolfo: Good point

lib/elastic_apm/spies/mongo.rb
estolfo: I've changed the line to use `event.command.to_s`

estolfo:
jenkins, retest this please

PR #484, opened 2019-08-15
Central Config: Once more for Jenkins

mikker:
Same as elastic/apm-agent-ruby#464 but on new branch

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -224,6 +224,8 @@ Whether or not to attach `ENV` from Rack to transactions and errors.
 Enable {kibana-ref}/agent-configuration.html[APM Agent Configuration via Kibana].
 If set to `true`, the client will poll the APM Server regularly for new agent configuration.
 
+Usually APM Server determines how often to poll, but if not the default interval is 5 minutes.
+
 NOTE: This feature requires APM Server v7.3 or later and that the APM Server is configured with `kibana.enabled: true`.
 
 [float]
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -214,6 +214,19 @@ Whether or not to attach the request headers to transactions and errors.
 Whether or not to attach `ENV` from Rack to transactions and errors.
 
 [float]
+[[config-central-config]]
+==== `central_config`
+|============
+| Environment                  | `Config` key     | Default
+| `ELASTIC_APM_CENTRAL_CONFIG` | `central_config` | `true`
+|============
+
+Enable {kibana-ref}/agent-configuration.html[APM Agent Configuration via Kibana].
+If set to `true`, the client will poll the APM Server regularly for new agent configuration.
+
+NOTE: This feature requires APM Server v7.3 or later and that the APM Server is configured with `kibana.enabled: true`.
+
+[float]
 [[config-custom-key-filters]]
 ==== `custom_key_filters`
 [options="header"]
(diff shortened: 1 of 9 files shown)


lib/elastic_apm/central_config.rb
```diff
@@ -0,0 +1,141 @@
...
+      resp = perform_request
+
+      case resp.status
+      when 200..299
+        resp
+      when 300..399
+        resp
+      when 400..499
+        raise ClientError, resp
+      when 500..599
```
estolfo: Maybe you want to make these Ranges constants with names describing each scenario so new Ranges are not created each time this method is called? Could be slightly more efficient.

estolfo: Update: I checked `Concurrent::Promise` and it saves `fetch_config` as a proc so the only benefit of using constants would be for descriptiveness, not efficiency.

lib/elastic_apm/central_config.rb
```diff
@@ -0,0 +1,141 @@
...
+      update.each_key do |key|
+        @modified_options[key] ||= config.get(key.to_sym)&.value
+      end
+
+      # If the new update doesn't set a previously modified option,
+      # revert it to the original
+      @modified_options.each_key do |key|
+        next if update.key?(key)
+        update[key] = @modified_options.delete(key)
+      end
```
estolfo: Maybe add one more test case when an option is set to nil just to make sure the existence of keys is always checked in the conditionals and not `hash[key]`?

mikker: I’m not sure I follow?

docs/configuration.asciidoc
```diff
@@ -213,6 +213,19 @@ Whether or not to attach the request headers to transactions and errors.
...
+[float]
+[[config-central-config]]
+==== `central_config`
+|============
+| Environment                  | `Config` key     | Default
+| `ELASTIC_APM_CENTRAL_CONFIG` | `central_config` | `true`
+|============
+
+Enable {kibana-ref}/agent-configuration.html[APM Agent Configuration via Kibana].
+If set to `true`, the client will poll the APM Server regularly for new agent configuration.
```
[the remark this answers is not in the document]
mikker: Sure! It’s determined actually by APM Server, but the default is 5 mins so we could mention that 👍

lib/elastic_apm/central_config.rb
mikker: Good call to check!

I went a bit back and forth of whether I wanted to be more explicit with the status codes we expect. We know what apm server might return. We are already explicit further down with 304 Not Changed. Idk. This is probably fine as is.

lib/elastic_apm/central_config.rb
estolfo: I was thinking about the scenario if someone wanted to "clear" out a setting they already modified. So using the example in the specs, setting ` { transaction_sample_rate: 0.5 }` then ` { transaction_sample_rate: nil }` 
I have only a surface-level understanding of this CentralConfig feature so perhaps that scenario is not relevant or worth testing.

mikker: I think this would happen by removing the key in Kibana. So _un-setting_ means it's not sent anymore. On the contrary I could imagine someone in the future wanting to set an option to `nil` and how would we handle this if `nil` means _un-set_?

estolfo: If it's not possible now to set an option to `nil` then I guess testing for that is not necessary = )

PR #485, opened 2019-08-15
ci(jenkins): codecov should happen once, for ruby:2.6

v1v:
## Highlights
- Matrix downstream job does run codecov for each run, let's run only once.

## Test Cases
- Without coverage

image

- With coverage

image

## Tasks
- [ ] Revert https://github.com/elastic/apm-agent-ruby/pull/485/files/d63bf53fa1056c84a80506fe3738f086fe544025#diff-8db37e8fcea0f1a8f2f39667e94ebcc4R208

### DIFF
--- a/.ci/.jenkins_codecov.yml
+++ b/.ci/.jenkins_codecov.yml
@@ -1,0 +1,7 @@
+---
+
+RUBY:
+  - ruby:2.6
+
+FRAMEWORK:
+  - rails-5.2
--- a/.ci/.jenkins_codecov.yml
+++ b/.ci/.jenkins_codecov.yml
@@ -2,4 +2,4 @@
 # Tuple of ruby version and frameworks which are in charge to send data to codecov.
 # # is the separator between the ruby version and the framework.
 ENABLED:
-  - ruby:2.6#rails-5.2
+  - ruby:2.6#rails-6.0.0
--- a/.ci/.jenkins_codecov.yml
+++ b/.ci/.jenkins_codecov.yml
@@ -1,7 +1,5 @@
 ---
-
-RUBY:
-  - ruby:2.6
-
-FRAMEWORK:
-  - rails-5.2
+# Tuple of ruby version and frameworks which are in charge to send data to codecov.
+# # is the separator between the ruby version and the framework.
+ENABLED:
+  - ruby:2.6#rails-5.2
--- a/.ci/.jenkins_codecov.yml
+++ b/.ci/.jenkins_codecov.yml
@@ -2,4 +2,4 @@
 # Tuple of ruby version and frameworks which are in charge to send data to codecov.
 # # is the separator between the ruby version and the framework.
 ENABLED:
-  - ruby:2.6#rails-6.0.0
+  - ruby:2.6#rails-6.0
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -205,7 +205,7 @@ def runJob(rubyVersion, boolean isCoverage=false){
     branch = env.CHANGE_TARGET
   }
 
-  build( job: "apm-agent-ruby/apm-agent-ruby-downstream/${branch}",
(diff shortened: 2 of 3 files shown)


mikker:
Nice! Will this still report it for every version of Rails/Sinatra (the framework axis of the matrix)?

v1v:
> Nice! Will this still report it for every version of Rails/Sinatra (the framework axis of the matrix)?

Indeed, it's not ideal, but the current implementation does care only for the ruby versions in the `.ci/Jenkinsfile` and the frameworks in the `.ci/downstreamTests.groovy`

I'm tempted to move that specific behavior to another file which contains the match of ruby-versions#frameworks that the coverage should be expected to, then `.ci/downstreamTests.groovy`
should read the file and ensure the matches.

So roughly something like:

```
cat .ci/.jenkins_coverage.yml
TUPLE:
  - ruby:2.6#rails-5.2
```

```
cat .ci/downstreamTests.groovy
...
     # read YAML file .ci/.jenkins_coverage.yml
     if (data[TUPLE].contains("${x}#") && data[TUPLE].contains("#${y}")) {
        steps.codecov(repo: "${steps.env.REPO}", basedir: "${steps.env.BASE_DIR}", secret: "${steps.env.CODECOV_SECRET}")
     }
...
```

What do you think?

.ci/Jenkinsfile
```diff
@@ -197,19 +198,20 @@ def runBenchmark(version){
...
+def runJob(rubyVersion, boolean isCoverage=false){
   def branch = env.BRANCH_NAME
 
   if (env.CHANGE_ID) {
     branch = env.CHANGE_TARGET
   }
-  
-  build( job: "apm-agent-ruby/apm-agent-ruby-downstream/${branch}", 
+
+  build( job: "apm-agent-ruby/apm-agent-ruby-downstream/PR-485",
```
v1v: This is only required for testing purposes as it does use the master branch as the target branch by default.

mikker:
I think the extra file is a good idea 👍. One single source of truth.

.ci/.jenkins_codecov.yml
```diff
@@ -0,0 +1,5 @@
+---
+# Tuple of ruby version and frameworks which are in charge to send data to codecov.
+# # is the separator between the ruby version and the framework.
+ENABLED:
+  - ruby:2.6#rails-5.2
```
mikker: Sorry, Rails 6 came out in the meantime, so should probably be that instead 😆

v1v:
```suggestion
  - ruby:2.6#rails-6.0.0
```

.ci/.jenkins_codecov.yml
```diff
@@ -0,0 +1,5 @@
+---
+# Tuple of ruby version and frameworks which are in charge to send data to codecov.
+# # is the separator between the ruby version and the framework.
+ENABLED:
+  - ruby:2.6#rails-6.0.0
```
v1v:
```suggestion
  - ruby:2.6#rails-6.0
```

.ci/Jenkinsfile
v1v:
```suggestion
  build( job: "apm-agent-ruby/apm-agent-ruby-downstream/${env.BRANCH_NAME}",
```

.ci/downstreamTests.groovy
```diff
@@ -160,3 +162,14 @@ def runScript(Map params = [:]){
     }
   }
 }
+
+/**
+* Whether the given ruby version and framework are in charge of sending the
+* codecov results. It does require the workspace.
+*/
+def isCodecovEnabled(ruby, framework){
```
mdelapenya: The `Enabled` part could be misleading compared to the javadocs, wdyt?

v1v: Do you mean `isCodecov` rather than `isCodecovEnabled`?

mikker:
💙💚💛💜❤️

PR #487, opened 2019-08-15
Add Ruby VM specific metrics

mikker:
Continued from elastic/apm-agent-ruby#437 because Jenkins

### DIFF
--- a/lib/elastic_apm/metrics/vm.rb
+++ b/lib/elastic_apm/metrics/vm.rb
@@ -4,12 +4,18 @@ module ElasticAPM
   module Metrics
     # @api private
     class VM
-      def initialize(_config)
+      def initialize(config)
         @total_time = 0
+        @disabled = false
       end
+
+      attr_reader :config
+      attr_writer :disabled
 
       # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
       def collect
+        return if disabled?
+
         stat = GC.stat
         thread_count = Thread.list.count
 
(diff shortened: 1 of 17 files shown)


spec/elastic_apm_spec.rb
```diff
@@ -14,7 +14,10 @@
   end
 
   context 'when running', :mock_intake do
-    before { ElasticAPM.start }
+    before do
+      # disable immediate metrics collection
+      ElasticAPM.start(metrics_interval: 0)
```
mikker: @estolfo Not really a fan but it solves the JRuby test failures. The delegation tests are already a bit weird (although they have discovered errors before.)

_Wdyt?_

mikker:
@estolfo Merging this as it was already reviewed and approved from @axw in its original PR.

The last many commits were just fighting with either JRuby or Jenkins which was quite a bit frustrating. Perhaps we can reevaluate this fix later for example https://github.com/elastic/apm-agent-ruby/pull/487/files#diff-6ad466b0c60404c7ae066e0168d3e910R53-R61

Testing it in a production app over the weekend and if everything looks good, doing a minor release on Monday.

lib/elastic_apm/metrics/vm.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module ElasticAPM
+  module Metrics
+    # @api private
+    class VM
+      include Logging
+
+      def initialize(config)
```
estolfo: The `config` argument is not captured. Is that intentional? `config` is an `attr_reader` so I'm assuming it should be captured.

mikker: IIRC, the `Logging` module expects a `config` method to be available. It's something that I eventually want to rewrite.

mikker: ... but it should of course be saved to an ivar, yes, good catch!

mikker: https://github.com/elastic/apm-agent-ruby/pull/519

PR #491, opened 2019-08-17
ci(jenkins): parentstream job should trigger the specific downstream jobs

v1v:
### Highlights
- Parentstream used to launch the `target` branch downstream job.
- This solution is trying to avoid the issues when merging PRs with the target branch as we get quite often some errors with `refusing to merge unrelated histories`


v1v:
I'd rather prefer to close this PR as the MERGE_TARGET parameter is still enabled and might confuse just a bit.

v1v:
Reopen it, and removed the discrepancy with the input parameter which it's not required anymore.

Issue #496, opened 2019-08-20
[automation][apm-ci] Allow failures for the sinatra master and rails master


v1v:
## Requirements
- Currently, the pipeline does not distinguish whether the test failure is coming from a stable ruby/framework version or an RC/experimental/development version.
- The idea, is to avoid fail fast when running the tests for the sinatra-master and rails-master, and notify in case there were test failures but keep going.

## Questions.
- The Jenkins pipeline does provide a build status for the whole build or a specific stage. A build is a set of stages, so, in case the stage for the sinatra-master fails, do we want to just unstable/fail the stage or the whole build? From the UI the visualization would be either `yellow` or `red` for the whole build, or just the stage icon.

PR #499, opened 2019-08-22
ci(jenkins): decouple test frameworks in stable and master

v1v:
Closes https://github.com/elastic/apm-agent-ruby/issues/496

## Highlights
- Split the tests in those which are based on stable versions and those which are based on the master branch.
- Tests will run in two sequential buckets, the first bucket for the stable versions and the second one with the master branches.
- The reason for splitting into two stages is to help with the visualization of the pipeline in the UI, we could potentially run more in parallel if required but the UI won't be so helpful, IMO.
- If master test bucket fails then it won't fail the build/pipeline but notify the build status as unstable. This will help to keep the attention of failures in the stable test frameworks and identify easily when a master test framework got issues

### DIFF
--- a/.ci/.jenkins_framework.yml
+++ b/.ci/.jenkins_framework.yml
@@ -7,6 +7,3 @@ FRAMEWORK:
 
   - sinatra-2.0
   - sinatra-1.4
-
-  - rails-master
-  - sinatra-master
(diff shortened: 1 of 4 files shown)


mikker:
💙💚💛💜❤️

Issue #385, opened 2019-04-15
Add support for global labels


axw:
See https://github.com/elastic/apm-agent-go/issues/507

Should `default_tags` be deprecated, and superseded by global labels? Note that global labels is intended to be implemented using stream-level metadata, which won't be available until 7.1.

mikker:
Yes, my plan was to transition default_tags to global_labels whenever it would make sense and the labels were in APM Server.

PR #505, opened 2019-08-23
Support Global labels

estolfo:
Add support for global labels.
Closes https://github.com/elastic/apm-agent-ruby/issues/385

1. Support `ELASTIC_APM_GLOBAL_LABELS` configuration option. Global labels are in the format `key=value[,key=value[,...]]`. Only accepted in APM server >= 7.2 so `default_tags` are still supported, as long as the server accepts `tags` in events' `context`. Will be deprecated at a later point.
2. Include configured `global_labels` in the `labels` field of a `metadata` object.
3. Documentation

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -259,6 +259,10 @@ Add default tags to add to every transaction.
 
 WARNING: Be aware that tags are indexed in Elasticsearch. Using too many unique keys will result in *https://www.elastic.co/blog/found-crash-elasticsearch#mapping-explosion[Mapping explosion]*.
 
+NOTE: `global_labels` are supported as of APM server version 7.2. `default_tags` will eventually be
+deprecated to please transition to using `global_labels` instead. In the meantime, any `default_tags`
+that are set will override `global_labels`.
+
 [float]
 [[config-disable-send]]
 ==== `disable_send`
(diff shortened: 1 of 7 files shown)


docs/configuration.asciidoc
```diff
@@ -259,6 +259,10 @@ Add default tags to add to every transaction.
 
 WARNING: Be aware that tags are indexed in Elasticsearch. Using too many unique keys will result in *https://www.elastic.co/blog/found-crash-elasticsearch#mapping-explosion[Mapping explosion]*.
 
+NOTE: `global_labels` are supported as of APM server version 7.2. `default_tags` will eventually be
+deprecated to please transition to using `global_labels` instead. In the meantime, any `default_tags`
```
mikker: _deprecated **so** please_, typo

docs/configuration.asciidoc
```diff
@@ -348,9 +352,22 @@ otherwise defaults to `nil`.
 | `ELASTIC_APM_FRAMEWORK_VERSION` | `framework_version` | Depending on framework
 |============
 
-Version number of the used framework.
-For Ruby on Rails and Sinatra, this defaults to the used version of the framework,
-otherwise, the default is `nil`.
+[float]
```
mikker: I think you deleted the body of the option above by accident?

lib/elastic_apm/transport/serializers/metadata_serializer.rb
```diff
@@ -10,7 +10,8 @@ def build(metadata)
             metadata: {
               service: build_service(metadata.service),
               process: build_process(metadata.process),
-              system: build_system(metadata.system)
+              system: build_system(metadata.system),
```
mikker: Sidenote: Do you prefer dangling commas? I see a lot of smart people preferring them to minimise git churn like this one. I see both sides, I guess. Curious if you've ever contemplated the issue?

estolfo: I'm fine with it, if you'd like. We can also add it to rubocop to enforce it.

docs/configuration.asciidoc
estolfo: 😬

lib/elastic_apm/transport/serializers/metadata_serializer.rb
mikker: I don't think we should change anything now. Just curious for your thoughts 😊

PR #507, opened 2019-08-26
Return Error object from report

mikker:
Suggested in https://discuss.elastic.co/t/make-elasticapm-report-return-an-error-or-error-id-instead-of-true/196504.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Changed
 
-- `ElasticAPM.report` and `ElasticAPM.report_message` now return the generated `Error` objects ([#]())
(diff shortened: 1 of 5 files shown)


greyblake:
Thanks for that quick reply and for the pull request.
I'd like to let you know that at the moment `ElasticAPM::Error` is marked with `@api private`.
Since we're know exposing `Error`, should `@api private` be removed as well? 
(I am not sure what's correct, but I'd like to bring your attention to this)

mikker:
You are right that we probably should make the resource objects public, like `Error`, `Transaction`, `Span` perhaps even `MetricSet` and `Metadata`. That does make the scope of this PR much larger though.

greyblake:
@mikker Would it be OK for you then just to return `error.id` ?
This would be enough to satisfy my use case (rendering error id in response body) at the same time we can keep `Error` private.

mikker:
That's a very nice alternative to keep this PR small 👍

greyblake:
@mikker Thank you!

PR #513, opened 2019-09-02
feat: disable shallow cloning when using downstream pipelines

v1v:
## Highlights
- Shallow cloning with a mergeTarget causes` the refusing to merge unrelated histories` issue.
- Avoid shallow cloning then.


v1v:
As we don't use the master branch but the specific PR pipeline, then this particular PR is not required anymore

PR #518, opened 2019-09-05
Set config_file from ENV before loading

### LINKED ISSUE #517: Cannot change config path via ELASTIC_APM_CONFIG_FILE environment variable
Hello,
Looks like it is impossible to change default config file path via `ELASTIC_APM_CONFIG_FILE` environment variable. 
I confirmed that environment variable is correctly passed to process via inspecting it  via `/proc/<PID>/environment`.

Instrumented application is more or less standard rails 5.2 app.

mikker:
Fixes elastic/apm-agent-ruby#517

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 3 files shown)


mikker:
@axw Tagged you because Emily is away. Small change 💃

mikker:
jenkins, test this please

Issue #409, opened 2019-05-07
Provide public api for Railtie code


mikker:
For when users want to do some sort of dynamic loading of the agent, see for example https://github.com/elastic/apm-agent-ruby/issues/404#issuecomment-489553386

PR #522, opened 2019-09-10
Extract Rails setup code into its own module so it can be called without Railtie

estolfo:
The `start` method originally in the railtie is extracted into a `ElasticAPM::Rails` module so it can be called explicitly. The user will have to require `elastic_apm/rails` in order to call `ElasticAPM::Rails.start(config)`.

Note that all references to `Rails` in the codebase must refer to the module from the `rails` gem as `::Rails`. 
Simply writing `Rails` would refer to the `ElasticAPM::Rails` module.

Closes elastic/apm-agent-ruby#409

### DIFF
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -31,8 +31,10 @@ ElasticAPM.start(server_url: 'http://localhost:8200')
   options.  See <<configuration,Configuration>>.
 
 If you are using <<getting-started-rails,Ruby on Rails>> this is done
-automatically for you.
-If not see <<getting-started-rack,Getting started with Rack>>.
+automatically for you. If you skip requiring the `elastic_apm` gem and want to start the
+agent and hook into Rails manually, see the section on
+<<rails-start,hooking into Rails explicitly.>>
+If you're not using Rails, see <<getting-started-rack,Getting started with Rack>>.
 
 [float]
 [[api-agent-stop]]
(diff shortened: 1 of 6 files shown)


estolfo:
todo:
add documentation

estolfo:
@bmorelli25 I've made some minor documentation changes in the last commit. If you could take a look, that'd be great. Thanks!
cc/ @mikker

docs/api.asciidoc
```diff
@@ -31,8 +31,10 @@ ElasticAPM.start(server_url: 'http://localhost:8200')
   options.  See <<configuration,Configuration>>.
 
 If you are using <<getting-started-rails,Ruby on Rails>> this is done
-automatically for you.
-If not see <<getting-started-rack,Getting started with Rack>>.
```
bmorelli25: What about something like:

```
If you are using <<getting-started-rails,Ruby on Rails>> this is done automatically for you.
If you choose not to require the `elastic_apm` gem and instead want to start the
agent and hook into Rails manually, see <<rails-start,hooking into Rails explicitly>>.
If you're not using Rails, see <<getting-started-rack,Getting started with Rack>>.
```

bmorelli25: The middle sentence was just a little tough for me to digest. Other than that, LGTM

estolfo:
Thanks @bmorelli25 
I think this is all set to merge. Let me know if you object, @mikker

PR #525, opened 2019-09-12
Fix CentralConfig polling

mikker:
As this is a bug I think we should just keep this PR to this and get it out of the door. Agree, @estolfo?


mikker:
@estolfo Ready for review, let's get it out as a patch release. CI error is only that weird one I mentioned on Slack.

PR #528, opened 2019-09-16
Resolve block binding/class definition race condition

estolfo:
This is an attempt to resolve an occasional error on JRuby:
```
NoMethodError: undefined method `enqueue' for #<ElasticAPM::Agent:0xc9518ee>
 # ./lib/elastic_apm/agent.rb:77:in `block in initialize'
```

### DIFF
--- a/lib/elastic_apm/agent.rb
+++ b/lib/elastic_apm/agent.rb
@@ -43,7 +43,9 @@ module ElasticAPM
           return
         end
 
-        @instance = new(config).start
+        @instance = allocate
+        @instance.send(:initialize, config)
+        @instance.start
       end
     end
     # rubocop:enable Metrics/MethodLength
@@ -70,8 +72,11 @@ module ElasticAPM
 
       @central_config = CentralConfig.new(config)
       @transport = Transport::Base.new(config)
-      @instrumenter = nil
-      @metrics = nil
+      @instrumenter = Instrumenter.new(
+        config,
+        stacktrace_builder: stacktrace_builder
+      ) { |event| enqueue event }
+      @metrics = Metrics.new(config) { |event| enqueue event }
     end
 
     attr_reader(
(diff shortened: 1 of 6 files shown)


mikker:
An issue could be threads as the `initialize` is called inside a mutex lock in `self.start` and these new ones might be called at the same time from different threads?

estolfo:
The `instrumenter` and `metrics` objects are initialized lazily when the `start` method is called, which is still in the mutex [here](https://github.com/elastic/apm-agent-ruby/blob/master/lib/elastic_apm/agent.rb#L46)
You bring up a good point - for example, one obscure issue could be that someone would call `current_transaction` on the agent from another thread, which would in turn initialize the instrumenter at the same time that the `start` method is being called. So this isn't a good solution...I'll investigate more.

estolfo:
As a note..the tests have passed so it's likely the JRuby error is due to a race condition between the file being loaded and the block binding being defined.

mikker:
You could use the same lock as in `self.start` to lock the lazy initialization. Similar to the quite verbose but safe `start`:
```ruby
def metrics
  return @metrics if @metrics

  MUTEX.lock
    @metrics = Metrics.new(...)
  end
end
```

estolfo:
You wouldn't be able to use the same lock from `self.start` though because you'd get a recursive locking error.

estolfo:
I tried another approach - allocating the object and then initializing it directly. Let's see if it has an effect...

mikker:
Only if `start` references `metrics`?

estolfo:
The class `start` method calls the `start` method on the `@instance` within the lock and `metrics` would be referenced from there.

mikker:
fwiw, I used to pass the agent along in the initializers. Moved to the _callback_ based approach to not have to make an agent for them when testing and decouple a bit. Worth moving back if it fixes it.

estolfo:
@mikker Makes complete sense - both designs are valid. Let's see what the JRuby team says on my issue and decide from there how we can work around it.

mikker:
This one can be closed too?

estolfo:
yes, I'll follow up on the JRuby ticket

PR #529, opened 2019-09-17
ci(jenkins): when for tags requires pattern but branch doesn't allow regex

v1v:
### Highlights
- Enable param to skip tests.
- branch in the when doesn't allow using regex. See https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/351


v1v:
This has been solved already with:
- https://github.com/elastic/apm-agent-ruby/pull/530

PR #531, opened 2019-09-18
Split span.type into type, subtype, action

mikker:
_Upgrades_ legacy values by splitting by `'.'` if no `subtype` is passed.

Currently, APM Server updates the dot-separated types for us, so this will not change how data looks in ES.

When we do breaking changes, I think we should make the top-level `type` argument in `start_span` into a keyword argument. If you agree, @estolfo, perhaps we should deprecate it now in a minor version to warn users in a cycle before the upgrade.

This is sort-of needed for the breakdown graphs too, so rebasing that branch on this.

### DIFF
--- a/lib/elastic_apm/normalizers/active_record.rb
+++ b/lib/elastic_apm/normalizers/active_record.rb
@@ -9,10 +9,13 @@ module ElasticAPM
       class SqlNormalizer < Normalizer
         register 'sql.active_record'
 
+        TYPE = 'db'
+        ACTION = 'sql'
+
         def initialize(*args)
           super
 
-          @type = format('db.%s.sql', lookup_adapter || 'unknown').freeze
+          @subtype = lookup_adapter || 'unknown'
           @summarizer = SqlSummarizer.new
         end
 
@@ -22,7 +25,7 @@ module ElasticAPM
           name = summarize(payload[:sql]) || payload[:name]
           context =
             Span::Context.new(db: { statement: payload[:sql], type: 'sql' })
-          [name, @type, context]
+          [name, TYPE, @subtype, ACTION, context]
         end
 
         private
--- a/lib/elastic_apm/span.rb
+++ b/lib/elastic_apm/span.rb
@@ -14,17 +14,26 @@ module ElasticAPM
 
     DEFAULT_TYPE = 'custom'
 
-    # rubocop:disable Metrics/ParameterLists
+    # rubocop:disable Metrics/ParameterLists, Metrics/MethodLength
     def initialize(
       name:,
       transaction_id:,
       trace_context:,
       type: nil,
+      subtype: nil,
+      action: nil,
       context: nil,
       stacktrace_builder: nil
     )
       @name = name
-      @type = type || DEFAULT_TYPE
+
+      if subtype.nil? && type&.include?('.')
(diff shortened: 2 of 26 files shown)


axw:
@mikker I think this is a breaking change, since the server won't split apart type prior to [6.7.0](https://www.elastic.co/guide/en/apm/server/current/release-notes-6.7.0.html). So your statement "Currently, APM Server updates the dot-separated types for us, so this will not change how data looks in ES." only holds for ES 6.7.0+

mikker:
How about making this change send them as dot-separated to not be breaking but keep them split internally. Join them in the serializer. So, when we do the breaking changes, we only need to _not_ re-concatenate them in the serializer?

That way I can still build upon them when doing the breakdown metrics.

axw:
@mikker sounds like a plan to me :+1:

lib/elastic_apm/normalizers/active_record.rb
```diff
@@ -9,10 +9,13 @@ module ActiveRecord
...
         register 'sql.active_record'
 
+        TYPE = 'db'
+        ACTION = 'sql'
+
         def initialize(*args)
           super
 
-          @type = format('db.%s.sql', lookup_adapter || 'unknown').freeze
+          @subtype = lookup_adapter || 'unknown'
```
estolfo: you're no longer calling `format` on the result of this line or freezing the string. Is that intentional/ok?

lib/elastic_apm/span.rb
```diff
@@ -14,28 +14,50 @@ class Span
...
+      :trace_context,
+      :type
+    )
+    attr_reader(
+      :context,
+      :duration,
+      :stacktrace,
+      :timestamp,
+      :transaction_id
+    )
```
estolfo: thank you for alphabetizing 😉

lib/elastic_apm/transport/serializers/span_serializer.rb
```diff
@@ -66,6 +66,12 @@ def build_http(http)
             }
           end
         end
+
+        private
+
+        def join_type(span)
+          [span.type, span.subtype, span.action].tap(&:compact!).join('.')
```
estolfo: Why do you call `tap` before `compact!`?

lib/elastic_apm/span.rb
mikker: you noticed 😆

lib/elastic_apm/transport/serializers/span_serializer.rb
mikker: `compact!` doesn't return `self` so I step around its return value. I used `compact!` to modify the original and skip allocating an additional array on each call. But with this I'm allocating a block instead. Don't know what's worse. I think I'll make this less fancy.

lib/elastic_apm/normalizers/active_record.rb
mikker: Honestly not sure why we were freezing it before. The string has to be built every time anyway.

lib/elastic_apm/transport/serializers/span_serializer.rb
estolfo: Oh, I understand it now. The way it's written is fine with me if you'd like to keep it.

PR #535, opened 2019-09-23
Mock transport object instead of #enqueue method on Agent

### LINKED ISSUE #523: Increase JRuby spec reliability
The JRuby branches of the CI test runs fail too often.

My current best theory is that the specs start throwing requests at the test apps before the testapp's agent has fully booted, thereby, sometimes, missing the first request.

But we should investigate more.

estolfo:
Closes #523 
Resolve random `NotMethodError` on `enqueue`.

### DIFF
--- a/spec/support/intercept.rb
+++ b/spec/support/intercept.rb
@@ -33,20 +33,19 @@ RSpec.configure do
   end
 end
 
-RSpec.shared_context 'intercept', shared_contex: :metadata do
+RSpec.shared_context 'intercept' do
   let(:config) { ElasticAPM::Config.new }
   let(:agent) { ElasticAPM.agent }
+  let(:intercepted) { Intercept.new }
   before(:each) { start_intercepted_agent(config) }
 
   after(:each) do
     ElasticAPM.stop
-    @intercepted = nil
   end
 
   def start_intercepted_agent(config = ElasticAPM::Config.new)
-    @intercepted = Intercept.new
     ElasticAPM.start(config).tap do |agent|
-      allow(agent).to receive(:transport).and_return(@intercepted)
+      allow(agent).to receive(:transport).and_return(intercepted)
     end
   end
 end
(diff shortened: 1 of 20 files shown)


spec/support/intercept.rb
```diff
@@ -31,13 +31,22 @@ def submit(obj)
     def start; end
     def stop; end
   end
+end
 
-  config.before :each, intercept: true do
-    @intercepted = Intercept.new
-    allow(ElasticAPM::Transport::Base).to receive(:new).and_return(@intercepted)
-  end
+RSpec.shared_context 'intercept', shared_contex: :metadata do
```
mikker: TIL about `shared_context`

mikker: missing a _t_ in the second `shared_contex:`

mikker:
Can you perhaps briefly explain why this is/works different(ly)?

spec/support/intercept.rb
estolfo: A `shared_context` allows you to create a scope to include in a test context and only interact with the objects within that context. Previously, we were stepping around a context in the Rspec configuration and making a global change (with `allow_any_instance_of`) that was bleeding into other tests. The `NoMethodError` on the agent object was indicative of one test's cleanup getting in the way of another test's execution.
Keeping the mocking to a context and its scope should avoid these problems. In the proposed changes, there is an `agent` object local to the context that is mocked to return a `@intercepted` object when `#transport` is called. These changes to the `agent` object will not affect other tests so we shouldn't see the strange `NoMethodError` anymore.

All that said, the tests are still failing and I'm not 100% convinced this is the solution yet. It's just a pattern I've typically seen used with Rspec. I'll hopefully have it working soon! If not, it's back to the drawing board ✏️

mikker:
Do we want to close this brave effort now that we went for the shorter one?

estolfo:
I think we should still try to use `shared_context`s in the future but if the current changes unblock us, we can table this for now.

Issue #412, opened 2019-05-07
Rename `disabled_spies` to `disable_instrumentations`


mikker:
Per https://github.com/elastic/apm/issues/85

PR #539, opened 2019-09-25
Rename disabled_spies to disabled_instrumentations

mikker:
Warns and redirects, non-breaking.

Closes elastic/apm-agent-ruby#412

### DIFF
--- a/lib/elastic_apm/config.rb
+++ b/lib/elastic_apm/config.rb
@@ -16,6 +16,7 @@ module ElasticAPM
   # @api private
   class Config
     extend Options
+    extend Deprecations
 
     DEPRECATED_OPTIONS = %i[
       compression_level=
@@ -160,6 +161,31 @@ module ElasticAPM
       end
     end
 
+    def use_ssl?
+      server_url.start_with?('https')
+    end
+
+    def collect_metrics?
+      metrics_interval > 0
+    end
+
+    def span_frames_min_duration?
+      span_frames_min_duration != 0
+    end
+
+    def span_frames_min_duration=(value)
+      super
+      @span_frames_min_duration_us = nil
+    end
+
+    def span_frames_min_duration_us
+      @span_frames_min_duration_us ||= span_frames_min_duration * 1_000_000
+    end
+
+    def inspect
+      super.split.first + '>'
+    end
+
     # DEPRECATED
     # rubocop:disable Metrics/MethodLength
     def capture_body=(value)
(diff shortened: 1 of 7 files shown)


lib/elastic_apm/config.rb
```diff
@@ -185,6 +186,18 @@ def capture_body=(value)
...
+    def disabled_spies=(list)
+      warn 'The option disabled_spies has been renamed to ' \
+        'disabled_instrumentations'
+      self.disabled_instrumentations = list
+    end
+
+    def enabled_spies
+      warn 'enabled_spies has been renamed to enabled_instrumentations'
+      enabled_instrumentations
+    end
```
estolfo: Is there a reason you didn't use the `deprecate` helper?

mikker: I guess not, no 👍

lib/elastic_apm/config.rb
```diff
@@ -49,7 +50,7 @@ class Config
     option :default_tags,                      type: :dict,   default: {}
     option :disable_send,                      type: :bool,   default: false
     option :disable_start_message,             type: :bool,   default: false
-    option :disabled_spies,                    type: :list,   default: %w[json]
+    option :disabled_instrumentations,         type: :list,   default: %w[json]
```
estolfo: In not including `:disabled_spies` in this list, aren't we dropping support for the `ELASTIC_APM_DISABLED_SPIES` environment variable as a result? I'm guessing we don't want to drop support until we actually remove the deprecated methods/option.

spec/elastic_apm/config_spec.rb
```diff
@@ -42,7 +42,7 @@ module ElasticAPM
         ['ELASTIC_APM_VERIFY_SERVER_CERT', 'true', true],
         ['ELASTIC_APM_VERIFY_SERVER_CERT', '0', false],
         ['ELASTIC_APM_VERIFY_SERVER_CERT', 'false', false],
-        ['ELASTIC_APM_DISABLED_SPIES', 'json,http', %w[json http]],
+        ['ELASTIC_APM_DISABLED_INSTRUMENTATIONS', 'json,http', %w[json http]],
```
estolfo: And here wouldn't we want to keep the test for `ELASTIC_APM_DISABLED_SPIES` as the option hasn't been fully removed yet?

lib/elastic_apm/config.rb
mikker: Nice catch!

spec/elastic_apm/config_spec.rb
mikker: Added a dedicated example in the deprecations section

PR #541, opened 2019-09-25
Allow non-String tag values

estolfo:
This is a breaking change - tag values were previously always converted to a string. This PR allows non-String tag values, while still truncating String values.

### DIFF
--- a/lib/elastic_apm.rb
+++ b/lib/elastic_apm.rb
@@ -361,7 +361,7 @@ module ElasticAPM # rubocop:disable Metrics/ModuleLength
     # Set a _tag_ value for the current transaction
     #
     # @param key [String,Symbol] A key
-    # @param value [Object] A value (will be converted to string)
+    # @param value [Object] A value
     # @return [Object] The given value
     def set_tag(key, value)
       agent&.set_tag(key, value)
--- a/lib/elastic_apm/instrumenter.rb
+++ b/lib/elastic_apm/instrumenter.rb
@@ -195,7 +195,7 @@ module ElasticAPM
       return unless current_transaction
 
       key = key.to_s.gsub(/[\."\*]/, '_').to_sym
-      current_transaction.context.tags[key] = value.to_s
+      current_transaction.context.tags[key] = value
     end
 
     def set_custom_context(context)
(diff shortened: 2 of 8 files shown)


lib/elastic_apm/instrumenter.rb
```diff
@@ -195,7 +195,7 @@ def set_tag(key, value)
       return unless current_transaction
 
       key = key.to_s.gsub(/[\."\*]/, '_').to_sym
```
mikker: We could save a string allocation by using `gsub!` instead. Could be too insignificant to worry about. How does `gsub!` even work with `frozen_string_literal`? Questions, questions.

mikker:
Approved, still needs
- [ ] docs
- [ ] changelog entry

estolfo:
@mikker what docs are needed?

lib/elastic_apm/instrumenter.rb
estolfo: You can't call `gsub!` on a frozen string, unfortunately. I can't think of another way to do this without creating another String. Let me know if you have any ideas...

estolfo: We could do it if we wrote this in a c extension ; )

lib/elastic_apm.rb
```diff
@@ -361,7 +361,7 @@ def report_message(message, context: nil, **attrs)
     # Set a _tag_ value for the current transaction
     #
     # @param key [String,Symbol] A key
-    # @param value [Object] A value (will be converted to string)
```
axw: should we perhaps leave `set_tag` alone, and make this change on `set_label` in #543? `set_tag` would call `set_label(key, value.to_s)`?

mikker: I like this idea.

mikker:
@estolfo `docs/api.asciidoc` is one. Perhaps do project-wide search, see if we mention `set_tag` anywhere else.

lib/elastic_apm.rb
estolfo: Yes, this is great! That way, it won't be a breaking change to allow boolean and numerical tag values. Thanks for the suggestion @axw

estolfo:
Changes have been added to PR #543

PR #543, opened 2019-09-25
Rename set_tag to set_label

### LINKED ISSUE #466: Rename set_tag to set_label
See https://github.com/elastic/apm/issues/42

estolfo:
Changes included in this PR:

- Add method `set_label` on `ElasticAPM` and deprecate `set_tag` with a warning
- Add ENV option `ELASTIC_APM_DEFAULT_LABELS` and deprecate `ELASTIC_APM_DEFAULT_TAGS`
- Add config option `default_labels`
- Override `default_tags` method to set `default_labels` internally. 
- Rename all `set_tag` private methods to `set_label`
- Rename all object attributes `tags` to `labels`
- Continue to serialize a context with a `tags` key but the value is its `labels` attribute.
- Raise an exception if both `default_labels` and `default_tags` are set in the config.

Close #466

Update: 
Merged in PR #541 to this one. It allows non-String values to be passed to `set_label` but preserves the behavior of `set_tag`.

### DIFF
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -260,7 +260,7 @@ Returns `[String]` ID of the generated `[ElasticAPM::Error]` object.
 
 Add a tag to the current transaction.
 Tags are basic key-value pairs that are indexed in your Elasticsearch database
-and therefore searchable.
+and therefore searchable. The value will always be converted to a String.
 
 TIP: Before using custom tags, ensure you understand the different types of
 {apm-overview-ref-v}/metadata.html[metadata] that are available.
@@ -280,6 +280,36 @@ Arguments:
 Returns the set `value`.
 
 WARNING: Be aware that tags are indexed in Elasticsearch. Using too many unique keys will result in *https://www.elastic.co/blog/found-crash-elasticsearch#mapping-explosion[Mapping explosion]*.
+
+NOTE: This method has been deprecated in favor of `set_label`, which does not convert values to Strings.
+
+[float]
+[[api-agent-set-label]]
+==== `ElasticAPM.set_label`
+
+Add a label to the current transaction.
+Labels are equivalent to what were previously called `tags`. They are basic key-value pairs that
+are indexed in your Elasticsearch database and therefore searchable.
+The value can be any type and will not be converted to Strings.
+
+TIP: Before using custom labels, ensure you understand the different types of
+{apm-overview-ref-v}/metadata.html[metadata] that are available.
+
+[source,ruby]
+----
+before_action do
+  ElasticAPM.set_label(:company_id, current_user.company.id)
+end
+----
+
+Arguments:
+
+  * `key`: A string key. Note that `.`, `*` or `"` will be converted to `_`.
+  * `value`: A value.
+
+Returns the set `value`.
+
+WARNING: Be aware that labels are indexed in Elasticsearch. Using too many unique keys will result in *https://www.elastic.co/blog/found-crash-elasticsearch#mapping-explosion[Mapping explosion]*.
 
 [float]
 [[api-agent-set-custom-context]]
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -289,7 +289,7 @@ NOTE: This method has been deprecated in favor of `set_label`, which does not co
 
 Add a label to the current transaction.
 Labels are basic key-value pairs that are indexed in your Elasticsearch database and therefore searchable.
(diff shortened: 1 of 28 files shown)


mikker:
What happens if `set_tag(:key, { object: true })` or  `set_tag(:key, User.new)`?

Do we want to, instead of checking for a string, flip it and check for Booleans and Numericals? These are the three allowed types, so if passed something else, should we then `.to_s` on that?

```ruby
case value
when Number, Integer, Float, MoreNumberClassesPerhapsAGeneralSuperClass, TrueClass, FalseClass, NilClass, AllTheThingsWeExplicitlySupport
  value
else
  value.to_s
end 
```

estolfo:
But in the past, we just called `to_s` on the tag value, so aren't these changes consistent with that? We called `to_s` [here](https://github.com/elastic/apm-agent-ruby/blob/master/lib/elastic_apm/instrumenter.rb#L198). Let me know if I'm missing something.

mikker:
Sure, I meant for `set_label`. We don't fall back `.to_s` there and that worried me a bit.

I guess I'd like the behaviour to be predictable. That's why I asked what happens when you pass it non-trivial values like `User.new`. If we explicitly handle the value types that we know we support and convert everything else `to_s`, I think it's more predictable.

estolfo:
Isn't it up to the user to pass a value that can be represented in Elasticsearch? Or is the agent typically strict with input?

estolfo:
On second thought, I think I understand - I'm thinking from a client perspective. A user would get immediate (error) feedback if they were using a client and initiating requests directly. With the agent, if the user sets the tag to a type that is not accepted, it will wreak havoc silently when the event is sent to the server.

mikker:
Exactly. There's not necessarily any immediate feedback. Better to put the "wrong" data in ES by calling `to_s` on a user because at least something shows up and they can fix it themselves. Nothing showing up or data dropping is harder to notice/debug.

estolfo:
Makes sense 👍🏼

docs/api.asciidoc
```diff
@@ -281,6 +281,36 @@ Returns the set `value`.
...
 WARNING: Be aware that tags are indexed in Elasticsearch. Using too many unique keys will result in *https://www.elastic.co/blog/found-crash-elasticsearch#mapping-explosion[Mapping explosion]*.
 
+NOTE: This method has been deprecated in favor of `set_label`, which does not convert values to Strings.
+
+[float]
+[[api-agent-set-label]]
+==== `ElasticAPM.set_label`
+
+Add a label to the current transaction.
+Labels are equivalent to what were previously called `tags`. They are basic key-value pairs that
```
mikker: I'm not sure we need the historical reference here. Users just want to know what they can do now and probably don't care much about how things used to be? (I could be wrong?)

lib/elastic_apm.rb
```diff
@@ -363,8 +364,29 @@ def report_message(message, context: nil, **attrs)
...
+    # @param key [String,Symbol] A key
+    # @param value [Object] A value
+    # @return [Object] The given value
+    def set_label(key, value)
+      case value
+      when TrueClass,
+           FalseClass,
+           Numeric,
+           NilClass,
+           String
```
mikker: It's so beautiful!

lib/elastic_apm/config.rb
```diff
@@ -210,6 +211,10 @@ def inspect
       super.split.first + '>'
     end
 
+    def default_labels
```
mikker: Could we use the same strategy as elsewhere in Config of overriding `def default_tags=()` and `deprecate :default_tags=, :default_labels=`?

This merges on every lookup which is once per txn/span. Overriding the setter, it'd only be called once on boot.

mikker:
I wanted to _Comment_ and not _Approve_. Too tired 😆

docs/api.asciidoc
estolfo: Sure, I'll take out the overexplanation = )

lib/elastic_apm/config.rb
estolfo: Making this work properly ended up being a bit tricky. I decided that instead of merging tags and labels that were set, I would always favor whichever was set last. But setting `default_tags` sets `default_labels` internally.

estolfo:
@mikker the latest changes include raising an exception for both `default_labels` and `default_tags` being set. If you're ok with this, the PR is ready for a final review.

PR #548, opened 2019-10-01
ci(jenkins): HOME env variable does not exist in the CI

v1v:
## Highlights
- Unfortunately, the `HOME` env variable is not defined in the CI.
- Map the path to the HOME CI User.


v1v:
I wanna confirm the docker volumes are now working as expected.

image


I'll revert the commit with the tests

PR #549, opened 2019-10-02
ci(jenkins): Reduce number of requested workers

v1v:
## Highlights
- Reuse the top-level worker as much as possible.
- Used to take 5 minutes to get to the `test` stage and now 2 minutes in the main pipeline.
- Used to take 3 minutes to get to the `test` stage and now 1.5 minutes in the downstream pipeline.
- Minor fix in the git reset --hard, as the GIT_BASE_COMMIT is the right env variable, created on the fly, with the commit sha either for the TAG/Branch/PR. GIT_COMMIT only works for branches.

## Test Cases
- Every single stage should run but the release stage.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -47,7 +47,6 @@ pipeline {
     Checkout the code and stash it, to use it on other stages.
     */
     stage('Checkout') {
-      agent { label 'immutable' }
       options { skipDefaultCheckout() }
       steps {
         deleteDir()
@@ -56,7 +55,6 @@ pipeline {
       }
     }
     stage('Sanity checks') {
-      agent { label 'linux && immutable && docker' }
       when {
         beforeAgent true
         anyOf {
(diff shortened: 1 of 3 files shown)


.ci/Jenkinsfile
```diff
@@ -202,7 +172,8 @@ pipeline {
                       sshagent(['f6c7695a-671e-4f4f-a331-acdce44ff9ba']) {
                         sh '.ci/prepare-git-context.sh'
                         sh 'gem install rake yard rspec'
-                        sh 'rake release'
+                        sh 'touch file'
```
mikker: Disabling it for now?

v1v: I'll revert https://github.com/elastic/apm-agent-ruby/pull/549/commits/20ba9c1ab94f3a3226fb8db1cac8843856ebc152 

I just wanted to validate every single stage does behave as expected and no weird chmod or similar privileges for certains files could affect the following stages

PR #550, opened 2019-10-02
Use monotonic time to calculate durations

### LINKED ISSUE #208: Investigate monotonic time
Should we use instead of `Time`?
https://blog.dnsimple.com/2018/03/elapsed-time-with-ruby-the-right-way/

mikker:
This changes a public api method in OpenTracing (`end_time` -> `end_clock`) however we could perhaps add a workaround.

I think I'd like to add a bit of explicitness to the `travel` method too from `mock_time`. It's not clear what unit we are traveling by (microseconds, now) and that might confuse a bit.

Closes #208

### DIFF
--- a/lib/elastic_apm/opentracing.rb
+++ b/lib/elastic_apm/opentracing.rb
@@ -65,10 +65,10 @@ module ElasticAPM
       end
       # rubocop:enable Lint/UnusedMethodArgument
 
-      def finish(end_time: Time.now)
+      def finish(end_clock: Util.monotonic_micros)
         return unless (instrumenter = ElasticAPM.agent&.instrumenter)
 
-        elastic_span.done end_time: Util.micros(end_time)
+        elastic_span.done end_clock: end_clock
 
         case elastic_span
         when ElasticAPM::Transaction
--- a/lib/elastic_apm/opentracing.rb
+++ b/lib/elastic_apm/opentracing.rb
@@ -65,10 +65,10 @@ module ElasticAPM
       end
       # rubocop:enable Lint/UnusedMethodArgument
 
-      def finish(end_clock: Util.monotonic_micros)
+      def finish(clock_end: Util.monotonic_micros)
         return unless (instrumenter = ElasticAPM.agent&.instrumenter)
 
-        elastic_span.done end_clock: end_clock
+        elastic_span.done clock_end: clock_end
 
         case elastic_span
         when ElasticAPM::Transaction
--- a/lib/elastic_apm/opentracing.rb
+++ b/lib/elastic_apm/opentracing.rb
@@ -65,8 +65,15 @@ module ElasticAPM
       end
       # rubocop:enable Lint/UnusedMethodArgument
 
-      def finish(clock_end: Util.monotonic_micros)
+      # rubocop:disable Metrics/MethodLength
+      def finish(clock_end: Util.monotonic_micros, end_time: nil)
         return unless (instrumenter = ElasticAPM.agent&.instrumenter)
+
+        if end_time
+          warn '[ElasticAPM] DEPRECATED: Setting a custom end time as a ' \
+            '`Time` is deprecated. Use `clock_end:` and monotonic time instead.'
+          clock_end = end_time
+        end
 
         elastic_span.done clock_end: clock_end
 
@@ -79,6 +86,7 @@ module ElasticAPM
 
         instrumenter.enqueue.call elastic_span
       end
+      # rubocop:enable Metrics/MethodLength
 
       private
 
--- a/lib/elastic_apm/span.rb
+++ b/lib/elastic_apm/span.rb
@@ -61,18 +61,19 @@ module ElasticAPM
 
     # life cycle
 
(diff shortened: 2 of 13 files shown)


spec/elastic_apm/span_spec.rb
```diff
@@ -77,12 +81,12 @@ module ElasticAPM
...
 
         expect(subject).to be_stopped
-        expect(subject.duration).to be 100_000
+        expect(subject.duration).to be 100
       end
     end
 
     describe '#done', :mock_time do
-      let(:duration) { 100 }
+      let(:duration_us) { 5_100 }
```
estolfo: what does `us` refer to?

lib/elastic_apm/span.rb
```diff
@@ -61,18 +61,19 @@ def initialize(
 
     # life cycle
 
-    def start(timestamp = Util.micros)
-      @timestamp = timestamp
-
+    def start(start_clock = Util.monotonic_micros)
+      @timestamp = Util.micros
+      @clock = start_clock
```
estolfo: `@clock` is a bit ambiguous of a name, as it's actually the starting micros. I don't necessarily have any better ideas though.
Maybe `@start_micros`?

spec/support/mock_time.rb
```diff
@@ -1,15 +1,21 @@
 # frozen_string_literal: true
 
 RSpec.configure do |config|
-  config.around :each, mock_time: true do |example|
+  config.before :each, mock_time: true do
     @mocked_date = Time.utc(1992, 1, 1)
+    @mocked_clock = 123_456_000_000
 
     def travel(distance)
```
estolfo: `travel_micros`?
Does the agent use any other time measurements? If not, then I think it's ok to leave the name as `travel`, as all time units will be assumed to be microseconds. If there are other time references using something other than microseconds, we might want to be more explicit about what time unit this method assumes.

spec/elastic_apm/span_spec.rb
mikker: it's _micro_ seconds. The u is an ASCII version of a μ char.

lib/elastic_apm/span.rb
mikker: It's referring to the _clock_ time so I think it makes sense to have clock in there. It might be because I'm differentiating it from what used to be but just _micros_ tells me it's related to ordinary `Time`.

mikker: Could call both the ivar and the argument `clock_start` instead.

spec/support/mock_time.rb
mikker: Yes, something like that. It's all μs internally until we serialise and convert to ms.

spec/elastic_apm/span_spec.rb
estolfo: Ok, so you added that because you wanted to distinguish between the units here and the other references to `duration` (like the accessor references in the `transaction_spec`  ?

mikker: Yes, I think I wanted it to purposefully look like just a value in the specs and not something coming from a trace.

mikker:
I'll add a fallback to the OpenTracing api changes.

lib/elastic_apm/opentracing.rb
```diff
@@ -65,10 +65,17 @@ def log_kv(timestamp: nil, **fields)
...
+      # rubocop:disable Metrics/MethodLength
+      def finish(clock_end: Util.monotonic_micros, end_time: nil)
         return unless (instrumenter = ElasticAPM.agent&.instrumenter)
 
-        elastic_span.done end_time: Util.micros(end_time)
+        if end_time
+          warn '[ElasticAPM] DEPRECATED: Setting a custom end time as a ' \
+            '`Time` is deprecated. Use `clock_end:` and monotonic time instead.'
+          clock_end = end_time
+        end
```
estolfo: Do we want to raise this deprecation warning in 3.0 or will this go into a minor on 2.x?

mikker: I think just 3.0. I don't think anyone will ever hit it so it can stay just fine until 4.0.

estolfo: ok because technically you don't _have_ to add a deprecation warning here by semver rules, as we're putting this in a major release.

mikker: Oh, we can remove it right away without a deprecation cycle in between? I guess we could. idk. I like to give people warnings and the _fix_ is so straight forward. Then again I didn't bother enough to add pointers to what monotonic time is, so `¯\_(ツ)_/¯`

wdyt?

estolfo: I doubt many people will get this warning so I guess we can just do whatever is a nicer experience for those few users (i.e. emit the deprecation warning)

mikker: We'll leave it like this then 👍

estolfo: lgtm then

PR #553, opened 2019-10-04
Log CentralConfig connection errors on debug level

mikker:
As Central Config comes enabled yet needs special setup in APM Server/Kibana, logging 404s every 5mins on the `error` log level seems a bit more critical than it is.

The Python logs http errors from Central Config on the `debug` level which is more sensible.


mikker:
jenkins, retest this please

PR #554, opened 2019-10-04
Config objects are not mutated and transactions have a local config

### LINKED ISSUE #536: Snapshot relevant config options at transaction start

estolfo:
This is the first in a series of 3 PRs involving how the config for an agent is set, updated, and accessed. 

The changes will be in three parts:
1) **→[This PR]** `CentralConfig` wraps a `Config` object and swaps it out for a new one when updates are received. A transaction gets a local `config` object that it refers to for the entirety of its lifetime. This achieves the functionality of "snapshotting" the config at transaction start.
2) All objects refer to `ElasticAPM.config` (forwarded to `ElasticAPM.agent.config`) when they need to reference a config. That way, all references to `config` will be the latest one, as managed by the `CentralConfig`.
3) The `Logging` module will reference `ElasticAPM.config` but allow the `#config` method to be overridden, in case an object needs to define a local one.

Closes #536

### DIFF
--- a/lib/elastic_apm/agent.rb
+++ b/lib/elastic_apm/agent.rb
@@ -231,6 +231,12 @@ module ElasticAPM
     def add_filter(key, callback)
       transport.add_filter(key, callback)
     end
+
+    def update_config(new_options)
+      @config = config.dup.tap do |_config|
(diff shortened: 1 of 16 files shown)


mikker:
Nice approach 👍

Issue #560, opened 2019-10-10
`rails middleware` throws an error


bguban:
Hey, I installed `elastic-apm (3.0.0)` into my app and now `rails middleware` throws an error:
```
...
/Users/bguban/.rvm/gems/ruby-2.6.4@price_manager/gems/bugsnag-6.12.1/lib/bugsnag/integrations/rake.rb:19:in `execute'
/Users/bguban/.rvm/gems/ruby-2.6.4@price_manager/gems/elastic-apm-3.0.0/lib/elastic_apm/spies/rake.rb:17:in `execute'
/Users/bguban/.rvm/gems/ruby-2.6.4@price_manager/gems/bugsnag-6.12.1/lib/bugsnag/integrations/rake.rb:19:in `execute'
/Users/bguban/.rvm/gems/ruby-2.6.4@price_manager/gems/elastic-apm-3.0.0/lib/elastic_apm/spies/rake.rb:17:in `execute'
/Users/bguban/.rvm/gems/ruby-2.6.4@price_manager/gems/bugsnag-6.12.1/lib/bugsnag/integrations/rake.rb:19:in `execute'
/Users/bguban/.rvm/gems/ruby-2.6.4@price_manager/gems/elastic-apm-3.0.0/lib/elastic_apm/spies/rake.rb:17:in `execute'
/Users/bguban/.rvm/gems/ruby-2.6.4@price_manager/gems/bugsnag-6.12.1/lib/bugsnag/integrations/rake.rb:19:in `execute'
/Users/bguban/.rvm/gems/ruby-2.6.4@price_manager/gems/elastic-apm-3.0.0/lib/elastic_apm/spies/rake.rb:17:in `execute'
/Users/bguban/.rvm/gems/ruby-2.6.4@price_manager/gems/bugsnag-6.12.1/lib/bugsnag/integrations/rake.rb:19:in `execute'
/Users/bguban/.rvm/gems/ruby-2.6.4@price_manager/gems/elastic-apm-3.0.0/lib/elastic_apm/spies/rake.rb:17:in `execute'
/Users/bguban/.rvm/gems/ruby-2.6.4@price_manager/gems/bugsnag-6.12.1/lib/bugsnag/integrations/rake.rb:19:in `execute'
/Users/bguban/.rvm/gems/ruby-2.6.4@price_manager/gems/elastic-apm-3.0.0/lib/elastic_apm/spies/rake.rb:17:in `execute'
[...]

Issue #379, opened 2019-04-10
spies should use Module.prepend vs alias


timogoebel:
Using `Module.prepend` is the default to override a method in Rails. Prior to that `alias_method_chain`.

We've been trying to use the APM with [Foreman](https://github.com/theforeman/foreman), which [overrides](https://github.com/theforeman/foreman/blob/b7d88a6c5c307312fae0404ceb1d5717e1cf313a/lib/foreman/http_proxy/net_http_extension.rb#L22) `net-http`'s `request` method as well as apm-agent-ruby. Foreman uses `Module.prepend`, [this gem](https://github.com/elastic/apm-agent-ruby/blob/87f9d2eaa3f9001fb120478ae4e7035b22be0527/lib/elastic_apm/spies/net_http.rb#L33) uses `alias`. And this leads to a `SystemStackError (stack level too deep)` exception.

As a workaround we currently disable the net-http spy in the config:
```yaml
disabled_spies:
  - json
  - net_http
```

Can we switch to `Module.prepend` as a default or make this configurable?

mikker:
Yes, we should definitely do this.

Thanks for bringing it up. I'll see to it when I find the time. PRs are of course welcome if you're impatient 😅

mcfearsome:
We are also experiencing this, but with the elasticsearch spy

mikker:
@mcfearsome Are you doing your own prepending? Or do you use a library that prepends?

mikker:
I reverted this _fix_ as it interfered with another library that didn't use Module.prepend.

I hope I'm wrong but I think we either choose to Module.prepend and then only work with libraries that use that too OR use `alias` and only work with other libraries that use that too.

bencrouse:
Hi @mikker, we're prepending on our own to add a query-caching middleware in our app.

I thought I'd check to see what a similar gem like New Relic does, but it seems like they're also using `alias`: https://github.com/newrelic/rpm/blob/c9a467d22e6d4b893be33be330d258a125cf813a/lib/new_relic/agent/instrumentation/net.rb

I took a quick look, but couldn't figure out when the `execute` block runs there, maybe the `defer`/`execute` stuff was written to get around this?

mikker:
I'm not familiar with their `defer` thing but I suspect it's their solution to something that we do as well: Modifying the constant if exists, otherwise add a _hook_ to do it when `require`d.

[...]

PR #561, opened 2019-10-11
Use prepend to load rake handler

bguban:
Fix for https://github.com/elastic/apm-agent-ruby/issues/560

### DIFF
--- a/lib/elastic_apm/spies/rake.rb
+++ b/lib/elastic_apm/spies/rake.rb
@@ -5,38 +5,38 @@ module ElasticAPM
   module Spies
     # @api private
     class RakeSpy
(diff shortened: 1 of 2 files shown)


mikker:
Hi @bguban! Thank you for contributing this. Ruby doesn't do well when you mix `prepend` and `alias` ([see](https://github.com/elastic/apm-agent-ruby/issues/379#issuecomment-500857044)) so I'm hesitant to merge this.

Could you perhaps fix your problems by changing the load order of your gems? To make it work with both implementations, make sure to load the one with  `Module#prepend` _after_ the one with `alias`.

mikker:
Closing this. Let me know if changing the load order fixes the problem for you – or if you want to discuss this further 😊 Thanks for the effort!

Issue #321, opened 2019-02-19
Add Grape support


mikker:
This has been requested several times and we should figure out an official way.

alilland:
👍

estolfo:
Completed by #562

PR #562, opened 2019-10-11
Grape support

### LINKED ISSUE #206: Implement Grape instrumentation using ActiveSupport::Notifications
Follows up: https://github.com/elastic/apm-agent-ruby/pull/119

estolfo:
This PR adds support for Grape. The changes are:

- Defining a `Grape.start` method so that the agent can be hooked up to a Grape app
- Normalizer for `endpoint_run`
- Set config framework values for Grape in `Grape.start`
- Documentation for Grape integration
- Add framework name and version to transaction context when a grape event is initiated by a Rails app.

Further work:
- ~Maybe add normalizer `format_response`~
- ~Consider not instrumenting the filter and validator endpoints, as they can be very noisy. There are a lot of internal filters and validators called and there's no way to distinguish between library and user-specified ones.~

Todo:
- [x] Test in Jenkins

Closes #206, #321


lib/elastic_apm/grape.rb
```diff
@@ -0,0 +1,46 @@
...
+  # Module for starting the ElasticAPM agent and hooking into Grape.
+  module Grape
+    extend self
+    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+    # Start the ElasticAPM agent and hook into Grape.
+    #
+    # @param app [Grape::API] A Grape app.
+    # @param config [Config, Hash] An instance of Config or a Hash config.
+    # @return [true, nil] true if the agent was started, nil otherwise.
+    def start(app, config)
```
estolfo: @mikker What do you think of this API? It centralizes the setting of config values into this file so it's not in the config (as we discussed).
However, it's inconsistent with `Rails.start` (which only accepts the `config` as a param)
I could also have it be `Grape.start(config, app: GrapeApp)`

What do you think?

mikker: I think I like it best like this. We'll also eventually have to do the same for a `Sinatra` helper that will also include the app as an argument. I see your point about being inconsistent with Rails but I don't think it's a problem. Most people(™) using Rails will probably not start the agent themselves anyway?

lib/elastic_apm/grape.rb
```diff
@@ -27,9 +25,18 @@ def start(app, config)
...
     end
+    # rubocop:enable Metrics/MethodLength, Metrics/AbcSize
 
     private
 
+    def set_config(app, config)
+      config.service_name ||= app.name
+      config.framework_name ||= 'Grape'
+      config.framework_version ||= ::Grape::VERSION
+      config.logger ||= app.logger
```
mikker: Can this be `nil`? Does Grape apps have a default logger?

lib/elastic_apm/grape.rb
estolfo: ok, I also like this API the best. I'll create a ticket to change the Rails API for 4.0. Good point about few people using `ElasitcAPM::Rails.start(..)`
I'll open a second PR to add a `ElasitcAPM::Sinatra.start(MyApp, config)` API, if you agree?

mikker: Yes please 🙏

lib/elastic_apm/grape.rb
estolfo: It says [here](https://github.com/ruby-grape/grape/blob/master/README.md#logging) that 

> Grape::API provides a logger method which by default will return an instance of the Logger class from Ruby's standard library.

So I think it's ok, no?

mikker: Yes, very ok 😊

estolfo:
@mikker btw, we already have documentation for `Rails.start` [here](https://www.elastic.co/guide/en/apm/agent/ruby/current/api.html#rails-start)

docs/api.asciidoc
```diff
@@ -193,16 +193,27 @@ Returns the built context.
...
 [source,ruby]
 ----
 ElasticAPM::Rails.start(server_url: 'http://localhost:8200')
 ----
 
+[float]
+[[grape-start]]
+=== Grape
+
+Start the agent and hook into Grape.
```
mikker: Perhaps we ought to mention _when_ this is supposed to be done? Before or after starting the app? Does it matter?

docs/configuration.asciidoc
```diff
@@ -22,7 +22,7 @@ secret_token: <%= ENV["VERY_SECRET_TOKEN"] %>
 Options are applied in the following order (last one wins):
 
 1. Defaults
-2. Arguments to `ElasticAPM.start` / `Config.new`
+2. Arguments to `ElasticAPM.start` / `ElasticAPM::Grape.start` / `ElasticAPM::Rails.start` / `Config.new`
```
mikker: Could _Arguments to `.start(…)`_ be explanatory enough? I'm afraid mentioning too many different frameworks here will confuse more than help?

docs/api.asciidoc
mikker: What if you run your Grape app as part of your Rails app (possible and common)? Then you don't have to explicitly start it. What happens to the config then?

docs/configuration.asciidoc
```diff
@@ -21,8 +21,9 @@ secret_token: <%= ENV["VERY_SECRET_TOKEN"] %>
 
 Options are applied in the following order (last one wins):
 
+1. Arguments to `ElasticAPM::<framework>.start`
```
mikker: I don't like this. Can't we keep it like it is? It wasn't _not_ true.

estolfo: You mean don't change this list at all to include the `ElasticAPM::framework.start` references?

mikker: Yeah, I think so. Or change it to just `.start` without the specific module/class.

Gemfile
```diff
@@ -33,24 +33,35 @@ else
...
-unless version =~ /^(master|6)/
-  gem 'delayed_job', require: nil
+if frameworks_versions.key?('rails')
+  unless frameworks_versions['rails'] =~ /^(master|6)/
+    gem 'delayed_job', require: nil
+  end
 end
 
+gem 'activerecord-jdbcsqlite3-adapter', platform: :jruby
+
```
estolfo: @mikker Is this ok with you? Now we can test two frameworks (ex: `FRAMEWORK=rails-6.0,grape-1.2`)

docs/api.asciidoc
estolfo: The second comment is now addressed by Grape events adding a framework name and version to transaction context when they are initiated by Rails.

docs/configuration.asciidoc
estolfo: So this?

1. Defaults
2. Arguments to `ElasticAPM.start` / `Config.new`
3. Config file eg. `config/elastic_apm.yml`
4. Environment variables
5. Arguments to a framwork's `.start`. Ex: `ElasticAPM::Grape.start`

lib/elastic_apm/middleware.rb
```diff
@@ -13,7 +13,7 @@ def initialize(app)
     # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
     def call(env)
       begin
-        if running? && !path_ignored?(env)
+        if running? && !path_ignored?(env) && !ElasticAPM.current_transaction
```
estolfo: @mikker I had to add this because otherwise the grape call to this method would try to create a transaction within a Rails transaction.

estolfo:
@mikker Since your approval, I've made significant changes to this PR. It'd be great if you could review the following:
* [ ] [The new Rails and Grape integration test](https://github.com/elastic/apm-agent-ruby/pull/562/files#diff-95c7bb1328199a9bdf8fefbba53e3ed2). It verifies that grape requests change the transaction context. They add the grape framework attributes even though the events are technically initiated by Rails.
* [ ] [The ability to install and require multiple frameworks](https://github.com/elastic/apm-agent-ruby/pull/562/files#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R36-R60) for testing so that rails and grape can be tested together in Jenkins.
* [ ] [Checking that there is no current transaction before creating one](https://github.com/elastic/apm-agent-ruby/pull/562/files#diff-8fb5c13ae8b624d09bf16821dd5b6820R16) in `ElasticAPM::Middleware`
* [ ] [Framework attributes are added to the transaction context](https://github.com/elastic/apm-agent-ruby/pull/562/files#diff-c5c77b8b4c1cfd6bb15feddd88e2b67fR17-R20) in the Grape endpoint event if the framework is set to Rails.
* [ ] Only transactions are created for Grape events. Should they be spans within the Rails transaction?

docs/api.asciidoc
estolfo: Addressing the first comment: you would put the code to start the agent and hook into Grape in config.ru or similar as is mentioned in [this comment](https://github.com/elastic/apm-agent-ruby/pull/562/files#diff-01c1a79a5a1e1b67003a1885365f2cf5R67)
I can add that clarification about config.ru here as well.

Gemfile
mikker: I like it a lot!

docs/configuration.asciidoc
mikker: But that's not true, is it?

What I meant:
1. Defaults
1. Arguments to `.start(...)`
1. Config file
1. ENV vars

lib/elastic_apm/middleware.rb
mikker: I think this looks completely fine 👍

spec/elastic_apm/grape_spec.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+if defined?(Grape)
+  require 'elastic_apm/grape'
```
mikker: Wouldn't this already be required by the agent if defined?

spec/integration/grape_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+          route_param :id do
+            get do
+              { status: params[:id] }
+            end
+          end
+        end
+      end
+
+      config = { api_request_time: '100ms',
+                 capture_body: 'all',
```
mikker: I don't think this is necessary unless we want to test that we actually capture the body

spec/integration/rails_grape_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    before :all do
+      class RailsGrapeTestApp < Rails::Application
+        config.logger = Logger.new(nil)
+        config.logger.level = Logger::DEBUG
+
+        config.eager_load = false
+
+        config.elastic_apm.api_request_time = '100ms'
+        config.elastic_apm.capture_body = 'all'
```
mikker: Same as other comment on `capture_body`

spec/integration/rails_grape_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+        config.elastic_apm.pool_size = Concurrent.processor_count
+        config.elastic_apm.service_name = 'RailsGrapeTestApp'
+        config.elastic_apm.log_path = 'spec/elastic_apm.log'
```
mikker: Unless we're testing this functionality, why configure it?

lib/elastic_apm/normalizers/grape/endpoint_run.rb
```diff
@@ -0,0 +1,34 @@
...
+        register 'endpoint_run.grape'
+
+        TYPE = 'app'
+        SUBTYPE = 'resource'
+
+        FRAMEWORK_NAME = 'Grape'
+
+        def normalize(transaction, _name, payload)
+          transaction.name = endpoint(payload[:env])
+          unless transaction.config.framework_name == FRAMEWORK_NAME
```
mikker: Do we need to check beforehand? If we were to get this twice (when would we?) it can set the same value twice, no worries 😊

mikker: Oh, it's checking against the config, I see. Sorry. Then it's fine to check as we'll same a few chars per transaction payload 👍

estolfo: I only want to add the transaction-level service info if it should differ from the metadata values.

spec/elastic_apm/grape_spec.rb
estolfo: Not with the way the gems were being listed in the Gemfile - when running `Bundler.require`, the agent was being required before the frameworks. I've now made [this change](https://github.com/elastic/apm-agent-ruby/pull/562/commits/f81df980351c9bd9a2ccb93982efc4894868e7bf#diff-8b7db4d5cc4b8f6dc8feb7030baa2478) so requiring the framework modules in tests is no longer necessary.

docs/configuration.asciidoc
estolfo: Yes, you're right. I keep confusing myself and forgetting I used `||=` in the `configure_app` method of the framework module. I've reverted my changes so it's back to the way it was.

docs/configuration.asciidoc
```diff
@@ -66,6 +66,23 @@ ElasticAPM::Sinatra.start(
 
 See <<getting-started-rack>>.
 
+[float]
+=== Grape and Rack
+
+When using APM with Grape and Rack you can configure it when starting
```
mikker: Do we want to be explicit about that this is only when you use Grape _without_ Rails? I might be over thinking it.

estolfo: No harm in being over-explicit. I've added a "(without Rails)" detail.

PR #566, opened 2019-10-14
Add ElasticAPM::Sinatra.start API

estolfo:
This allows users to explicitly start the agent + hook in their Sinatra app.

`ElasticAPM::Sinatra.start(MyApp, config)`

### DIFF
--- a/lib/elastic_apm/sinatra.rb
+++ b/lib/elastic_apm/sinatra.rb
@@ -11,7 +11,7 @@ module ElasticAPM
     # @return [true, nil] true if the agent was started, nil otherwise.
     def start(app, config)
       config = Config.new(config) unless config.is_a?(Config)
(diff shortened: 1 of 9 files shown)


lib/elastic_apm/sinatra.rb
```diff
@@ -0,0 +1,36 @@
...
+      ElasticAPM.start(config)
+      ElasticAPM.running?
+    rescue StandardError => e
+      config.logger.error format('Failed to start: %s', e.message)
+      config.logger.debug "Backtrace:\n" + e.backtrace.join("\n")
+    end
+
+    private
+
+    def set_config(app, config)
```
mikker: I don't think this method name describes what's going on. How about `set_app` or `configure_app` or something similar?

PR #567, opened 2019-10-15
Add API on event context to set framework name and version

estolfo:
This change is motivated by [this comment](https://github.com/elastic/apm-agent-ruby/pull/562/files#r334814742)

This PR adds a variable on a `Context` that represents `service` data, previously only ever set on `Metadata`. It adds a method `context#set_service` that would allow the Grape support implementation to set the framework and framework version to its own attributes. This is important for when Grape is used alongside Rails and the events need to be tracked as coming from Grape, not Rails.

### DIFF
--- a/lib/elastic_apm/context.rb
+++ b/lib/elastic_apm/context.rb
@@ -9,25 +9,36 @@ require 'elastic_apm/context/user'
 module ElasticAPM
   # @api private
   class Context
-    def initialize(custom: {}, labels: {}, user: nil)
+    def initialize(custom: {}, labels: {}, user: nil, service: nil)
       @custom = custom
       @labels = labels
       @user = user || User.new
+      @service = service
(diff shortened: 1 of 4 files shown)


estolfo:
I'm not 100% proud of this implementation @mikker 
It feels a bit messy so let me know if you have any other ideas.

mikker:
No grand ideas, I think this is fine.

I've long felt the context implementation is a bit of a mess already. It's sort of caught between convenience and performance.

This line for example https://github.com/elastic/apm-agent-ruby/pull/567/files#diff-6e7bae38ef769f09cdedc6f643f9f627R15
Ideally we wouldn't allocate anything unless there was a `User` to include. I feel like there's a bunch of places where we could lazy allocate things when necessary.

That said, I think this is fine as is 👍

PR #569, opened 2019-10-17
Remove trailing slash from server_url Config option

### LINKED ISSUE #568: Connect to ELASTIC_APM_SERVER_URL fails if contains a backslash at the end
**Describe the bug**

If you set ELASTIC_APM_SERVER_URL to “http://apm-server:8200/”, it would not work, for some reason, you have to remove the final backslash and set it to “http://apm-server:8200" then it works.

**To Reproduce**
1. Start Elastic Stack with APM
2. start the Opbeans-ruby with the ELASTIC_APM_SERVER_URL environment variable set to “http://apm-server:8200/”
3. Connect to the Opbeans-ruby to generate transactions
4. Check that there is nothing on the APM UI

**Expected behavior**
After generating some transactions on the Opbeans-ruby those should be sent to the APM server but are not.

estolfo:
This PR adds a converter `url` type that removes a trailing slash from the `server_url` config option.

Closes #568

### DIFF
--- a/lib/elastic_apm/config/options.rb
+++ b/lib/elastic_apm/config/options.rb
@@ -48,7 +48,7 @@ module ElasticAPM
           when :bool then normalize_bool(val)
           when :list then normalize_list(val)
           when :dict then normalize_dict(val)
-          when :url then normalize_trailing_slash(val)
+          when :url then normalize_url(val)
           else
             # raise "Unknown options type '#{type.inspect}'"
             val
@@ -71,7 +71,7 @@ module ElasticAPM
           Hash[val.split(/[&,]/).map { |kv| kv.split('=') }]
         end
 
-        def normalize_trailing_slash(val)
+        def normalize_url(val)
           return val unless val.is_a?(String)
           val.end_with?('/') ? val.chomp('/') : val
         end
--- a/lib/elastic_apm/config/options.rb
+++ b/lib/elastic_apm/config/options.rb
@@ -48,6 +48,7 @@ module ElasticAPM
           when :bool then normalize_bool(val)
           when :list then normalize_list(val)
           when :dict then normalize_dict(val)
+          when :url then normalize_trailing_slash(val)
           else
             # raise "Unknown options type '#{type.inspect}'"
             val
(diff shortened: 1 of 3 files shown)


lib/elastic_apm/config/options.rb
```diff
@@ -48,6 +48,7 @@ def normalize(val)
           when :bool then normalize_bool(val)
           when :list then normalize_list(val)
           when :dict then normalize_dict(val)
+          when :url then normalize_trailing_slash(val)
```
mikker: Maybe align with the other types and call it normalize_url

estolfo: good point, fixed!

lib/elastic_apm/config/options.rb
```diff
@@ -69,6 +70,11 @@ def normalize_dict(val)
           return val unless val.is_a?(String)
           Hash[val.split(/[&,]/).map { |kv| kv.split('=') }]
         end
+
+        def normalize_url(val)
+          return val unless val.is_a?(String)
```
mikker: I know this is merged already but why this check? Wouldn't we want to `ArgumentError` if given anything but a string?

mikker: Anything but a string as `server_url` will error somewhere else down the line.

estolfo: I was following the pattern of the other `normalize_*` methods. I think I should actually call `to_s` on it, as that is [what we were doing before](https://github.com/elastic/apm-agent-ruby/blob/master/lib/elastic_apm/config/options.rb#L45). What do you think?

mikker: The others make sense because they need to handle for example both an array and a string value, keeping the array if given one or converting into an array if given a string. This can only be passed and used as a string – and not, say, stored as a `URI`.

`to_s` instead is good 👍

mikker: With the current change, this would set `server_url` to a `URI` if given one, exploding later on when we append strings to it using `+`, etc...

estolfo: got it. So I'll change it to `to_s`

estolfo: https://github.com/elastic/apm-agent-ruby/pull/573

PR #570, opened 2019-10-17
Include stacktraces for successful, ActiveSupport::Notifications spans

estolfo:


mikker:
If this works with Grape then great. Did you try? I think I need more code or some examples of how it'll work to get a sense.

The problem with Rails and using `caller` in the Notifications hook is that the actual method isn't part of the call stack at that point so no matter how much we pop off then we won't get to the true sauce.

_Guessing_ the source location in Rails by the controller+action and following conventions _feels_ a bit reckless even though it could probably work in 99.9% of cases. Would love to see an experiment.

lib/elastic_apm/subscriber.rb
```diff
@@ -35,6 +37,9 @@ def start(name, id, payload)
 
       normalized = @normalizers.normalize(transaction, name, payload)
 
+      _caller = caller
```
estolfo: `caller` is a method, so we save the result in a variable to avoid calling it again.

estolfo:
Closing as this is not a good approach.

Issue #369, opened 2019-04-01
Support async spans inside transactions


mikker:
Closes https://github.com/elastic/apm-agent-ruby/issues/342

PR #571, opened 2019-10-18
Support starting more than one span on a transaction

estolfo:
The allows a span to be started in a thread with a parent from another thread.
It introduces a parm `parent` to `start_span` to achieve this.

It also exposes an attribute `sync` to be set when a span is created. The value for this field was previously *always* `true` but now the default is `nil` and it can be set by the user.

Closes elastic/apm-agent-ruby#369

### DIFF
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -155,7 +155,7 @@ Arguments:
 Returns the created span.
 
 If you'd like to create an asynchronous span, you have to pass the parent `Span` or `Transaction` to the `start_span` method.
-You can also set the `sync` flag to `false`, if you'd like the span to be marked as asynchronous.
+You can also set the `sync` flag to `false`, if you'd like the span to be marked as asynchronous. This has no effect other than being queryable in Elasticsearch.
 
 [source,ruby]
 ----
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -159,7 +159,7 @@ You can also set the `sync` flag to `false`, if you'd like the span to be marked
 
 [source,ruby]
 ----
-transaction = ElasticAPM.start_transaction
+transaction = ElasticAPM.current_transaction # or start one with `.start_transaction`
 Thread.new do
   ElasticAPM.start_span(
     'job 1',
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -147,10 +147,28 @@ Arguments:
   * `context`: An instance of `Span::Context`.
   * `include_stacktrace`: Whether or not to collect a Stacktrace.
   * `trace_context:`: An optional `TraceContext` object for Distributed Tracing.
+  * `parent:`: An optional parent transaction or span. Relevant when the span is created in another thread.
+  * `sync:`: An boolean to indicate whether the span is created synchronously or not.
   * `&block`: An optional block to wrap with the span.
   The block is passed the span as an optional argument.
 
 Returns the created span.
+
+If you'd like to create an asynchronous span, you have to pass the parent `Span` or `Transaction` to the `start_span` method.
+You can also set the `sync` flag to `false`, if you'd like the span to be marked as asynchronous.
+
+[source,ruby]
+----
+transaction = ElasticAPM.start_transaction
+Thread.new do
+  ElasticAPM.start_span(
+    'job 1',
+    parent: transaction,
+    sync: false
+  ) { Worker.perform }
+  ElasticAPM.end_span
+end
+----
 
 [float]
 [[api-agent-end_span]]
@@ -173,10 +191,27 @@ Arguments:
   * `context`: An instance of `Span::Context`.
   * `include_stacktrace`: Whether or not to collect a Stacktrace.
   * `trace_context:`: An optional `TraceContext` object for Distributed Tracing.
+  * `parent:`: An optional parent transaction or span. Relevant when the span is created in another thread.
+  * `sync:`: An boolean to indicate whether the span is created synchronously or not.
   * `&block`: An optional block to wrap with the span.
   The block is passed the span as an optional argument.
 
 Returns the return value of the given block.
+
+If you'd like to create an asynchronous span, you have to pass the parent `Span` or `Transaction` to the method.
+You can also set the `sync` flag to `false`, if you'd like the span to be marked as asynchronous.
+
+[source,ruby]
+----
+transaction = ElasticAPM.start_transaction
+Thread.new do
+  ElasticAPM.with_span(
+    'job 1',
+    parent: transaction,
+    sync: false
+  ) { Worker.perform }
+end
+----
 
 [float]
 [[api-agent-build-context]]
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -203,7 +203,7 @@ You can also set the `sync` flag to `false`, if you'd like the span to be marked
 
 [source,ruby]
 ----
-transaction = ElasticAPM.start_transaction
+transaction = ElasticAPM.current_transaction # or start one with `.start_transaction`
 Thread.new do
   ElasticAPM.with_span(
     'job 1',
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -198,7 +198,7 @@ Arguments:
 
 Returns the return value of the given block.
 
-If you'd like to create an asynchronous span, you have to pass the parent `Span` or `Transaction` to the method.
+If you'd like to create an asynchronous span, you have to pass the parent `Span` or `Transaction` to the `with_span` method.
 You can also set the `sync` flag to `false`, if you'd like the span to be marked as asynchronous.
 
 [source,ruby]
--- a/lib/elastic_apm.rb
+++ b/lib/elastic_apm.rb
@@ -226,7 +226,8 @@ module ElasticAPM
       action: nil,
       context: nil,
       include_stacktrace: true,
-      trace_context: nil
+      trace_context: nil,
+      parent: nil
     )
       unless block_given?
         raise ArgumentError,
@@ -244,7 +245,8 @@ module ElasticAPM
             action: action,
             context: context,
             include_stacktrace: include_stacktrace,
-            trace_context: trace_context
+            trace_context: trace_context,
+            parent: parent
           )
         yield span
       ensure
--- a/lib/elastic_apm.rb
+++ b/lib/elastic_apm.rb
@@ -180,7 +180,8 @@ module ElasticAPM
       context: nil,
       include_stacktrace: true,
       trace_context: nil,
-      parent: nil
+      parent: nil,
+      sync: nil
     )
       agent&.start_span(
         name,
@@ -189,7 +190,8 @@ module ElasticAPM
         action: action,
         context: context,
         trace_context: trace_context,
-        parent: parent
+        parent: parent,
+        sync: sync
       ).tap do |span|
         break unless span && include_stacktrace
         break unless agent.config.span_frames_min_duration?
@@ -227,7 +229,8 @@ module ElasticAPM
       context: nil,
       include_stacktrace: true,
       trace_context: nil,
-      parent: nil
+      parent: nil,
+      sync: nil
     )
       unless block_given?
         raise ArgumentError,
@@ -246,7 +249,8 @@ module ElasticAPM
             context: context,
             include_stacktrace: include_stacktrace,
             trace_context: trace_context,
-            parent: parent
+            parent: parent,
+            sync: sync
           )
         yield span
       ensure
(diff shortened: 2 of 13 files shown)


lib/elastic_apm.rb
```diff
@@ -167,15 +167,17 @@ def start_span(
       action: nil,
       context: nil,
       include_stacktrace: true,
-      trace_context: nil
+      trace_context: nil,
+      parent_transaction: nil
```
mikker: Why `parent_transaction` and not just `transaction`? The span is always related to only _one_ transaction and it'll always be its parent.

estolfo: It's to keep the naming consistent with the argument in the `instrumenter` method. In that method, I need a way to refer to a current_transaction, variable transaction, and argument transaction. [These lines](https://github.com/elastic/apm-agent-ruby/pull/571/files#diff-b907effb9782fac8cec4233c342822a6R149) could change so that I don't need to use a `transaction` variable.
```
transaction = parent_transaction || current_transaction
return unless transaction
return unless transaction.sampled?	
```

to

```
return unless(parent_transaction || current_transaction)
return unless(parent_transaction || current_transaction).sampled?	
```

estolfo: Do you have any other naming suggestions for how to distinguish between the argument transaction and current transaction?

mikker: I don't have any good suggestions unless you want me to dive further in?

My point was more from the perspective of the user. This is the top-most, outward facing API. And for the user, `parent_` is given – at worst confusing; eg _why wouldn't it be?_

I think we should start from there and then dance our way around juggling the different internal complexities afterwards if we have to. Makes sense?

estolfo: ok, I see your point. I'll make the argument `transaction` and see how I can handle it in the `instrumenter`.

estolfo: I changed the argument to `transaction`

mikker:
I like the simplicity of the general idea. I'm surprised that it didn't really take more than this.

What if a span's parent is a span – don't we have to have an argument for specifying a general parent txn/span in `ElasticAPM.start_span`?

```text
|--- txn -------------|
  |- span 1 --------|
    |- span 1a --|   <-┬- These two a nested under span 1
    |- span 1b -|    <-┘
```

axw:
> What if a span's parent is a span – don't we have to have an argument for specifying a general parent txn/span in ElasticAPM.start_span?

Would it make sense to instead introduce new `start_span` methods to the `Transaction` and `Span` classes, and have this method continue to just use the current span/transaction as the parent? That's what several other agents are doing.

estolfo:
that's a great idea @axw, I'll investigate doing that and see how it works with our api.

estolfo:
[I added a method](https://github.com/elastic/apm-agent-ruby/pull/571/files#diff-3ff0f7dfc745f9ac0268a337d9caf3e0R81-R97) on `transaction` called `start_span`.
I can't create the Span directly in the method and have to go through the `ElasticAPM.start_span` chain. This is because I need the `stacktrace_builder` object when instantiating a Span and this is only available in the `instrumenter`.
That means I have to keep the argument `transaction` on the `ElasticAPM.start_span` method.

estolfo:
> What if a span's parent is a span – don't we have to have an argument for specifying a general parent txn/span in ElasticAPM.start_span?

@mikker Would you mind showing me a code example so I'm clear on what you mean?

lib/elastic_apm/instrumenter.rb
```diff
@@ -143,32 +143,29 @@ def start_span(
...
-      transaction.inc_started_spans!
+      parent_transaction = transaction || current_transaction
+      return unless parent_transaction
+      return unless parent_transaction.sampled?
 
-      if transaction.max_spans_reached?
-        transaction.inc_dropped_spans!
-        return
-      end
+      parent = transaction || current_span || current_transaction
```
mikker: See my example here, https://github.com/elastic/apm-agent-ruby/pull/571#issuecomment-544682821. The `parent` is the span, but we're also passing a transaction (?) to the async `start_span` calls.

My point is, this line would set the parent to `txn` from the example where as it should be `span 1`?

lib/elastic_apm/instrumenter.rb
```diff
@@ -144,28 +144,35 @@ def start_span(
...
+          parent
+        else
+          current_transaction
+      end
+      return unless transaction
+      return unless transaction.sampled?
+      return unless transaction.inc_started_spans!
 
-      return unless parent_transaction.inc_started_spans!
+      parent = parent || current_span || current_transaction
```
mikker: `parent ||= ` 😊

lib/elastic_apm/instrumenter.rb
```diff
@@ -159,7 +159,7 @@ def start_span(
       return unless transaction.sampled?
       return unless transaction.inc_started_spans!
 
-      parent = parent || current_span || current_transaction
+      parent ||= current_span || current_transaction
```
mikker: I would `parent ||= (current_span || current_transaction)` around the result, so it's not mistaken for `(parent ||= current_span) || current_transaction`

lib/elastic_apm/instrumenter.rb
```diff
@@ -144,19 +144,23 @@ def start_span(
...
-        transaction.inc_dropped_spans!
-        return
+      transaction = case parent
+        when Span
+          parent.transaction
+        when Transaction
+          parent
+        else
+          current_transaction
       end
```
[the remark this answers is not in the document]
mikker: Suggestion:

```ruby
transaction =
  case parent
  when Span then parent.transaction
  when Transaction then parent
  else current_transaction
  end
```

lib/elastic_apm/transaction.rb
```diff
@@ -78,15 +79,14 @@ def done(result = nil, clock_end: Util.monotonic_micros)
...
-
-    def max_spans_reached?
-      started_spans > config.transaction_max_spans
+      MUTEX.synchronize do
+        @started_spans += 1
+        if @started_spans > config.transaction_max_spans
+          @dropped_spans += 1
+          return false
+        end
+      end
```
mikker:
```suggestion
      end

```

spec/elastic_apm_spec.rb
```diff
@@ -152,6 +152,97 @@
     after { ElasticAPM.stop }
   end
 
+  context 'async spans', :intercept do
+    context 'transaction parent' do
+      it 'allows async spans' do
+        with_agent do
+          transaction = ElasticAPM.start_transaction
+          span1 = Thread.new do
+            ElasticAPM.start_span('job 1', parent: transaction)
```
mikker: Do we want to `end_span` or `with_span` these so they close, just for good measure?

spec/elastic_apm_spec.rb
```diff
@@ -152,6 +152,97 @@
...
+            end.value
+            expect(span2.parent_id).to eq(span.trace_context.child.parent_id)
+
+            span3 = Thread.new do
+              ElasticAPM.start_span('job 2', parent: span)
+            end.value
+            expect(span3.parent_id).to eq(span.trace_context.child.parent_id)
+            span
+          end
+          transaction.done
```
mikker:
```suggestion
          transaction.done

```

spec/elastic_apm/transaction_spec.rb
```diff
@@ -103,30 +103,28 @@ module ElasticAPM
     end
 
     describe '#inc_started_spans!' do
-      it 'increments count' do
-        expect { subject.inc_started_spans! }
-          .to change(subject, :started_spans).by 1
+      let!(:result) { subject.inc_started_spans! }
```
mikker: Experience has taught me to stay away from `let!` with a bang. Personal preference only but I much prefer either using `before` or just inlining in each example with `result = ...`.

estolfo: I would advocate for using `let!` here because of these reasons
1) It's a construct of the `RSpec` framework, meant to be used exactly as it is here.
2) Setting local variables inside `it` block while there are test framework constructs to achieve the same functionality seems like an anti-pattern.
3) I'd like to have the two separate tests, one for the result of calling `subject.inc_started_spans!` and one for the counter change after `subject.inc_started_spans!`. If I use the `let!` block, the value of `result` is reused for each test.

spec/elastic_apm_spec.rb
estolfo: done!

lib/elastic_apm/instrumenter.rb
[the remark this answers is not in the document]
estolfo: done manually as part of rubocop cleanup

estolfo:
> Did you run this through Rubocop?
All offenses are now addressed.
> Did you do any manual e2e of this? Something like setting it up in Opbeans, seeing the spans show up correctly in Kibana?
Good idea, I'll do that.

spec/elastic_apm/transaction_spec.rb
mikker: My experience is that excessive `let` use (as in much more than 1 or 2 per spec file) leads to less code for the cost of increased indirection. Especially when coming back to a test 6 months later, figuring out what goes on before getting to the actual `it` example can be a task on its own. This is why I'm often sacrificing the brevity over ease of understanding.

I'm not sure I would argue that _plain ruby_ eg. inline variables in this (or maybe any) case would be an anti-pattern.

These are just my thoughts and suggestions though. Keep it like this if this is your preference 😊

estolfo: I see your point about excessive let blocks getting confusing. I do think that if the `it` blocks are kept lean-- ideally to a single validation-- then the test can be very clear; all the setup is kept out of the validation and resources are not wasted as the `let`'s are memoized.
Anyway, let's just leave this as it is, as I think it makes a lot of sense. I'll be sure not to use `let`'s excessively though = )

mikker: You're right. And I like rules like _only use `let!` if it makes sense to call it `:result`_ or something like that. That way it's indirection but indirection based on common knowledge in that project/context.

👍

spec/elastic_apm/transport/serializers/span_serializer_spec.rb
```diff
@@ -66,6 +68,29 @@ module Serializers
...
+                    http: { url: 'dsa' }
+                  )
+                )
+              end
+
+              it 'adds context object' do
+                expect(result.dig(:span, :context, :db, :statement))
+                  .to eq 'asd'
+                expect(result.dig(:span, :context, :http, :url)).to eq 'dsa'
+                expect(result.dig(:span, :context, :sync)).to be nil
```
mikker: It might be more interesting to check that `span.context` doesn't have a `:sync` key?
It's essentially the same but tells more about the implementation.

mikker:
I think it would make sense too to add a page to the docs about async spans and what to be aware of (if anything?)

docs/api.asciidoc
```diff
@@ -147,11 +147,29 @@ Arguments:
...
   * `trace_context:`: An optional `TraceContext` object for Distributed Tracing.
+  * `parent:`: An optional parent transaction or span. Relevant when the span is created in another thread.
+  * `sync:`: An boolean to indicate whether the span is created synchronously or not.
   * `&block`: An optional block to wrap with the span.
   The block is passed the span as an optional argument.
 
 Returns the created span.
 
+If you'd like to create an asynchronous span, you have to pass the parent `Span` or `Transaction` to the `start_span` method.
+You can also set the `sync` flag to `false`, if you'd like the span to be marked as asynchronous.
```
mikker:
```suggestion
You can also set the `sync` flag to `false`, if you'd like the span to be marked as asynchronous. This has no effect other than being queryable in Elasticsearch.
```

docs/api.asciidoc
```diff
@@ -147,11 +147,29 @@ Arguments:
...
+
+[source,ruby]
+----
+transaction = ElasticAPM.start_transaction
```
mikker:
```suggestion
transaction = ElasticAPM.current_transaction # or start one with `.start_transaction`
```

docs/api.asciidoc
```diff
@@ -173,11 +191,28 @@ Arguments:
...
   The block is passed the span as an optional argument.
 
 Returns the return value of the given block.
 
+If you'd like to create an asynchronous span, you have to pass the parent `Span` or `Transaction` to the method.
+You can also set the `sync` flag to `false`, if you'd like the span to be marked as asynchronous.
+
+[source,ruby]
+----
+transaction = ElasticAPM.start_transaction
```
mikker:
```suggestion
transaction = ElasticAPM.current_transaction # or start one with `.start_transaction`
```

PR #573, opened 2019-10-21
Convert config server_url value to a string

estolfo:


estolfo:
Thank you for pointing that out!

PR #575, opened 2019-10-21
Shuffle spec order on each run

mikker:
It might help us dig out unwanted inconsistencies in the specs. Or it might be annoying or confusing. I'm for it, _wdyt_ @estolfo?


estolfo:
I didn't know this setting existed. I've unintentionally had hidden dependencies between tests before, which would be great to expose sooner rather than later. I don't mind having this setting but let's see how it goes...

PR #580, opened 2019-10-21
Make sure to close spans after tests in OpenTracing spec

mikker:
I think yesterdays CI failures were due to some OT spans bleeding into the following examples. This should fix.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,8 +4,6 @@ source 'https://rubygems.org'
 
 git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
 
-gemspec
-
 gem 'rack-test'
 gem 'rspec'
 gem 'rspec-its'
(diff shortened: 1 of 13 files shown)


mikker:
I've moved all the external requires up front. Many were required multiple times and the specs were failing because we didn't require Http in CentralConfig. I think this makes it easier. You can't run the agent without these. All optional requires stay, all internal requires are still strewn about. Perhaps we should collect all those too? I've seen gems do that.

estolfo:
I like the idea of centralizing all external requires into one place. The only downside is that it's not always clear what depends on each requirement.
I've always put internal requires where they are relevant (for example, in the module file that corresponds to a subdirectory) and that seems to make most sense to me.

mikker:
Merging this. Test failures seem to be something awry with Rails 4.2. I'll investigate in another PR.

PR #585, opened 2019-10-22
Lazy load ActiveRecord adapter in SQL normalizer

edzhelyov:
Hello,

I've installed and configured the gem to test in our organization and we hit a strange issue with our configuration. We set the database settings during Rails' initialization process and due to a case how we run the background workers (using forking approach) the database credentials are available after Rails is fully booted. But the current code will force the AR adapter to initialize a collection and break. I know it's an edge case, but I think it's safer to lazy load the AR connection call when you actually try to track an SQL.

### DIFF
--- a/lib/elastic_apm/normalizers/rails/active_record.rb
+++ b/lib/elastic_apm/normalizers/rails/active_record.rb
@@ -15,7 +15,6 @@ module ElasticAPM
         def initialize(*args)
           super
 
-          @subtype = lookup_adapter || 'unknown'
           @summarizer = SqlSummarizer.new
         end
 
@@ -25,10 +24,14 @@ module ElasticAPM
           name = summarize(payload[:sql]) || payload[:name]
           context =
             Span::Context.new(db: { statement: payload[:sql], type: 'sql' })
-          [name, TYPE, @subtype, ACTION, context]
+          [name, TYPE, subtype, ACTION, context]
         end
 
         private
+
+        def subtype
+          @subtype ||= (lookup_adapter || 'unknown')
+        end
 
         def summarize(sql)
           @summarizer.summarize(sql)


mikker:
That does sound like an edge case but also something I'm fine with merging. I'm wondering whether we really have to memoize the result? Can apps change their adapter down the line? Your description of the issue suggests they can just fine. Can we be sure, we can cache it then?

edzhelyov:
Hm, I'm not sure. With recent Rails, you can connect to different databases, but I'm not sure how the adapter logic is implemented there. In our case, the call to AR::Base.connection fails at it tries to initialize the connection, but lacks credentials, so it's not that someone is changing them at runtime.

In the AS::Subscriber we receive the connection_id that is making the query, but I don't see a good way to get that connection. And I'm not sure if ActiveRecord::Base.connection is always the same connection that you catch in the notification. However, calling `ActiveRecord::Base.connection` every time will incur some overhead.

I can try to play a bit and do some benchmarks, but I'm unable to load Rails and run the specs. Any advice on how to run this normalizer specs against Rails' ActiveRecord?

I think we are good with caching for now and think about it if someone provides a use-case.

estolfo:
I'm pretty sure that you have to change your app's configuration files in order to use a different db adapter, which would require a restart. That said, I don't think it's possible to change the adapter at runtime, so I think it's ok to memoize.

mikker:
jenkins, run the tests please

mikker:
Let's merge this as is then 👍 do we want a changelog entry?

estolfo:
It wouldn't hurt to have a changelog entry I think.

edzhelyov:
Newer Rails supports [multiple databases](https://edgeguides.rubyonrails.org/active_record_multiple_databases.html), so in theory, you could have 2 connections with different adapters. But in the subscriber, it's unknown which connection was used, having object_id is useless in our case.

I [found that in Rails 6](https://github.com/rails/rails/pull/34568), it will send the connection object, so we might rework this to use the `payload[:connection]` and fallback to the Base's one.

PR #587, opened 2019-10-23
Add MongoSpy unit test and move original test to integration

### LINKED ISSUE #581: Split MongoSpy tests into integration and unit
The current MongoSpy tests start/stop a mongodb server, so take quite a bit of time.
We should split the tests into unit and integration.
Integration tests should use a real MongoDB server and unit tests should not require MongoDB to be running.

estolfo:
Closes elastic/apm-agent-ruby#581

### DIFF
--- a/spec/integration/mongo_spec.rb
+++ b/spec/integration/mongo_spec.rb
@@ -1,0 +1,135 @@
+# frozen_string_literal: true
+
+require 'mongo'
+
+module ElasticAPM
+  RSpec.describe 'Spy: MongoDB' do
+    before(:context) do
+      start_mongodb
+    end
+
+    after(:context) do
+      stop_mongodb
+    end
+
+    def stop_mongodb
+      `docker-compose -f spec/docker-compose.yml down -v 2>&1`
+    end
+
+    def start_mongodb
+      stop_mongodb
+      `docker-compose -f spec/docker-compose.yml up -d mongodb 2>&1`
+    end
+
+    let(:url) do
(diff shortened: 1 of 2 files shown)


estolfo:
@mikker Would you like me to add a flag so that the mongo integration tests conditionally run?

mikker:
Sure, yes, I think a flag would be good. Otherwise having both kind of disputes the purpose of making the suite faster.

1. Check for `ENV['CI']`
1. Check for `ENV['RUN_FULL_SUITE']` or something else we define now
1. Check for `END['INCLUDE_MONGO_SPEC']` or similar

Which one do you prefer?

estolfo:
Let's start with the simplest: `ENV['CI']`. (and I'm assuming this is set in Jenkins)
If we need to further categorize our specs later, we can add a separate env variable. I've just commit a check for `ENV['CI']`

estolfo:
@mikker I've updated this to fit with the recent test changes. Let me know if you're ok with it and I'll merge.

spec/integration/mongo_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+
+    after(:context) do
+      stop_mongodb
+    end
+
+    def stop_mongodb
+      `docker-compose -f spec/docker-compose.yml down -v 2>&1`
+    end
+
+    def start_mongodb
```
mikker: With the new dev setup we can just rely on this being true 😅

PR #588, opened 2019-10-25
ci(jenkins): add new framework test to the matrix

kuisathaverat:
add new frameworks test requested


.ci/.jenkins_framework.yml
```diff
@@ -11,3 +11,4 @@ FRAMEWORK:
   - grape-1.2
 
   - grape-1.2,rails-6.0
```
kuisathaverat:
```suggestion
```

PR #589, opened 2019-10-26
Run each integration test in their own process

mikker:
Skips `spec/integration` by default. Instead it runs then separately in each their own spec.

It's very hard to not have the Rails specs bleed into each other if we want to test different behaviours with different setups. This way each integration test has its very own, close-to pristine universe.


bin/run-tests
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+base="bundle exec rspec -f progress -f JUnit -o spec/ruby-agent-junit.xml"
```
mikker: @kuisathaverat We probably want to separate the junit outputs too but I'm unsure what's possible in Jenkins.

mikker:
Only test failure is one of the custom jruby images failing to build `http-parser` (dep via `http`). Let's hear if the automation folks can figure out what's up.

spec/scripts/spec.sh
```diff
@@ -43,4 +43,4 @@ RUBY_VERSION=${VERSION} docker-compose run \
   /bin/bash -c "\
     gem install rake && \
     bundle install && \
-    timeout -s9 5m bundle exec rspec -f progress -f JUnit -o spec/ruby-agent-junit.xml ${TEST}"
+    timeout -s9 5m bin/run-tests ${TEST}"
```
kuisathaverat: I do not see a reason to split the script, it calls another script that is in the folder bin, it is weird

mikker: Did you see the script?

mikker: It needs to run _one_ command when passed a `$1` and multiple if not.

bin/run-tests
```diff
@@ -0,0 +1,16 @@
...
+set -x
+
+base="bundle exec rspec -f progress -f JUnit -o spec/ruby-agent-junit.xml"
+specific_spec=$1
+
+if [[ $specific_spec = '' ]]; then
+  echo 'Running all specs, including integration'
+
+  $base
+  find spec/integration -name '*_spec.rb' -exec $base {} \;
```
kuisathaverat: it launches several tests in a sequential way, each one will generate a junit file, if all use the same name we will only have the latest one, so we have to make the name uniq for each test, we would need a function

kuisathaverat:
```suggestion
  find spec/integration -name '*_spec.rb' -exec runBaseCase {} \;
```

bin/run-tests
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -x
+
+base="bundle exec rspec -f progress -f JUnit -o spec/ruby-agent-junit.xml"
+specific_spec=$1
+
+if [[ $specific_spec = '' ]]; then
+  echo 'Running all specs, including integration'
+
+  $base
```
kuisathaverat:
```suggestion
  runBaseCase
```

bin/run-tests
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -x
+
+base="bundle exec rspec -f progress -f JUnit -o spec/ruby-agent-junit.xml"
```
kuisathaverat:
```suggestion
runBaseCase(){
  local case=${1:-""}
  bundle exec rspec -f progress -f JUnit -o spec/ruby-agent-junit${case}.xml ${case}
}
```

bin/run-tests
```diff
@@ -0,0 +1,16 @@
...
+
+if [[ $specific_spec = '' ]]; then
+  echo 'Running all specs, including integration'
+
+  $base
+  find spec/integration -name '*_spec.rb' -exec $base {} \;
+else
+  echo "Running only $specific_spec"
+
+  $base $1
```
kuisathaverat:
```suggestion
  runBaseCase $1
```

spec/scripts/spec.sh
kuisathaverat: we can use a function

spec/elastic_apm/spies/mongo_spec.rb
```diff
@@ -53,32 +56,32 @@ def start_mongodb
...
+      with_agent do
+        client =
+          Mongo::Client.new(
+            [url],
+            database: 'elastic-apm-test',
+            logger: Logger.new(nil),
+            server_selection_timeout: 5
+          )
+
+        # ParallelCollectionScan can only be run on an existing collection.
```
estolfo: This comment can be removed now.

mikker:
@kuisathaverat I don't see how your function makes anything different? The proposal format makes the changes a bit hard to follow.

spec/elastic_apm/spies/mongo_spec.rb
```diff
@@ -89,28 +92,25 @@ def start_mongodb
       expect(db.type).to eq 'mongodb'
       # ParallelCollectionScan doesn't send 'lsid' in the command so we can
       # validate the entire command document.
-      expect(db.statement).to eq '{"parallelCollectionScan"=>"testing", ' \
-        '"numCursors"=>1}'
+      expect(db.statement).to match('"delete"=>"testing"')
```
estolfo: You can remove the comment about ParallelCollectionScan but I still don't understand how you can validate the entire delete command? When I was writing this test, the `lsid` value was in the command so I couldn't use delete and chose ParallelCollectionScan instead. Is there some setup somewhere I'm not aware of?

mikker:
I got it on the second read. Thanks! Did you see the build failure on Jenkins?

spec/elastic_apm/spies/mongo_spec.rb
estolfo: For example, I get this `"{\"delete\"=>\"testing\", \"ordered\"=>true, \"deletes\"=>[{\"q\"=>{}, \"limit\"=>0}], \"lsid\"=>{\"id\"=><BSON::Binary:0x70210864279520 type=uuid data=0x4c4b83c656b94211...>}}"`

kuisathaverat:
yep, I am on it it is failing on jRuby 9.2-12 only should be somethignon this image or in that version I have to check it, probably I ask for help I'm not a ruby guy :)

spec/elastic_apm/spies/mongo_spec.rb
mikker: `match(String)` (as opposed to `eq()`) uses `include?` under the hood, I think, so it's just a matter of the string being in there somewhere.

The reason I changed it at all was that my Mongo (that I finally figured out had moved to their own homebrew tap) was telling me that `parallelCollectionScan` had been removed.

estolfo: Oh great, I didn't notice you used `match` instead. This works too!

kuisathaverat:
jenkins, run the tests

mikker:
omg @kuisathaverat this is so gloriously green

green-light

PR #593, opened 2019-10-29
Use the connection from payload to get adapter, when available

### LINKED ISSUE #592: Use payload[:connection] to get adapter in SqlNormalizer
See this comment https://github.com/elastic/apm-agent-ruby/pull/585#issuecomment-545546649

estolfo:
Pass the payload object to the `lookup_adapter` method so that when it contains the `connection` object, it can be used get the adapter name. Fall back to `::ActiveRecord::Base.connection`, which is the previous behavior.

Made the names to skip a constant so that a new array isn't created everytime the `normalize` method is called.

Closes elastic/apm-agent-ruby#592

### DIFF
--- a/lib/elastic_apm/normalizers/rails/active_record.rb
+++ b/lib/elastic_apm/normalizers/rails/active_record.rb
@@ -32,9 +32,9 @@ module ElasticAPM
         private
 
         def subtype(payload)
-          return connection_adapter(payload[:connection]) if payload.key?(:connection)
-
-          @default_adapter ||= connection_adapter(::ActiveRecord::Base.connection)
+          connection_adapter(
+            payload[:connection] ||::ActiveRecord::Base.connection
+          )
         end
 
         def summarize(sql)
@@ -42,7 +42,10 @@ module ElasticAPM
         end
 
         def connection_adapter(connection)
-          connection.adapter_name.downcase || UNKNOWN
+          @adapters ||= {}
+          return UNKNOWN unless connection.adapter_name
+          @adapters[connection.adapter_name] ||
+            (@adapters[connection.adapter_name] = connection.adapter_name.downcase)
         rescue StandardError
           nil
         end
--- a/lib/elastic_apm/normalizers/rails/active_record.rb
+++ b/lib/elastic_apm/normalizers/rails/active_record.rb
@@ -33,7 +33,7 @@ module ElasticAPM
 
         def subtype(payload)
           connection_adapter(
-            payload[:connection] ||::ActiveRecord::Base.connection
+            payload[:connection] || ::ActiveRecord::Base.connection
           )
         end
 
@@ -41,11 +41,11 @@ module ElasticAPM
           @summarizer.summarize(sql)
         end
 
-        def connection_adapter(connection)
+        def connection_adapter(conn)
           @adapters ||= {}
-          return UNKNOWN unless connection.adapter_name
-          @adapters[connection.adapter_name] ||
-            (@adapters[connection.adapter_name] = connection.adapter_name.downcase)
+          return UNKNOWN unless conn.adapter_name
+          @adapters[conn.adapter_name] ||
+            (@adapters[conn.adapter_name] = conn.adapter_name.downcase)
         rescue StandardError
           nil
         end
--- a/lib/elastic_apm/normalizers/rails/active_record.rb
+++ b/lib/elastic_apm/normalizers/rails/active_record.rb
@@ -12,7 +12,7 @@ module ElasticAPM
         TYPE = 'db'
         ACTION = 'sql'
         SKIP_NAMES = %w[SCHEMA CACHE].freeze
-        UNKNOWN = 'unknown'.freeze
+        UNKNOWN = 'unknown'
 
         def initialize(*args)
           super
--- a/lib/elastic_apm/normalizers/rails/active_record.rb
+++ b/lib/elastic_apm/normalizers/rails/active_record.rb
@@ -11,6 +11,7 @@ module ElasticAPM
 
         TYPE = 'db'
         ACTION = 'sql'
+        SKIP_NAMES = %w[SCHEMA CACHE].freeze
 
         def initialize(*args)
           super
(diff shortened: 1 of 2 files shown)


lib/elastic_apm/normalizers/rails/active_record.rb
```diff
@@ -19,26 +20,27 @@ def initialize(*args)
...
-          [name, TYPE, subtype, ACTION, context]
+          [name, TYPE, subtype(payload), ACTION, context]
         end
 
         private
 
-        def subtype
-          @subtype ||= (lookup_adapter || 'unknown')
+        def subtype(payload)
+          @subtype ||= (lookup_adapter(payload) || 'unknown')
```
mikker: This is still using `||=`. Is that what we want? If so, then is there in the end any reason to get the connection from the payload?

mikker: I realise the overhead of checking, but if we're still just checking once, is there really a need? Do we need to memoize scoped on some key from the payload/connection?

estolfo: Good point, I missed that.
I think in Rails 6, the connection object could be different each time this method is reached, so you would want to check the payload object.
What about this? (**untested**)

```
def subtype(payload)
  return @subtype if @subtype
  if payload[:connection]
    lookup_adapter(payload)
  else
    @subtype = lookup_adapter(payload)
  end
end
```

That way, you'll continue looking up the adapter in the `payload` while it is present, otherwise, it'll memoize the `@subtype`.

mikker: That's not totally true if `payload[:connection]` is only sometimes present?

```ruby
def subtype(payload)
  return lookup_adapter(payload[:connection]) if payload.key?(:connection)

  @default_adapter ||= lookup_adapter(::ActiveRecord::Base.connection)
end
```

`lookup_adapter` is the wrong method name in this version. Perhaps `adapter_from_connection`.

estolfo: Sorry, I should have explained that this is kind of a compromise - we will always looks up the `payload[:connection]` value if it exists but the first time it doesn't exist, we are going to memoize the result and not check anymore. I doubt Rails will sometimes have the `connection` in the payload and sometimes not (I'll check this), therefore I think it will never get to memoize the value. So this allows us to have different lookup methods for Rails 6 and < Rails 6 without us checking the Rails version.

estolfo: I just went through the active_support notifications source code and I can't find where the `connection` key is set so I don't know if it's always there for a `sql.active_record` event. I'll keep looking but maybe it makes more sense to do what you suggested.

estolfo: Update: I found where the `connection` is set in the payload. It's done [here](https://github.com/rails/rails/blob/966d3a7bf2c08c5c20213d29686cbb409a07734a/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L682) in the`AbstractAdapter`. So we can count on it always being there in Rails 6.0 for `sql.active_record` events.

mikker: Wasn't one purpose of this to allow folks to have multiple adapters?

estolfo: The adapter will only ever be memoized if the payload doesn't have a `connection` key. Which apparently will never be the case with Rails 6.
So yes, people can have multiple adapters because the `connection` value in the payload will always be checked.

estolfo: But in any case, if the code isn't showing that without explanation, we should make it clearer = )
I'll incorporate your suggestion and change the method name to be more accurate.

mikker: You're right and thanks. We could still memoize the `adapter => formatted string` method scoped on some (cheap to look up) key on `connection`.

lib/elastic_apm/normalizers/rails/active_record.rb
```diff
@@ -11,6 +11,8 @@ class SqlNormalizer < Normalizer
 
         TYPE = 'db'
         ACTION = 'sql'
+        SKIP_NAMES = %w[SCHEMA CACHE].freeze
+        UNKNOWN = 'unknown'.freeze
```
mikker: This is already frozen with `frozen_string_literal: true`

estolfo: doh! Sorry, old habits.

lib/elastic_apm/normalizers/rails/active_record.rb
```diff
@@ -19,26 +21,31 @@ def initialize(*args)
...
           @summarizer.summarize(sql)
         end
 
-        def lookup_adapter
-          ::ActiveRecord::Base.connection.adapter_name.downcase
+        def connection_adapter(conn)
+          @adapters ||= {}
+          return UNKNOWN unless conn.adapter_name
+          @adapters[conn.adapter_name] ||
+            (@adapters[conn.adapter_name] = conn.adapter_name.downcase)
```
mikker: Would be perfect if the `[]` lookup is slower than `downcase` 😅

lib/elastic_apm/normalizers/rails/active_record.rb
```diff
@@ -19,26 +21,31 @@ def initialize(*args)
...
         end
 
         def summarize(sql)
           @summarizer.summarize(sql)
         end
 
-        def lookup_adapter
-          ::ActiveRecord::Base.connection.adapter_name.downcase
+        def connection_adapter(conn)
+          @adapters ||= {}
```
mikker: We could set this in `initialize` and skip the `||=` here.

spec/elastic_apm/normalizers/active_record_spec.rb
```diff
@@ -53,6 +55,49 @@ def normalize_payload(payload)
...
+              sql: sql,
+              connection: double(adapter_name: nil)
+            )
+
+            expect(name).to eq 'SELECT FROM pizza'
+            expect(type).to eq 'db'
+            expect(subtype).to eq 'unknown'
+            expect(action).to eq 'sql'
+            expect(context_.db.statement).to eq sql
+          end
```
mikker: Just my tastes, but I don't think you need to expect every value with every adapter. Keeping it to the expected changes makes it easier to read  the test and understand what's under the microscope.

PR #594, opened 2019-10-30
Fix configuration in Kibana requests

mikker:
We were constructing the requests with an outdated method. This uses the proper, documented approach instead.

This PR also makes the CentralConfig polling share functionality with the Intake API sending. Setting headers, SSLContext, proxy settings.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,10 +32,14 @@ else
 end
 
 ## Install Framework
-GITHUB_REPOS = { 'grape' => 'ruby-grape/grape'}
+GITHUB_REPOS = {
+  'grape' => 'ruby-grape/grape',
+  'rails' => 'rails/rails',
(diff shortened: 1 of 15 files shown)


Gemfile
```diff
@@ -32,18 +32,22 @@ else
...
+frameworks_versions = parsed_frameworks.inject({}) do |frameworks, str|
   framework, *version = str.split('-')
   frameworks.merge(framework => version.join('-'))
 end
 
 frameworks_versions.each do |framework, version|
   case version
   when 'master'
-    gem framework, github: (GITHUB_REPOS[framework] || "#{framework}/#{framework}")
+    gem framework, github: GITHUB_REPOS.fetch[framework]
```
mikker: Rubocop warnings

PR #599, opened 2019-11-04
Sneakers agent using Sneakers Middleware support

redwrx:
Sneakers agent leveraging latest sneakers middleware

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ gem 'sequel', require: nil
 gem 'sidekiq', require: nil
 gem 'simplecov', require: false, group: :test
 gem 'simplecov-cobertura', require: false, group: :test
+gem 'sneakers', require: nil
 gem 'yard', require: nil
 gem 'yarjuf'
 
--- a/lib/elastic_apm/spies/sneakers.rb
+++ b/lib/elastic_apm/spies/sneakers.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module ElasticAPM
+  # @api private
+  module Spies
+    # @api private
+    class SneakersSpy
+      include Logging
+
+      def install
+        # sneakers 2.12.0 introduced middleware concept and the spy needs that
+        if Gem.loaded_specs["sneakers"].version < Gem::Version.create("2.12.0")
+          warn("Sneakers version is below 2.12.0. Sneakers spy installation failed")
+        else
+          Sneakers.middleware.use(Middleware, nil)
+        end
+      end
+      # @api private
+      class Middleware
+        def initialize(app, *args)
+          @app = app
+          @args = args
+        end
+
+        def call(deserialized_msg, delivery_info, metadata, handler)
+          transaction = ElasticAPM.start_transaction(delivery_info.consumer.queue.name, 'Sneakers')
+          ElasticAPM.set_label(:routing_key, delivery_info.routing_key)
+          res = @app.call(deserialized_msg, delivery_info, metadata, handler)
+          transaction.done :success if transaction
+          res
+        rescue ::Exception => e
+          ElasticAPM.report(e, handled: false)
+          transaction.done :error if transaction
+          raise
+        ensure
+          ElasticAPM.end_transaction
+        end
+      end
+    end
+    register 'Sneakers', 'sneakers', SneakersSpy.new
+  end
+end
--- a/spec/elastic_apm/spies/sneakers_spec.rb
+++ b/spec/elastic_apm/spies/sneakers_spec.rb
@@ -1,0 +1,89 @@
+require 'sneakers'
+require 'elastic_apm/spies/sneakers'
+module ElasticAPM
+  RSpec.describe 'Spy: Sneakers', :intercept do
+    # sneakers 2.12.0 introduced middleware concept and the spy needs that
+    sneakers_middleware_not_avail = Gem.loaded_specs["sneakers"].version < Gem::Version.create("2.12.0")
+
+    class Queue
+      def name
+        'q1'
+      end
+    end
+
+    class Consumer
+      def queue
+        Queue.new
+      end
+    end
+
+    class DeliveryInfo
+      def routing_key
+        'r1234'
+      end
+      def consumer
+        Consumer.new
+      end
+    end
+
+    class TestWorker
+      include Sneakers::Worker
+      from_queue 'q1',
+                 ack: false
+      def work(message)
+      end
+    end
+
+    class TestErrorWorker
+      include Sneakers::Worker
+      from_queue 'q1',
+                 ack: false
+      def work(message)
+        1/0
+      end
+    end
+
+    before :all do
+      Sneakers.configure
+    end
+
+    before { ElasticAPM.start }
+    after { ElasticAPM.stop }
+
+    it 'instruments job transaction' do
+      worker = TestWorker.new
+      worker.process_work(DeliveryInfo.new, nil, nil, nil)
+      transaction, = @intercepted.transactions
+
+      if sneakers_middleware_not_avail
+        expect(transaction).to eq nil
+      else
+        expect(transaction.name).to eq DeliveryInfo.new.consumer.queue.name
+        label, = transaction.context.labels
+        expect(label[:routing_key]).to eq DeliveryInfo.new.routing_key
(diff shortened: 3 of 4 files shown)


elasticcla:
Hi @redwrx, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/apm-agent-ruby/pull/599.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

estolfo:
Hi @redwrx, thanks for the PR! Would you be able to add a test or would you like us to add one?
You can use any of [these](https://github.com/elastic/apm-agent-ruby/tree/master/spec/elastic_apm/spies) as a model.

Also, is there a reason you've put an rc version in the Gemfile? Wouldn't we want to test the latest stable version?

mikker:
Hi @redwrx! This looks great! Can you perhaps check to be sure that it works @estolfo? I'd be very happy to include this.

I'll insert this here so we wont forget:

- [ ] Add this to the _Supported tech_ docs with a version going back as far as possible with the used middleware API.
- [ ] More docs? I don't think we have any special docs for Sidekiq for example. Do we need that?
- [ ] Changelog entry
- [ ] Tests

mikker:
We also need you @redwrx to resolve the CLA issue posted by the bot above. It'll only take a minute and the lawyers will be happy too 😅

estolfo:
Sure @mikker 
I've just added a Tests checkbox as well.

redwrx:
> Hi @redwrx, thanks for the PR! Would you be able to add a test or would you like us to add one?
> You can use any of [these](https://github.com/elastic/apm-agent-ruby/tree/master/spec/elastic_apm/spies) as a model.
> 
> Also, is there a reason you've put an rc version in the Gemfile? Wouldn't we want to test the latest stable version?

Thanks for your feedback, I will add the tests and make sure it uses the latest stable gem. It shouldn't take long

redwrx:
> We also need you @redwrx to resolve the CLA issue posted by the bot above. It'll only take a minute and the lawyers will be happy too 😅

sure, I tried it couple times with all my emails (on my github account) but none of them worked.

redwrx:
> > Hi @redwrx, thanks for the PR! Would you be able to add a test or would you like us to add one?
> > You can use any of [these](https://github.com/elastic/apm-agent-ruby/tree/master/spec/elastic_apm/spies) as a model.
> > Also, is there a reason you've put an rc version in the Gemfile? Wouldn't we want to test the latest stable version?
> 
> Thanks for your feedback, I will add the tests and make sure it uses the latest stable gem. It shouldn't take long

@estolfo I added the unit tests and forced pushed. About the sneakers rc gem, they have introduced the Middleware concept in sneakers in that version and I am using that to start the transaction. I prefer to use their latest implementation so we don't have to adapt to the middleware later when they release it. thanks

redwrx:
> > We also need you @redwrx to resolve the CLA issue posted by the bot above. It'll only take a minute and the lawyers will be happy too 😅
> 
> sure, I tried it couple times with all my emails (on my github account) but none of them worked.

@mikker  so far I think at least I submitted 10 CLA forms but the bot doesn't like it

mikker:
I completely missed that Emily had already commented when I did. Sorry about the confusion 🙄

lib/elastic_apm/spies/sneakers.rb
```diff
@@ -0,0 +1,31 @@
+module ElasticAPM
+  # @api private
+  module Spies
+    # @api private
+    class SneakersSpy
+      def install
+        Sneakers.middleware.use(Middleware, nil)
+      end
+      #
```
estolfo:
```suggestion
      # @api private
```

lib/elastic_apm/spies/sneakers.rb
```diff
@@ -0,0 +1,31 @@
+module ElasticAPM
```
estolfo:
```suggestion
# frozen_string_literal: true

module ElasticAPM
```

spec/elastic_apm/spies/sneakers_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'sneakers'
+
```
estolfo: I think you'll have to require the spy here as well:

`require 'elastic_apm/spies/sneakers'`

lib/elastic_apm/spies/sneakers.rb
```diff
@@ -0,0 +1,31 @@
...
+      end
+      #
+      class Middleware
+        def initialize(app, *args)
+          @app = app
+          @args = args
+        end
+
+        def call(deserialized_msg, delivery_info, metadata, handler)
+          transaction = ElasticAPM.start_transaction(delivery_info.routing_key , 'Sneakers')
```
estolfo: Do we want the first argument to `start_transaction` (i.e. the name) to be something related to the message? Then have the label on the transaction be the `routing_key`. 
I'm asking because in the Sidekiq spy, we use the job's class name as the name for the transaction and set a label for the queue identifier. 
Perhaps you can explain what the `routing_key`'s significance is and we can see what attribute on the transaction suits it best. For example, if it's something that changes a lot, perhaps it'd be better to have it be a label.

Gemfile
```diff
@@ -22,6 +22,7 @@ gem 'sequel', require: nil
 gem 'sidekiq', require: nil
 gem 'simplecov', require: false, group: :test
 gem 'simplecov-cobertura', require: false, group: :test
+gem 'sneakers', '2.12.0.rc1', require: nil
```
estolfo: This was released on 4 March, 2019. Do you know how far off a GA is? The reason I'm asking is, it would be a pity to add support for Sneakers but only support a rc for an indefinite amount of time.

redwrx: I agree however it doesn't look like they are going to release that soon, you can follow here:
https://github.com/jondot/sneakers/issues/367

lib/elastic_apm/spies/sneakers.rb
redwrx: you are absolutely right. The problem is that in Sneakers, the middleware doesn't have access to the worker class/name. I opened a pull request to change that https://github.com/jondot/sneakers/pull/414

That is why I used routing_key as a replacement for now.

lib/elastic_apm/spies/sneakers.rb
redwrx: added

lib/elastic_apm/spies/sneakers.rb
redwrx: added

spec/elastic_apm/spies/sneakers_spec.rb
redwrx: added

lib/elastic_apm/spies/sneakers.rb
redwrx: @estolfo  I changed the sneakers spy to use queue name for transaction and uses the routing_key as a label

mikker:
The issue has been closed and the gem released! (yay!)

I'm wondering whether we want to do some version checking in the spy, so as to not use a non-existent API in sneakers version below 2.12.0? Keeping the spirit of _just works_ I think we could check on load and log a warning/debug message if version is below, otherwise install the middleware as we do now?

redwrx:
> The issue has been closed and the gem released! (yay!)
> 
> I'm wondering whether we want to do some version checking in the spy, so as to not use a non-existent API in sneakers version below 2.12.0? Keeping the spirit of _just works_ I think we could check on load and log a warning/debug message if version is below, otherwise install the middleware as we do now?

That is a good idea, I will work on that this week and let you know when it is ready

Gemfile
redwrx: Gemfile updated

redwrx:
@mikker @estolfo I updated the Gemfile, the spy and the tests to check for sneakers 2.12.0

mikker:
Wonderful! 💙💚💛💜❤️

So we don't forget:
- [ ] Add to the _Supported tech_ docs with version requirements
- [ ] Changelog entry

mikker:
Continued this in #676. Thanks a lot for your work in researching and implementing this, @redwrx!

PR #605, opened 2019-11-08
Get single-line stacktrace for Grape using source from payload's endpoint object

estolfo:
Our approach for getting a stacktrace for Grape will be in three phases:

1. → (this PR) For older versions of Grape, provide a single-line stacktrace using the source taken from the `endpoint` object in `payload`.
2. Open a PR to Grape extracting execution of the user's code to another method so we can have a spy for it and get a relevant stacktrace.
3. Open a PR to Grape providing an interface so that we can define a block that is run before the user's code is executed. We can get the stacktrace using the interface.

### DIFF
--- a/lib/elastic_apm/normalizers/rails/action_mailer.rb
+++ b/lib/elastic_apm/normalizers/rails/action_mailer.rb
@@ -15,8 +15,6 @@ module ElasticAPM
           [endpoint(payload), TYPE, SUBTYPE, ACTION, nil]
         end
 
-        def backtrace(payload); end
-
         private
 
         def endpoint(payload)
--- a/lib/elastic_apm/normalizers/rails/action_mailer.rb
+++ b/lib/elastic_apm/normalizers/rails/action_mailer.rb
@@ -15,8 +15,7 @@ module ElasticAPM
           [endpoint(payload), TYPE, SUBTYPE, ACTION, nil]
         end
 
-        def stacktrace_top(span); end
-        def source_location(payload); end
+        def backtrace(payload); end
 
         private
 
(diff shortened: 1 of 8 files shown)


lib/elastic_apm/subscriber.rb
```diff
@@ -54,14 +54,20 @@ def start(name, id, payload)
...
-    def finish(_name, id, _payload)
+    def finish(name, id, payload)
       # debug "AS::Notification#finish:#{name}:#{id}"
       return unless (transaction = @agent.current_transaction)
 
       while (notification = transaction.notifications.pop)
         next unless notification.id == id
 
         if (span = notification.span)
+          _caller = caller
```
estolfo: I will probably change this to check if the normalizer can get a stacktrace so as not to waste calling `caller` unless I need to.

lib/elastic_apm/subscriber.rb
```diff
@@ -54,14 +54,15 @@ def start(name, id, payload)
...
-    def finish(_name, id, _payload)
+    def finish(name, id, payload)
       # debug "AS::Notification#finish:#{name}:#{id}"
       return unless (transaction = @agent.current_transaction)
 
       while (notification = transaction.notifications.pop)
         next unless notification.id == id
 
         if (span = notification.span)
+          span.original_backtrace ||= @normalizers.backtrace(name, payload)
```
mikker: It's probably very little overhead, but as this is something that's potentially happening many times per second, we could skip the few allocations by checking `config.span_frames_min_duration?` before calling this?

lib/elastic_apm/normalizers/rails/action_mailer.rb
```diff
@@ -15,6 +15,8 @@ def normalize(_transaction, _name, payload)
           [endpoint(payload), TYPE, SUBTYPE, ACTION, nil]
         end
 
+        def backtrace(payload); end
```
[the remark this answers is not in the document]
estolfo: oh yes, of course! I missed that they all inherit from Normalizer.

lib/elastic_apm/subscriber.rb
estolfo: ok, done!

estolfo:
@mikker Ready for review again.

PR #606, opened 2019-11-11
Add runner service to docker-compose for local test runs

mikker:
Okay, this grew a bit in scope –

For usage instructions, see added section to README:
https://github.com/mikker/apm-agent-ruby/blob/runner/README.md#development

@estolfo: Do you feel that this is something you'd use? (If not, what's wrong/missing?)
@v1v: Does the Docker setup look sane enough or have I gone completely off the rails?

Keeping two separate Docker setups might be overkill, but I wanted to avoid keeping in legacy stuff unless it was strictly necessary. The best approach might be to use this new setup (if possible) for Jenkins too?


mikker:
We should probably add a _framework_ argument too, like `-f rails:master,sinatra:2.0,grape:1.2`.

bin/dev
```diff
@@ -0,0 +1,29 @@
...
+IMAGE = options.fetch(:image, 'ruby:latest')
+IMAGE_PATH_SAFE = IMAGE.gsub(':', '_')
+IMAGE_NAME = "apm-agent-ruby:#{IMAGE_PATH_SAFE}"
+VENDOR_PATH = "/vendor/#{IMAGE_PATH_SAFE}"
+
+def run(cmd)
+  system "IMAGE_NAME=#{IMAGE_NAME} #{cmd}"
+end
+
+run 'docker-compose build ' \
```
v1v: Will the docker container might change the source code somehow? If so, then it might be a good practice to match both userids, the one from the host and the one in the container when running the docker-compose run

Dockerfile
```diff
@@ -0,0 +1,32 @@
+ARG RUBY_IMAGE
+FROM ${RUBY_IMAGE}
+
+ARG VENDOR_PATH
+ARG BUNDLER_VERSION
+
+# For tzdata
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN [ apt-get ] \
```
v1v: This conditional might not be required if the RUBY_IMAGE is Debian based, or are any plans to support other OS flavors?

.dockerignore
```diff
@@ -0,0 +1 @@
+Gemfile.lock
```
v1v: I'd exclude some of the files which are not required for the local test runs:

```suggestion
Gemfile.lock
*.asciidoc
*.md
docs/
.ci/
```

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    && ( apt-get update -qq \
+        && apt-get install -qq -y build-essential libpq-dev git tzdata) \
```
v1v: Let's add some minor tweaks to reduce the docker image size just a bit:

```suggestion
        && apt-get install -qq -y --no-install-recommends build-essential libpq-dev git tzdata \
        && rm -rf /var/lib/apt/lists/*) \
```

bin/dev
kuisathaverat: this is important, if you do not change the user, the build would create files in the folder `vendor` and `.` as the root user

mikker: It should not change anything. Unless the dev opens an editor inside and changes things.

Dockerfile
mikker: True! I copied it from the other Dockerfile where we also have support for `yum`.

bin/dev
```diff
@@ -0,0 +1,29 @@
...
+  " --build-arg RUBY_IMAGE=#{IMAGE}" \
+  " --build-arg VENDOR_PATH=#{VENDOR_PATH}"
+run "docker-compose run --rm specs #{ARGV.join}"
```
kuisathaverat:
```suggestion
run "docker-compose run -u $(id -u):$(id -g) --rm specs #{ARGV.join}"
```

bin/dev
mikker: I remember switching to the host's user in the old setup at some point too. What's the minimal approach to doing so?

estolfo:
@mikker To answer your question 
> Do you feel that this is something you'd use? (If not, what's wrong/missing?)

It would be great to be able to specify specific frameworks to test, otherwise test them all (rails, grape, mongodb).
I think the reason we started doing this yesterday was so that we could recreate the errors we are getting in CI. So I would use this and try different ruby images, starting with jruby to try to get the errors we often see in Jenkins. If this would allow me to get those errors, I would definitely use it a lot = )

bin/dev
mikker: 👍 didn't notice this before writing the comment above

mikker:
> what's the diff between the docker setup in spec and this one, or how we could normalize both.

I guess this is tailored from the start to make it easier for us (me, really) to spin up a running container with any env that we use on jenkins, so I don't have to go all the way around building on every test run. It also brings caching the deps so I also don't have to do that with every new container (unless changed of course).

idk, maybe I'll end up reconstructing all of the old setup here, but at least it felt healthy to use a clean slate.

docker-compose.yml
```diff
@@ -14,6 +14,7 @@ services:
         BUNDLER_VERSION: '2.0.2'
     image: '$IMAGE_NAME'
     environment:
+      HOME: '/tmp'
```
mikker: Am I going to jail for this one?

v1v: `/dev/null` might be a good reason to go prison :P

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+ENV GEM_HOME=$VENDOR_PATH
+ENV BUNDLE_PATH=$GEM_HOME \
+  BUNDLE_JOBS=4 BUNDLE_RETRY=3
+ENV BUNDLE_APP_CONFIG=$BUNDLE_PATH \
+  BUNDLE_BIN=$BUNDLE_PATH/bin
+ENV PATH=/app/bin:$BUNDLE_BIN:$PATH
+
+ENV FRAMEWORKS $FRAMEWORKS
+
+RUN mkdir -p $VENDOR_PATH && \
```
v1v: Cosmetic feedback:

This could be potentially moved to the very end and use the USER directive rather than using the chown 
command?

```Dockerfile
...
WORKDIR /app

USER $USER_ID_GROUP
RUN mkdir -p $VENDOR_PATH
```

Although I'm not sure whether the value for the USER_ID_GROUP varible might be based on the `gid:usid` format.

estolfo:
I've been using this locally and if you switch between images like this:
```
bin/dev -i jruby:9.2
# Let tests run, then try:
bin/dev -i jruby:9.1
```
You get this error: 
```
Could not find rake-12.3.3 in any of the sources
Run `bundle install` to install missing gems.
```
So I think the `Gemfile.lock` needs to be cleared or an option needs to be passed to `bundle` in `entrypoint` so that the gems are updated.

mikker:
I thought `Gemfile.lock` would be ignored by `.dockerignore`? Does the ignore only apply to `ADD` and `COPY` commands?

estolfo:
not sure, do you get an error like that if you switch between images locally?

v1v:
> Does the ignore only apply to ADD and COPY commands?

Indeed. See https://docs.docker.com/engine/reference/builder/#dockerignore-file

mikker:
In that case we can skip `.dockerignore` altogether as all files will be included in the mounted root volume.

Dockerfile
mikker: I couldn't make `USER` work, so reverted to the `chown` approach

mikker:
@estolfo Can you try your jruby version dance again? I think it's fixed now.

mikker:
@v1v @kuisathaverat Perhaps we ought to bundle the custom jruby dockerfiles for the versions we test against inside this repo? It's only 5/6 combinations, and having just flat, static Dockerfiles bundled would make it wayyy easier to set them up locally.

estolfo:
No problem switching between images now 👍🏼

v1v:
> @v1v @kuisathaverat Perhaps we ought to bundle the custom jruby dockerfiles for the versions we test against inside this repo? It's only 5/6 combinations, and having just flat, static Dockerfiles bundled would make it wayyy easier to set them up locally.

No strong opinion, I'd love to see the support for different `jruby-jdk` tuples within the upstream project as tracked https://github.com/cpuguy83/docker-jruby/issues/50.

If those repos are moved then we need somehow ensure those docker build images are published too.

mikker:
AFAIK the custom JRuby images' Dockerfiles are dynamically generated with a script, then built, then published to our internal registry? My proposal was just that we include those built Dockerfiles (what is it, 6 or 7?) in this repo and then publish them from here instead?

Same result, but the Dockerfiles are in a different repo (this one).

mikker:
@kuisathaverat @v1v I'll make a new issue about bundling the JRuby Dockerfiles. Thanks for helping out with this one!

PR #615, opened 2019-11-18
Set remote_addr to immediate socket

### LINKED ISSUE #583: remote_addr should be the immediate socket
See: https://github.com/elastic/apm/issues/157#issuecomment-544881634

mikker:
Closes #583

### DIFF
--- a/spec/elastic_apm/context/request/socket_spec.rb
+++ b/spec/elastic_apm/context/request/socket_spec.rb
@@ -17,7 +17,7 @@ module ElasticAPM
         end
 
         # 'Trusted' as per Rack's definition:
-        # https://github.com/rack/rack/blob/2-0-stable/lib/rack/request.rb#L419-L421
+        # https://github.com/rack/rack/blob/2.0.7/lib/rack/request.rb#L419-L421
         context 'with a "trusted" remote addr and forwarding header' do
           let(:req) do
             Rack::Request.new(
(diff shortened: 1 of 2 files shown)


spec/elastic_apm/context/request/socket_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+              Rack::MockRequest.env_for('/', 'REMOTE_ADDR' => '127.0.0.1')
+            )
+          end
+
+          its(:remote_addr) { is_expected.to eq '127.0.0.1' }
+        end
+
+        # 'Trusted' as per Rack's definition:
+        # https://github.com/rack/rack/blob/2-0-stable/lib/rack/request.rb#L419-L421
+        context 'with a "trusted" remote addr and forwarding header' do
```
estolfo: This is really minor, but it might be better to link to a tagged version's file. If you link to the lines in a branch, they may change as time goes on and we lose the reference.
For example https://github.com/rack/rack/blob/2.0.7/lib/rack/request.rb#L419-L421

spec/elastic_apm/context/request/socket_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+            Rack::Request.new(
+              Rack::MockRequest.env_for('/', 'REMOTE_ADDR' => '127.0.0.1')
+            )
+          end
+
+          its(:remote_addr) { is_expected.to eq '127.0.0.1' }
+        end
+
+        # 'Trusted' as per Rack's definition:
+        # https://github.com/rack/rack/blob/2-0-stable/lib/rack/request.rb#L419-L421
```
mikker:
```suggestion
        # https://github.com/rack/rack/blob/2.0.7/lib/rack/request.rb#L419-L421
```

PR #617, opened 2019-11-19
Refactor worker so that it manages its own threads and can be mocked easily

estolfo:
This refactor of the worker has the following benefits:
- The worker manages its own thread so that only threads are created and stopped/killed, not workers. Previously, workers were created inside threads with nothing referencing them. This meant that when a thread was stopped or killed, the worker would be disposed of and garbage collected, thus wasting objects.
- The workers can now be easily accessed via `ElasticAPM.agent.transport.workers` and therefore be mocked easily. We can now, for example, use the dependency-injection (`conn_adapter`) built into the `Worker` in the Rails tests.
- There was a mismatch in naming between `transport.workers` and what they actually were. Previously, `transport.workers` returned a list of threads, not workers. Now `transport.workers` returns a list of `Worker`s.
- Workers are created greedily, only once, in `Transport#initialize`, instead of as part of `Transport#start`. They are now only *started* via `Transport#start`.

### DIFF
--- a/lib/elastic_apm/transport/base.rb
+++ b/lib/elastic_apm/transport/base.rb
@@ -30,7 +30,13 @@ module ElasticAPM
         @filters = Filters.new(config)
 
         @stopped = Concurrent::AtomicBoolean.new
-        @workers = Array.new(config.pool_size)
+        @workers = Array.new(config.pool_size) do
+          Worker.new(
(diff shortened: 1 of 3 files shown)


estolfo:
Closing as we won't take this approach.

Issue #612, opened 2019-11-15
Spec suite shape-up


mikker:
Experiments/suggestions that we want to try:

- [ ] No manual WebMock resetting
- [ ] Logging#error raises in test (unless expected)
- [ ] Move wait_for to MockIntake.wait_for
- [ ] Stripped down defaults for with_agent
- [ ] Rails.application.config.deep_merge!(DEFAULT_RAILS_TEST_OPTIONS)
	- [ ] WithAgent::DEFAULT_TEST_OPTIONS maybe?
- [ ] Omit `-r spec_helper` in `.rspec`. Only require in the specs that need it.
        - Evaluate whether we want to rename the current `spec_helper` to `agent_helper` or something similar and have a **minimal** `spec_helper` (setting up load path, requiring `rspec/its`, bare minimum)
- [ ] Add `require: nil` to everything in `Gemfile`

PR #620, opened 2019-11-19
Test suite fixes

estolfo:
@mikker This PR now is stable in CI. I'm wondering how far you think I should go with the changes?

The current changes are:
1. New `EventCollector` with a shared context that is used in the `rails_spec`
2. Error handling in `CentralConfig` when the `error` object in `handle_error` is a `WebMock` error and doesn't respond to the methods we expect
3.  `adapter` setting on the `Worker` class
4.  Explicit method for setting rails config in `spec_helper` (`self.setup_rails_test_config`)

Further work could include the following:
- Change all uses of `MockIntake` to `EventCollector`
- Add default options for agent
- Add default options for an agent with disabled send
- Remove error handling in `CentralConfig` (see 2 above), as it's only useful for tests. Figure out another way to handle this situation - i.e. by making sure to mock central config requests everywhere in the test suite.
- Other items from the list in this issue https://github.com/elastic/apm-agent-ruby/issues/612


mikker:
Do you feel strongly about the name `RequestParser`? I don't really feel that that's what it does, viewed from the perspective of a test, ie. where we'd use it.

estolfo:
No, not at all. This PR isn't ready for review yet. When I clean it up, I'll rethink about the name and propose something else.

bin/run-tests
```diff
@@ -18,6 +18,7 @@ if [[ $specific_spec = '' ]]; then
   runRspec
   for i in $(find spec/integration -name '*_spec.rb')
   do
+    echo $i
```
mikker: Let's make it easily scannable, eg:

```suggestion
    echo "========================================"
    echo $i
    echo "========================================"
```

lib/elastic_apm/central_config.rb
```diff
@@ -149,9 +153,10 @@ def headers
       { 'Etag': @etag }
     end
 
-    def schedule_next_fetch(resp)
+    def schedule_next_fetch(resp=nil)
```
mikker:
```suggestion
    def schedule_next_fetch(resp = nil)
```

lib/elastic_apm/central_config.rb
```diff
@@ -129,9 +129,13 @@ def handle_error(error)
...
+    ensure
       assign({})
-
-      schedule_next_fetch(error.response)
+      if error.respond_to?(:response)
+        schedule_next_fetch(error.response)
+      else
+        schedule_next_fetch
+      end
     end
```
mikker: I feel this part is a bit weird. We're rescue'ing with `ensure` from calling `.response` on something not having a `.response` and then we do the `.respond_to` dance afterwards.

How about:

```ruby
    def handle_error(error)
      # For tests, WebMock failures don't have real responses
      response = error.respond_to(:response) ? error.response : nil

      debug(
        'Failed fetching config: %s, trying again in %d seconds',
        response&.body, DEFAULT_MAX_AGE
      )

      assign({})

      schedule_next_fetch(response)
    end
```

lib/elastic_apm/central_config.rb
```diff
@@ -149,9 +153,10 @@ def headers
       { 'Etag': @etag }
     end
 
-    def schedule_next_fetch(resp)
+    def schedule_next_fetch(resp=nil)
+      headers = resp ? resp.headers : {}
       seconds =
-        if (cache_header = resp.headers['Cache-Control'])
+        if (cache_header = headers['Cache-Control'])
```
mikker: If you use the implementation I suggested, `resp` will either be a real response or `nil`, so this will be enough:

```suggestion
        if (cache_header = resp&.headers['Cache-Control'])
```

I think...

lib/elastic_apm/transport/worker.rb
```diff
@@ -17,15 +25,15 @@ def initialize(
         queue,
         serializers:,
         filters:,
-        conn_adapter: Connection
+        conn_adapter: nil
```
mikker: I don't think we should have both methods of Dependency Injection. IMO let's remove `conn_adapter`. It's only used in `worker_spec.rb`.

spec/elastic_apm/grape_spec.rb
```diff
@@ -3,26 +3,41 @@
 if defined?(Grape)
   RSpec.describe Grape do
     describe '.start' do
-      before do
+
```
mikker: I think Rubocop will complain over this empty line 😅

```suggestion
```

spec/elastic_apm/grape_spec.rb
```diff
@@ -3,26 +3,41 @@
...
 
-      after do
-        ElasticAPM.stop
+      after(:all) do
         Object.send(:remove_const, :GrapeTestApp)
       end
 
       context 'with no overridden config settings' do
+        before do
+          ElasticAPM::Grape.start(GrapeTestApp, config)
```
mikker: Would this not need to stub the request for CentralConfig? Or can we just _not care_ with these changes? (If so, 👍)

spec/integration/rails_spec.rb
```diff
@@ -12,41 +12,31 @@
   require 'action_controller/railtie'
   require 'action_mailer/railtie'
 
-  # We don't use :mock_intake, as we want the stubs to stay around between
-  # individual examples
-  RSpec.describe 'Rails integration', :allow_running_agent, :spec_logger do
+  RSpec.describe 'Rails integration' do
     include Rack::Test::Methods
-    include MockIntake::WaitFor
+    include_context 'event_collector'
```
mikker: I don't have any objections but are there reasons you prefer `shared_context` over the example tags approach that's used elsewhere? (Eg, `describe "thing", :mock_intake`)

spec/integration/rails_spec.rb
```diff
@@ -12,41 +12,31 @@
...
-
-    after :each do
-      MockIntake.clear!
+    let(:app) do
+      Rails.application
     end
 
     after :all do
       ElasticAPM.stop
+      ElasticAPM::Transport::Worker.adapter = nil
```
mikker: I feel like this and the `TestAdapter` ought to be part of `EventCollector` so it could be more easily shared between multiple specs.

spec/integration/rails_spec.rb
```diff
@@ -57,15 +47,11 @@ class FancyError < ::StandardError; end
           ElasticAPM.set_user(current_user)
         end
 
-        def index
-          render_ok
-        end
-
         http_basic_authenticate_with(
-          name: 'dhh', password: 'secret123', only: [:create]
+            name: 'dhh', password: 'secret123', only: [:create]
```
mikker:
```suggestion
          name: 'dhh', password: 'secret123', only: [:create]
```

spec/integration/rails_spec.rb
```diff
@@ -118,211 +104,231 @@ def ping(recipient, subject)
...
 
-      @mock_intake = MockIntake.stub!
+      class TestAdapter < ElasticAPM::Transport::Connection
+        def write(payload)
+          EventCollector.catalog(JSON.parse(@metadata))
+          EventCollector.catalog JSON.parse(payload)
+        end
+      end
+
+      ElasticAPM::Transport::Worker.adapter = TestAdapter
```
mikker: See above comment about including in `EventCollector` itself. I'm also not a fan of subclassing `Connection`. It feels like it's too easy to result in mysterious, unwanted behaviour. I'd much rather have exceptions from calling methods, `TestAdapter` doesn't implement (and then add them as noops where necessary.)

spec/integration/rails_spec.rb
```diff
@@ -118,211 +104,231 @@ def ping(recipient, subject)
...
+          expect(user['email']).to eq 'person@example.com'
+        end
       end
 
-      it 'ignores url patterns' do
-        get '/ping'
-        get '/'
+      context 'when there are ignored url patterns defined', :allow_running_agent do
+        it 'does not create events for the patterns' do
+          get '/ping'
```
mikker: Did you remove `/ping` from the routes on purpose?

spec/support/mock_intake.rb
```diff
@@ -187,3 +187,116 @@ def print_received
...
+      instance.send(name, *args, &block)
+    end
+
+    def instance
+      @instance ||= new
+    end
+  end
+
+  attr_reader(
+      :errors,
```
mikker:
```suggestion
    :errors,
```

spec/support/mock_intake.rb
```diff
@@ -187,3 +187,116 @@ def print_received
...
+      :metadatas,
```
mikker:
```suggestion
    :metadatas,
```

spec/support/mock_intake.rb
```diff
@@ -187,3 +187,116 @@ def print_received
...
+      :metricsets,
```
mikker:
```suggestion
    :metricsets,
```

spec/support/mock_intake.rb
```diff
@@ -187,3 +187,116 @@ def print_received
...
+      :requests,
```
mikker:
```suggestion
    :requests,
```

spec/support/mock_intake.rb
```diff
@@ -187,3 +187,116 @@ def print_received
...
+      :spans,
```
mikker:
```suggestion
    :spans,
```

spec/support/mock_intake.rb
```diff
@@ -187,3 +187,116 @@ def print_received
...
+      :transactions
```
mikker:
```suggestion
    :transactions
```

spec/support/mock_intake.rb
```diff
@@ -187,3 +187,116 @@ def print_received
...
+    @transactions = []
+  end
+
+  def transaction_metrics
+    metricsets.select { |set| set['transaction'] && !set['span'] }
+  end
+
+  def span_metrics
+    metricsets.select { |set| set['transaction'] && set['span'] }
+  end
```
mikker: I'm not sure I think these are generic enough to extract.

spec/support/mock_intake.rb
```diff
@@ -187,3 +187,116 @@ def print_received
...
+
+  def wait_for(timeout: 5, **expected)
+
```
mikker:
```suggestion
```

spec/support/mock_intake.rb
```diff
@@ -187,3 +187,116 @@ def print_received
...
+  end
+end
+
+shared_context 'event_collector' do
+  before do
+    EventCollector.clear!
+  end
+  after do
+    EventCollector.clear!
+  end
```
mikker: Is this to be _absolutely_ sure it's empty? 😅

spec/elastic_apm/grape_spec.rb
estolfo: On one hand, the more defensive code in `CentralConfig` makes it not entirely necessary to mock `CentralConfig` requests in test. On the other hand, it's kind of "messy" to not mock `CentralConfig`.

So I think I've talked myself into having both the defensive code in `CentralConfig` in case a test forgets to mock it and also mock `CentralConfig` in this grape spec for completeness. What do you think @mikker ?

spec/support/mock_intake.rb
estolfo: Do you mean as opposed to using the current inline block approach (i.e. [this strategy](https://github.com/elastic/apm-agent-ruby/blob/master/spec/integration/rails_spec.rb#L294-L296))? I got a lot of random test failures with jRuby using the block approach, where the `EventCollector`/`MockIntake` object was `nil` inside the block . I therefore decided to do a more explicit approach using a method and those random failures went away. I can't explain why but I'm guess it has something to do with defining a block's binding in JRuby.

estolfo: i.e. this works so I'd rather have an explicit method even though it's not perfect object interface design.

spec/support/mock_intake.rb
estolfo: 🧹

spec/support/mock_intake.rb
mikker: We can keep them as is for now if you'd like 👍

estolfo: I'm afraid to change anything 😬

spec/integration/rails_spec.rb
estolfo: No, I didn't. I'll add it back!

spec/support/mock_intake.rb
mikker: Welcome to the `rails_spec.rb`

lib/elastic_apm/central_config.rb
estolfo: I had the `ensure` in there because I was concerned that this method throwing an uncaught error could have such a significant effect on the test suite. It's not entirely necessary to have it but if we see strange "bleeding" behavior in the test suite, it's worth checking this--and similar-- methods for uncaught exceptions.

spec/integration/rails_spec.rb
estolfo: I think including the context is more explicit. When you see a `include_context`, you assume that there are some `before`/`after`/`let`/etc blocks defined + included. With the `:mock_intake` metadata, it's not clear to me whether it does something like what `:allow_running_agent` does (raise an exception if some condition is not met) or if it defines `before`/`after`/`let`/etc blocks that have a direct role in the test. In this case, what we want to include, is, in fact, an extra context so I think it makes most sense to define it as such.

spec/integration/rails_spec.rb
estolfo: good idea!

spec/integration/rails_spec.rb
estolfo: I'm not sure I agree. Essentially what I'm doing here is using the real `Connection` class and mocking the `#write` method. Ideally, I could access the real `connection` object associated with a `worker` and mock the method there, but the current architecture doesn't allow me to do that; you can't access workers via `agent.transport`. (This is why I opened this [PR](https://github.com/elastic/apm-agent-ruby/pull/617), so that the workers owned their threads and were accessible via `agent.transport` and then could be mocked)

If we rewrite the methods that are necessary for this test to run, then those mock method definitions could eventually diverge from the real definitions, also resulting in mysterious behavior. For example, if I were to remove the inheritance, I would have to define an `initialize` method that matches the initialize method signature currently on `Connection`. I rely on the definition of `@metadata`, contained in `Connection#initialize`, so the mock `initialize` method definition can't be a noop. I'd rather use the real `Connection#initialize` method so that I can be sure that when I check the `@metadata` later, it was defined through the real code, not my mock object's definition.

Let me know what you think and if I've missed some points!

lib/elastic_apm/central_config.rb
mikker: We don’t see them because the agent swallows and logs most of them. We talked about having `Logging#error` raise if test. I think that is a good approach to try?

spec/integration/rails_spec.rb
mikker: With Sorbet types we could write an abstract interface and make sure any adapter would have the same api. I guess that's for another day 😅

Subclassing is fine. You make some good points 👍

spec/spec_helper.rb
```diff
@@ -102,3 +114,15 @@ def configure_rails_for_test
...
+
+RSpec.shared_context 'stubbed_central_config' do
+  before(:all) do
+    WebMock.stub_request(
+        :get, %r{^http://localhost:8200/config/v1/agents/?$}
+    ).to_return(body: '{}')
+  end
+
+  after(:all) do
+    WebMock.reset!
```
mikker: After this PR we can use the regular `webmock/rspec` require and not have to do this.

spec/integration/rails_spec.rb
estolfo: Yes, that is true! I was also going to make the argument that this is similar to how static languages mock objects (using abstract interfaces) in tests but didn't want to go overboard.
Thank you for hearing me out = )

lib/elastic_apm/central_config.rb
estolfo: Yes, but my concern is not that errors are swallowed. I think it's a good approach to raise in tests from Logging#error.
My concern was that I observed that following some errors in this method, the agent wasn't able to stop the central config tasks and the agent was still running in other tests. I'll try to find a build to show you, it was sometime last week.

estolfo: Unfortunately the build from last week is no longer available in Jenkins. In any case, let's take the approach of raising in `Logging#error`.

spec/support/mock_intake.rb
```diff
@@ -183,3 +183,138 @@ def print_received
...
+        transactions: transactions.map { |o| o['name'] },
+        spans: spans.map { |o| o['name'] },
+        errors: errors.map { |o| o['culprit'] },
+        metricsets: metricsets,
+        metadatas: metadatas.count
+    )
+  end
+end
+
+shared_context 'event_collector' do
```
mikker:
```suggestion
RSpec.shared_context 'event_collector' do
```

Issue #582, opened 2019-10-22
Concurrency issue in `add_filter`


estolfo:
Occasionally we see the following error

```
expected no Exception, got #<RuntimeError: can't add a new key into hash during iteration> with backtrace:
# ./lib/elastic_apm/transport/filters.rb:22:in `add'
# ./lib/elastic_apm/transport/base.rb:74:in `add_filter'
# ./lib/elastic_apm/agent.rb:232:in `add_filter'
# ./lib/elastic_apm.rb:324:in `add_filter'
# ./spec/elastic_apm_spec.rb:190:in `block in <main>'
```

which is indicative of two threads trying to access/mutate the filter list at the same time. We should investigate and add a lock.

mikker:
Fixed by #624

PR #624, opened 2019-11-21
Make Filters thread-safe

mikker:
Re: #582 

As the `Filters` object is shared between the workers, one worker may try to add a filter while another is applying the current filters.

This PR fixes it by altering a copy of the filters hash instead of modifying the original.

### DIFF
--- a/spec/elastic_apm/transport/filters_spec.rb
+++ b/spec/elastic_apm/transport/filters_spec.rb
@@ -45,14 +45,18 @@ module ElasticAPM
         end
       end
 
-      describe 'thread safety' do
-        it 'may add filters while applying' do
+      describe 'from multiple threads' do
+        it "doesn't complain" do
           threads =
             (0...100).map do |i|
-              Thread.new { subject.apply!(payload: i) }
+              Thread.new do
+                if i.even?
+                  subject.apply!(payload: i)
+                else
+                  subject.add :"filter_#{i}", ->(_) { '' }
+                end
+              end
             end
-
-          subject.add :new, ->(_) { '' }
 
           threads.each(&:join)
         end
--- a/spec/elastic_apm/transport/filters_spec.rb
+++ b/spec/elastic_apm/transport/filters_spec.rb
@@ -44,6 +44,19 @@ module ElasticAPM
           expect(untouched).to_not have_received(:call)
         end
       end
+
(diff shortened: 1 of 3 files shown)


mikker:
Actually I like that immutable approach better.

estolfo:
+1 for making it immutable!

spec/elastic_apm/transport/filters_spec.rb
```diff
@@ -44,6 +44,19 @@ module Transport
...
       end
+
+      describe 'thread safety' do
+        it 'may add filters while applying' do
+          threads =
+            (0...100).map do |i|
+              Thread.new { subject.apply!(payload: i) }
+            end
+
+          subject.add :new, ->(_) { '' }
```
estolfo: Were you able to reproduce the bug with this test before adding a fix? I ran this without the fix in `Filters` and couldn't get it to fail.

This only gives the `add` and `apply` one chance to step on each other. I would suggest instead increasing the likelihood of the error by adding many new keys into the hash while applying the filters.
I'm able to reproduce the bug every time I run this test:

```ruby
    context 'altering and accessing filters at the same time' do
        it 'does not mutate the original' do
          threads = []
          100.times do |i|
            threads <<
              if i % 2 == 0
                Thread.new { subject.apply!(payload: i) }
              else
                Thread.new { subject.add "filter#{i}", ->(_) {''} }
              end
          end
          threads.each(&:join)
        end
      end
```

```ruby
$ rspec spec/elastic_apm/transport/filters_spec.rb:48
Randomized with seed 29514

ElasticAPM::Transport::Filters
  altering and accessing filters at the same time
    does not mutate the original (FAILED - 1)

Failures:

  1) ElasticAPM::Transport::Filters altering and accessing filters at the same time does not mutate the original
     Failure/Error: @filters[key] = filter
     
     RuntimeError:
       can't add a new key into hash during iteration
     # ./lib/elastic_apm/transport/filters.rb:22:in `add'
     # ./spec/elastic_apm/transport/filters_spec.rb:56:in `block in Transport'
```

mikker:
Very nice suggestions, thanks!

mikker:
CI errors unrelated

mikker:
Oh no! Forgot to change the e-mail in the dropdown 😬

PR #627, opened 2019-11-22
Repo clean-up and Rubocop action

mikker:
It looks like we miss Rubocop's warnings if we don't have them here. This action annotates directly on Github which is a big help.

I also disabled a bunch of the checks in the `Metrics/*Length` category as we've discussed earlier.


bin/dev
```diff
@@ -28,7 +28,7 @@ USER_ID_GROUP = %w[u g].map { |f| `id -#{f}`.chomp }.join(':')
 RUBY_IMAGE = options.fetch(:image, 'ruby:latest')
 FRAMEWORKS = options.fetch(:frameworks, 'rails,sinatra,grape')
 
-IMAGE_PATH_SAFE = RUBY_IMAGE.gsub(':', '_')
+IMAGE_PATH_SAFE = RUBY_IMAGE.tr(':', '_')
```
estolfo: Why did you change this?

lib/elastic_apm/central_config/cache_control.rb
```diff
@@ -26,7 +26,7 @@ def initialize(value)
       def parse!(value)
         value.split(',').each do |token|
           k, v = token.split('=').map(&:strip)
-          instance_variable_set(:"@#{k.gsub('-', '_')}", v ? v.to_i : true)
+          instance_variable_set(:"@#{k.tr('-', '_')}", v ? v.to_i : true)
```
estolfo: and this?

spec/integration/rails_grape_spec.rb
```diff
@@ -11,7 +11,8 @@
 if enabled
   require 'action_controller/railtie'
 
-  RSpec.describe 'Rails and Grape integration', :mock_intake, :allow_running_agent do
+  RSpec.describe 'Rails and Grape integration', :mock_intake,
+    :allow_running_agent do
```
estolfo: can we line this up under `:mock_intake`?

bin/dev
mikker: `rubocop-performance` suggests that it's faster than `gsub` for regular string (non-regex)

lib/elastic_apm/central_config/cache_control.rb
mikker: same 😊

spec/integration/rails_grape_spec.rb
mikker: We can break both? I don't like special situation indenting levels, really.

spec/integration/rails_grape_spec.rb
```diff
@@ -11,7 +11,8 @@
 if enabled
   require 'action_controller/railtie'
 
-  RSpec.describe 'Rails and Grape integration', :mock_intake, :allow_running_agent do
+  RSpec.describe 'Rails and Grape integration', :mock_intake,
```
mikker:
```suggestion
  RSpec.describe 'Rails and Grape integration',
    :mock_intake, :allow_running_agent do
```

spec/integration/rails_grape_spec.rb
mikker:
```suggestion
```

bin/dev
estolfo: oh nice, I didn't know that it was faster.

PR #629, opened 2019-11-25
Sanitise auth info from outgoing HTTP requests' span.context.url

### LINKED ISSUE #474: Sanitise userinfo in URLs

mikker:
1. For incoming reqs I've confirmed that Rack doesn't pass the _userinfo_ through.
1. For outgoing reqs I added code to strip the password part.

---

Test setup:

```ruby
# config.ru
require 'rack'
require 'elastic_apm/context/request/url'

class App
  def call(env)
    req = Rack::Request.new(env)
    puts req.host # => localhost

    url = ElasticAPM::Context::Request::Url.new(req)

    [200, {}, [url.inspect]]
  end
end

use Rack::Auth::Basic do |user, pass|
  user == 'user' && pass = 'pass'
end

run App.new
```

Fixes elastic/apm-agent-ruby#474

### DIFF
--- a/lib/elastic_apm/span/context.rb
+++ b/lib/elastic_apm/span/context.rb
@@ -28,12 +28,28 @@ module ElasticAPM
       # @api private
       class Http
         def initialize(url: nil, status_code: nil, method: nil)
-          @url = url
+          @url = sanitize_url(url)
           @status_code = status_code
           @method = method
         end
 
         attr_accessor :url, :status_code, :method
+
+        private
+
+        def sanitize_url(url)
+          uri = URI(url)
+
+          return url unless uri.userinfo
+
+          format(
+            '%s://%s@%s%s',
+            uri.scheme,
+            uri.userinfo.split(':')[0], # username, no password
+            uri.hostname,
+            uri.path
+          )
+        end
       end
     end
   end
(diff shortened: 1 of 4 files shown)


mikker:
Forgot the last part, checking `context.http.url` now.

lib/elastic_apm/span/context.rb
```diff
@@ -28,12 +28,28 @@ def initialize(instance: nil, statement: nil, type: nil, user: nil)
...
+
+        def sanitize_url(url)
+          uri = URI(url)
+
+          return url unless uri.userinfo
+
+          format(
+            '%s://%s@%s%s',
+            uri.scheme,
+            uri.userinfo.split(':')[0], # username, no password
```
estolfo: You can just call ` uri.user` I believe.

mikker: You are right!

mikker:
```suggestion
            uri.user,
```

spec/elastic_apm/span/context_spec.rb
```diff
@@ -22,6 +22,19 @@ module ElasticAPM
...
+
+        context 'when given auth info' do
+          subject do
+            described_class.new(
+              http: { url: 'https://user%40email.com:pass@example.com/q=a@b' }
+              #                         %40 => @
+            )
+          end
+
+          it 'strips the auth part' do
```
estolfo: it only strips the password, not the whole auth part...

mikker:
```suggestion
          it 'omits the password' do
```

PR #632, opened 2019-11-28
Remove default framework and add env var in benchmark script

estolfo:
This removes a default framework in the `Gemfile` and adds a `FRAMEWORK` environment variable to the docker container running the benchmark tests.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,15 +8,10 @@ repos:
     -   id: check-merge-conflict
     -   id: check-yaml
     -   id: check-xml
-    -   id: end-of-file-fixer
 
 -   repo: git@github.com:elastic/apm-pipeline-library
     rev: current
     hooks:
     -   id: check-bash-syntax
-    -   id: check-abstract-classes-and-trait
-    -   id: check-jsonslurper-class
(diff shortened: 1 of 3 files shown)


mikker:
@v1v It's unclear to me what actually failed in this CI run?

estolfo:
The linter failed https://github.com/estolfo/apm-agent-ruby/commit/ed58635377b59d33c45c18ac0a293bc7322e0230/checks
but the result are empty and when I run `rubocop` locally, it passes. So 🤷🏻‍♀️

v1v:
> regarding the linter

`spec/scripts/benchmarks.sh` got some end-of-file issue

https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-ruby%2Fapm-agent-ruby-linting-mbp/detail/PR-632/2/tests

The sanity check stage does run the `pre-commit` command but the rubocop is not yet in place. 

What commands should I run for validating the rubocop? Then I can add it

mikker:
I recently added a GH action for Rubocop that annotates the code directly on GH. I'd prefer to use that instead. See https://github.com/elastic/apm-agent-ruby/pull/627/files#diff-90829e76e906f1c73140c7ded7e1b268

PR #635, opened 2019-11-28
ci(jenkins): trigger opbeans release process

v1v:

Add a final stage in the release process to bump the opbeans agent dependency.

## Why is it important?

When a new tag, aka a new release, is created then, the automation will kick the opbeans build which consists of the following tasks:
- [x] Bump dependency versions in the opbeans-ruby
- [x] Push changes. (Implementation within the opbeans-ruby but the call it's done within this pipeline)
- [x] Create a tag to trigger the release pipeline in the opbeans-ruby. That particular tag will match with the same one that the one in the agent itself.

This particular pipeline will push the changes into the `master` branch for the opbeans-ruby and tag it with the name of the tag version too. Therefore, the opbeans-ruby pipeline will trigger two builds, the one for its master branch and the other one for the just created tag. This particular flow is implemented within the opbeans-ruby.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -24,6 +24,7 @@ pipeline {
     GITHUB_CHECK_ITS_NAME = 'Integration Tests'
     ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
     RELEASE_SECRET = 'secret/apm-team/ci/apm-agent-ruby-rubygems-release'
+    OPBEANS_REPO = 'opbeans-ruby'
   }
   options {
     timeout(time: 2, unit: 'HOURS')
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -190,26 +191,47 @@ pipeline {
...
-            script {
-              dir(BASE_DIR){
-                docker.image("${env.RUBY_DOCKER_TAG}").inside('-v /etc/passwd:/etc/passwd -v ${HOME}/.ssh:${HOME}/.ssh') {
-                  withEnv(["HOME=${env.WORKSPACE}/${env.BASE_DIR ?: ''}"]) {
-                    rubygemsLogin.withApi(secret: "${env.RELEASE_SECRET}") {
-                      sshagent(['f6c7695a-671e-4f4f-a331-acdce44ff9ba']) {
-                        sh '.ci/prepare-git-context.sh'
-                        sh 'gem install rake yard rspec'
-                        sh 'rake release'
+        stages {
```
v1v: Created a top-level release stage which contains:
- a nested `release` stage as used to be -> https://github.com/elastic/apm-agent-ruby/pull/635/files#diff-8db37e8fcea0f1a8f2f39667e94ebcc4R195
- a new nested `opbeans` stage <--- this is the new thing. https://github.com/elastic/apm-agent-ruby/pull/635/files#diff-8db37e8fcea0f1a8f2f39667e94ebcc4R218

PR #636, opened 2019-11-28
Add mutex to Counter Metric

mikker:
Fixes a potential thread issue.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,16 +10,16 @@ require:
 
 # Layout/
 
-Layout/AlignArguments:
+Layout/ArgumentAlignment:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -9,6 +9,6 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Rubocop checks
-      uses: mikker/rubocop-action@master
+      uses: mikker/rubocop-action@mikker-patch-1
```
estolfo: what does this mean?

mikker: Means what git ref to use. I was testing the Rubocop GH action. Removed these commits again.

Issue #611, opened 2019-11-14
mongo GetMore command collection name is wrong


estolfo:
A GetMore command has these properties:
`name:"elastic-apm-test.#<BSON::Int64:0x00007ffa6eb883f8>.getMore`
it should instead be:
`name:"elastic-apm-test.collection_name.getMore`

To reproduce, insert 3 items to a collection, then run the following command:
`client['testing'].find({}, batch_size: 2).to_a`

PR #637, opened 2019-11-29
Correct span context collection name for Mongo getMore commands

estolfo:
Closes https://github.com/elastic/apm-agent-ruby/issues/611

### DIFF
--- a/lib/elastic_apm/spies/mongo.rb
+++ b/lib/elastic_apm/spies/mongo.rb
@@ -41,9 +41,15 @@ module ElasticAPM
           # Some MongoDB commands are not on collections but rather are db
           # admin commands. For these commands, the value at the `command_name`
           # key is the integer 1.
-          unless event.command[event.command_name] == 1
-            collection = event.command[event.command_name]
-          end
+          # For getMore commands, the value at `command_name` is the cursor id
+          # and the collection name is at the key `collection`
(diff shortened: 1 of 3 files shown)


lib/elastic_apm/spies/mongo.rb
```diff
@@ -41,9 +41,15 @@ def push_event(event)
           # Some MongoDB commands are not on collections but rather are db
           # admin commands. For these commands, the value at the `command_name`
           # key is the integer 1.
-          unless event.command[event.command_name] == 1
-            collection = event.command[event.command_name]
-          end
+          # For getMore commands, the value at `command_name` is the cursor id
+          # and the collection name is at the key `collection`
+          collection = if event.command[event.command_name] == 1 ||
```
mikker:
```ruby
collection =
  if event.command[event.command_name] == 1 ||
    event.command[event.command_name].is_a?(BSON::Int64)
    event.command[:collection]
  else
    event.command[event.command_name]
  end
```

Not sure if Rubocop will complain but fwiw I hate arbitrary indenting levels. If you break it like above instead, it's just 2 spaces like everywhere else.

💅, I know.

estolfo: Sure, I'll try that and see if rubocop complains.

PR #639, opened 2019-12-02
Rails stacktraces for controller actions

estolfo:
This PR adds a spy on the`BasicImplicitRender` module to capture the stacktrace before a controller action is called. It's missing the final call to the user's controller method but they could use the request metadata (the route specifically) to determine which action was called.

It would work for [Rails 5.x](https://github.com/rails/rails/blob/v5.2.4/actionpack/lib/action_controller/metal/basic_implicit_render.rb#L5-L7) and [Rails 6.x](https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/metal/basic_implicit_render.rb#L5-L7).

### DIFF
--- a/lib/elastic_apm/rails.rb
+++ b/lib/elastic_apm/rails.rb
@@ -30,14 +30,6 @@ module ElasticAPM
       ElasticAPM.start(config).tap do |agent|
         attach_subscriber(agent)
       end
-
-      if ElasticAPM.running? &&
-         !ElasticAPM.agent.config.disabled_instrumentations.include?(
-           'action_dispatch'
-         )
-        require 'elastic_apm/spies/action_dispatch'
-      end
-
       ElasticAPM.running?
     rescue StandardError => e
       if config.disable_start_message?
--- a/lib/elastic_apm/rails.rb
+++ b/lib/elastic_apm/rails.rb
@@ -36,6 +36,7 @@ module ElasticAPM
            'action_dispatch'
          )
         require 'elastic_apm/spies/action_dispatch'
+        require 'elastic_apm/spies/action_controller'
       end
 
       ElasticAPM.running?
--- a/lib/elastic_apm/rails.rb
+++ b/lib/elastic_apm/rails.rb
@@ -36,7 +36,6 @@ module ElasticAPM
            'action_dispatch'
          )
         require 'elastic_apm/spies/action_dispatch'
-        require 'elastic_apm/spies/action_controller'
       end
 
       ElasticAPM.running?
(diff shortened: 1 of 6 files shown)


lib/elastic_apm/rails.rb
```diff
@@ -36,6 +36,7 @@ def start(config)
            'action_dispatch'
          )
         require 'elastic_apm/spies/action_dispatch'
+        require 'elastic_apm/spies/action_controller'
```
mikker: Do we want this (these even?) loaded like the other spies? So a user could easily `config.enabled_spies -= ['action_controller']`

lib/elastic_apm/spies/action_controller.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module ElasticAPM
+  # @api private
+  module Spies
+    # @api private
```
mikker: I don't think you need this for modules/classes that're only there for the namespacing.

mikker: Did you find out a way to get Rubocop into RubyMine btw?

lib/elastic_apm/rails.rb
estolfo: I dont know, what do you think? Would people want to disable these?

lib/elastic_apm/spies/action_controller.rb
estolfo: Yeah, I have Rubocop in RubyMine but don't always have it enabled because it's distracting.
We have the comments `api private` for all the other namespace declarations and rubocop complains if I don't have the comments for `api private`. So I think we have to leave them.

estolfo:
We currently say that we support Rails since version 4.2. This PR introduces a change that would only work with Rails 5.x and up. I don't think it's worth saying we "dropped" support for 4.x because of this PR because it'll just silently not generate stacktraces for controller actions, which is the current behavior. So I'll just add a note that stacktraces for controller actions will only be generated for Rails versions 5.x and up. Does that sound good?

mikker:
That sounds perfect!

lib/elastic_apm/rails.rb
mikker: As `caller` adds overhead, especially on JRuby, I'd say we ought to let people disable it.

estolfo:
The failure in the latest test run is unrelated. The only change in the commit that triggered the run was to the changelog.
The test run before that succeeded: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-ruby%2Fapm-agent-ruby-mbp/detail/PR-639/14/pipeline

PR #641, opened 2019-12-03
Add Ruby 2.7-rc (currently preview-3) to test matrix

mikker:
Released on Christmas day (25/12) we should probably start looking at compatibility.


mikker:
3x3 is next year!

PR #647, opened 2019-12-05
Add span.context.destination to relevant spans

### LINKED ISSUE #613: Add `destination` ECS fields to spans
See ~https://github.com/elastic/apm/issues/115~ https://github.com/elastic/apm/issues/180

mikker:
This PR adds information to a new field for all _outgoing_ spans, eg. HTTP, databases.

> For the service maps feature, we need to report the logical destination of a span in order to discover the unique connections between services without processing every single span. We also need to have a field that will eventually help detect if two non-instrumented services are the same. Sameness detection will not be perfect in this iteration, but it will be good enough for the first version.

Closes elastic/apm-agent-ruby#613

### DIFF
--- a/lib/elastic_apm/span/context/destination.rb
+++ b/lib/elastic_apm/span/context/destination.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module ElasticAPM
+  class Span
+    class Context
+      # @api private
+      class Destination
+        def initialize(name: nil, resource: nil, type: nil)
+          @name = name
+          @resource = resource
+          @type = type
+        end
+
+        attr_reader :name, :resource, :type
+      end
+    end
+  end
+end
--- a/lib/elastic_apm/span/context/destination.rb
+++ b/lib/elastic_apm/span/context/destination.rb
@@ -12,6 +12,14 @@ module ElasticAPM
         end
 
         attr_reader :name, :resource, :type
+
+        def self.from_uri(uri)
+          new(
+            name: Util.sanitize_url(uri),
+            resource: "#{uri.host}:#{uri.port}",
+            type: 'external'
+          )
+        end
       end
     end
   end
--- a/lib/elastic_apm/span/context/destination.rb
+++ b/lib/elastic_apm/span/context/destination.rb
@@ -13,7 +13,9 @@ module ElasticAPM
 
         attr_reader :name, :resource, :type
 
-        def self.from_uri(uri, type: 'external')
+        def self.from_uri(uri_or_str, type: 'external')
+          uri = normalize(uri_or_str)
+
           new(
             name: only_scheme_and_host(uri),
             resource: "#{uri.host}:#{uri.port}",
@@ -22,10 +24,19 @@ module ElasticAPM
         end
 
         def self.only_scheme_and_host(uri_or_str)
-          uri = uri_or_str.is_a?(URI) ? uri_or_str.dup : URI(uri_or_str)
+          uri = normalize(uri_or_str)
           uri.path = ''
           uri.password = uri.query = uri.fragment = nil
           uri.to_s
+        end
+
+        class << self
+          private
+
+          def normalize(uri_or_str)
+            return uri_or_str.dup if uri_or_str.is_a?(URI)
+            URI(uri_or_str)
+          end
         end
       end
     end
--- a/lib/elastic_apm/span/context/destination.rb
+++ b/lib/elastic_apm/span/context/destination.rb
@@ -13,12 +13,19 @@ module ElasticAPM
 
         attr_reader :name, :resource, :type
 
-        def self.from_uri(uri)
+        def self.from_uri(uri, type: 'external')
           new(
-            name: Util.sanitize_url(uri),
+            name: only_scheme_and_host(uri),
             resource: "#{uri.host}:#{uri.port}",
-            type: 'external'
+            type: type
           )
+        end
+
+        def self.only_scheme_and_host(uri_or_str)
+          uri = uri_or_str.is_a?(URI) ? uri_or_str.dup : URI(uri_or_str)
+          uri.path = ''
+          uri.password = uri.query = uri.fragment = nil
+          uri.to_s
         end
       end
     end
--- a/lib/elastic_apm/spies/http.rb
+++ b/lib/elastic_apm/spies/http.rb
@@ -19,17 +19,31 @@ module ElasticAPM
             method = req.verb.to_s.upcase
             host = req.uri.host
 
+            destination =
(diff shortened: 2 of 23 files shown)


lib/elastic_apm/spies/net_http.rb
```diff
@@ -36,30 +38,49 @@ def request(req, body = nil, &block)
...
 
-            host ||= address
+            cls = use_ssl? ? URI::HTTPS : URI::HTTP
+            uri = cls.build([nil, host, port, path, query, nil])
 
-            name = "#{method} #{host}"
+            context =
+              ElasticAPM::Span::Context.new(
+                http: { url: uri, method: method },
+                destination: {
```
axw: should this be using `ElasticAPM::Span::Context::Destination.from_uri`?

spec/elastic_apm/span/context/destination_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module ElasticAPM
+  class Span
+    class Context
+      RSpec.describe Destination do
+        describe '.from_uri' do
+          let(:uri) { URI('http://example.com') }
```
axw: perhaps add a path/query to the URI, to show that they don't end up in the destination context

spec/elastic_apm/spies/net_http_spec.rb
```diff
@@ -86,5 +106,44 @@ module ElasticAPM
...
+          ElasticAPM.with_transaction 'Net::HTTP test' do
+            Net::HTTP.start('example.com', 1234) do |http|
+              http.get '/some/path?a=1'
+            end
+          end
+        end
+
+        span, = @intercepted.spans
+
+        expect(span.context.destination.name).to eq 'http://example.com:1234/some/path?a=1'
```
axw: this should just be `http://example.com:1234`

spec/elastic_apm/span/context/destination_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+          its(:name) { is_expected.to eq 'http://example.com' }
+          its(:resource) { is_expected.to eq 'example.com:80' }
+          its(:type) { is_expected.to eq 'external' }
+
+          context 'https' do
+            let(:uri) { URI('https://example.com') }
+            its(:resource) { is_expected.to eq 'example.com:443' }
+          end
+
```
axw: I think it would be good to have another test here with the default port explicitly specified, asserting that it is removed from `destination.service.name`.

lib/elastic_apm/spies/net_http.rb
```diff
@@ -36,30 +38,49 @@ def request(req, body = nil, &block)
...
               return request_without_apm(req, body, &block)
             end
+
             if ElasticAPM::Spies::NetHTTPSpy.disabled?
               return request_without_apm(req, body, &block)
             end
 
-            host, = req['host']&.split(':')
-            method = req.method
+            host = req['host']&.split(':')&.first || address
```
axw: Apologies for my Ruby ignorance: what's the content of `req['host']` when IPv6 is used? Where's this `address` variable come from?

That splitting looks suspiciously like it won't work for IPv6, but I see you have a test for net_http and IPv6. Does it cover all branches?

mikker:
> I think I remember you mentioning somewhere: we aren't using the `destination.address` and `destination.post` field names because it'd be a breaking change, correct?

I don't remember saying anything about breaking changes with regards to this. As far as I understand, adding things is never (very rarely?) _breaking_?

> Would you mind adding the strategy to the description so it's clear what the PR is doing?

Updated the description with a quote from the canonical, cross-agent issue.

> Additionally, why don't we get the host info in the `ActiveRecord` and `Sequel` spies? I know that for `ActiveRecord` at least you have that info in the `payload[:connection]`.

You're thinking for the _old_ `address` field?

> And why is there no change for the `Mongo` spy?

Oversight. Will add 👍

lib/elastic_apm/spies/net_http.rb
mikker: Comes from all the way back in the Opbeat days: https://github.com/opbeat/opbeat-ruby/blob/master/lib/opbeat/injections/net_http.rb#L14 !
I'm not actually really sure it's set anymore. `address` is a method on the object we're currently prying ourselves into (yay Ruby!).

mikker:
> Additionally, why don't we get the host info in the `ActiveRecord` and `Sequel` spies? I know that for `ActiveRecord` at least you have that info in the `payload[:connection]`.

Actually, see the spec how anything but http is supposed to be just their types, mostly.

spec/elastic_apm/span/context/destination_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+            its(:name) { is_expected.to eq 'https://example.com' }
+            its(:resource) { is_expected.to eq 'example.com:443' }
+          end
+
+          context 'non-default port' do
+            let(:uri) { URI('http://example.com:8080/path?a=1') }
+            its(:name) { is_expected.to eq 'http://example.com:8080' }
+            its(:resource) { is_expected.to eq 'example.com:8080' }
+          end
+        end
```
estolfo: Shouldn't we test also the scenario that the argument is a `String`?

lib/elastic_apm/span/context/destination.rb
```diff
@@ -0,0 +1,33 @@
...
+        def self.from_uri(uri, type: 'external')
+          new(
+            name: only_scheme_and_host(uri),
+            resource: "#{uri.host}:#{uri.port}",
+            type: type
+          )
+        end
+
+        def self.only_scheme_and_host(uri_or_str)
+          uri = uri_or_str.is_a?(URI) ? uri_or_str.dup : URI(uri_or_str)
```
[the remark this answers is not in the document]
mikker: Correct 😊

lib/elastic_apm/spies/http.rb
```diff
@@ -19,17 +19,31 @@ def perform(req, options)
...
+              action: method,
+              context: context
             ) do |span|
               trace_context = span&.trace_context || transaction.trace_context
               req['Elastic-Apm-Traceparent'] = trace_context.to_header
-              perform_without_apm(req, options)
+              result = perform_without_apm(req, options)
+
+              if (http = span&.context&.http)
+                http.status_code = result.status.to_s
```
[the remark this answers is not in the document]
mikker: Yes, it was just missing.

spec/elastic_apm/span/context/destination_spec.rb
mikker: Good thing we did – there was an error 😊

PR #651, opened 2019-12-10
Add missing parameters to documentation in elastic_apm.rb

estolfo:
I noticed these minor documentation issues when adding the `sync` parameter in [another PR](https://github.com/elastic/apm-agent-ruby/pull/571).


lib/elastic_apm.rb
```diff
@@ -203,8 +209,12 @@ def end_span
     #
     # @param name [String] A description of the span, eq `SELECT FROM "users"`
     # @param type [String] The kind of span, eq `db.mysql2.query`
```
mikker:
```suggestion
    # @param type [String] The kind of span, eq `db`
```

PR #655, opened 2019-12-12
Support api keys

### LINKED ISSUE #654: Support setting and sending api key it in the Authorization header
See https://github.com/elastic/apm/issues/183 for spec

estolfo:
This PR adds support for api keys. A user can set a `api_key` Base64 encoded string in the config.
If both a `secret_token` and `api_key` are set, the `api_key` will be sent.

Todo:
- [x] Changelog entry
- [x] end-to-end testing

Closes elastic/apm-agent-ruby#654


features/api_key.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Api Key
```
alvarolobato: Great to see this. ❤️

features/api_key.feature
```diff
@@ -0,0 +1,15 @@
...
+
+  Scenario: A configured api key is sent in the Authorization header
+    When an api key is set to 'RTNxMjlXNEJt' in the config
+    Then the Authorization header is 'ApiKey RTNxMjlXNEJt'
+
+  Scenario: An configured api key takes precedence over a secret token
+    When an api key is set in the config
+    When a secret_token is set in the config
```
mdelapenya:
```diff
-    When a secret_token is set in the config
+    And a secret_token is set in the config
```

features/api_key.feature
```diff
@@ -0,0 +1,15 @@
...
+    Then the api key is sent in the Authorization header
+
+  Scenario: A configured secret token is sent if no api key is configured
+    When a secret_token is set in the config
+    When an api key is not set in the config
```
mdelapenya: Same here

features/api_key.feature
estolfo: thanks!

docs/configuration.asciidoc
```diff
@@ -145,7 +145,23 @@ One example to generate a secure secret token is:
...
+
+[options="header"]
+|============
+| Environment           | `Config` key | Default | Example
+| `ELASTIC_APM_API_KEY` | `api_key`    | `nil`   | A base64-encoded string
+|============
+
+This base64-encoded string is used to ensure that only your agents can send data to your APM server.
+You must have created the api key using the APM server command line tool. Please see the APM server
+documentation for details on how to do that.
```
estolfo: I don't know if this is exactly true yet. I'll update it when the tooling is available in the apm server.

Gemfile
```diff
@@ -6,6 +6,7 @@ git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
 
 # Tools
 gem 'bootsnap', require: false
+gem 'cucumber', require: false, group: :test
```
mikker: I see we already have a `:test` group but it really doesn't make much sense? _Everything_ in the `Gemfile` is only for testing.

features/step_definitions/stepdefs.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+Given('an agent') do
```
mikker: Just a question: Can you specify teardown instructions to a `Given` step? In this case for example, when it actually boots an agent, we'd prefer it to stop it again afterwards.

mikker: Follow up – do we want it to say `Given('a config')` instead? Isn't that in reality what we're testing? Then have it set up `@config`. That way we can omit `||=` in every `When` below. This feels a bit brittle to me. Like how the (other) specs have a tendency to easily blend into each other.

estolfo:
I did split up the docs/implementation: I removed the docs in `configuration.asciidoc` and will put that in a separate PR.

Gemfile
estolfo: I was following the convention that we have for `simplecov`. I don't feel strongly about whether we use the `:test` group...so it's your call :)

mikker: Let's remove them all then 💅

features/step_definitions/stepdefs.rb
estolfo: hmm, I need to think about how best to do this. I added the `Given an agent` step because Python requested it but the Ruby agent needs the config to be defined before/with the agent. I'll look into how we can adjust this to avoid the `||=` in every step.

estolfo:
Test failures are CI issues

PR #657, opened 2019-12-18
Allow action dispatch spy to be disabled

estolfo:


lib/elastic_apm/rails.rb
```diff
@@ -30,13 +30,6 @@ def start(config)
         attach_subscriber(agent)
       end
 
-      if ElasticAPM.running? &&
-         !ElasticAPM.agent.config.disabled_instrumentations.include?(
-           'action_dispatch'
-         )
-        require 'elastic_apm/spies/action_dispatch'
-      end
```
mikker: Why are we _removing_ this?

estolfo: Because the spy is automatically required now, like the others when `action_dispatch` is loaded. Am I missing something?

mikker: Oh right, no, I was missing something. Sorry!

estolfo: This line https://github.com/elastic/apm-agent-ruby/pull/657/files#diff-eebd381bf9bfbb2a1f22c795c4add2a6R121 makes it so that the spy is required when `action_dispatch` (the original) is required. I thought that's how all the other spies work?

mikker: Yes yes, you're right. But you forgot to sort the list alphabetically 😉

estolfo: now I don't feel so bad about my grammar code review suggestion this morning!

PR #664, opened 2019-12-25
Fix Console top-level namespace

olegantonyan:
If you have `Console` top-level constant defined, then `::Rails.const_defined? 'Rails::Console'` will always return true. This is how ruby handles namespaces: traverse all the way to the top-level. To prevent this you have to pass optional argument set to false. In this case `::Rails.const_defined?('Console', false)` we're telling Ruby to search only `Rails` namespace and do not go outside.


mikker:
jenkins, please run the tests

mikker:
jenkins, please run the tests

mikker:
Test failure is unrelated.

mikker:
Thanks for this @olegantonyan!

PR #668, opened 2020-01-03
Include number of `rows_affected` in span db context for Sequel

### LINKED ISSUE #574: Include number of `rows_affected` if available
https://github.com/elastic/apm/issues/112
APM Server issue: https://github.com/elastic/apm-server/pull/3095 (✅ merged!)

Have not investigated whether this is something we can get easily or at all, so I suppose that will be the first step.

estolfo:
Closes elastic/apm-agent-ruby#574 

Changes in this PR include:
- Use `start_span` and `end_span` instead of `with_span` in the Sequel spy because the span's db context needs to be updated before it is queued for transport.
- Add `rows_affected` attribute on `Span::Context::DB` object
- Add `rows_affected` field in serialized span JSON object

** Opportunity for improvement: it'd be nice if we could somehow use the SQL summarizer to check if the statement is an `update` or `delete` to avoid using a regex [here](https://github.com/elastic/apm-agent-ruby/pull/668/files#diff-01afcd3d0b7119d1e312d5089a5bd3e6R48). Maybe we could return a parsed object instead of a `String` from the new SQL summarizer that you can call `to_s` on and other methods like `update?` or `delete?`.


estolfo:
Yes, I agree this is fine for now. Maybe we can change the summarizer to return a value object when we've tested the new one enough and switch over to using it as the default.

I have no problem leaving this and waiting to merge until the server completes its implementation. 👍🏼

PR #669, opened 2020-01-03
Add gRPC support

### LINKED ISSUE #628: Add gRPC support
placeholder,TBD

estolfo:
This PR adds support for gRPC. It instruments unary request/response calls by adding `ClientInterceptor` and `ServerInterceptor` classes.

Todo:

- [x] Figure out if more info can be extract from the intercept method arguments and added to the transaction/span contexts.
- [x] Documentation
- [x] Changelog
- [x] Tests for `traceparent` and `tracestate` propagation

Closes #628


mikker:
Re: the first question – I don't mind the manual work of inserting the agent _interceptor_. It's similar to how Sinatra/Rack users need to insert the middleware manually too. I think the code in your description looks completely fine 👍

axw:
> What should the context schema be for transactions?

If you can access the underlying HTTP/2 request, it would be great if the regular HTTP context could be added. Otherwise, I'd be inclined to focus just on getting trace propagation working and worry about filling in context later on.

> Should we only add interceptors for unary request/response calls? I.e. should we only implement the request_response on the Interceptors? Go seems to do that.

Yes, I think so. We _might_ later want to add metrics for gRPC streams, but there hasn't been any demand for it and request/response is the most important thing cover.

spec/elastic_apm/grpc_spec.rb
```diff
@@ -0,0 +1,269 @@
...
+          expect(transaction.result).to eq('success')
+          expect(transaction.trace_id).to eq(trace_context.trace_id)
+          expect(transaction.parent_id).to eq(trace_context.id)
+
+          server.stop
+          thread.kill
+        end
+      end
+
+      context 'when there\'s and error' do
```
mikker:
```suggestion
      context "when there's an error" do
```

CHANGELOG.asciidoc
```diff
@@ -42,7 +42,7 @@ endif::[]
 [float]
 ===== Added
 
-- Support for gRPC using the `grpc` gem {pull}669[#669]
+- Support for gRPC using the `grpc` gem (Experimental) {pull}669[#669]
```
mikker: 👍

PR #653, opened 2019-12-11
Serialize Span Context labels

eric-pigeon:
Include labels when sending a Span#context to the APM server

elastic/apm-agent-ruby#652

PR #671, opened 2020-01-07
Serialize Span Context labels

mikker:
Bringing this home. Merged `master`, changed to only include labels in the payload when they are set.

Continued from #653

Closes elastic/apm-agent-ruby#653


mikker:
CI failures are unrelated.

PR #656, opened 2019-12-13
Add Shoryuken worker instrumentation

nikriek:
This PR adds support for instrumentation of Shoryuken background processing. I tested the implementation with a running Elastic APM server with different scenarios. 

For me, it's still unclear how sophisticated the specs should be. The following ones are possible: 
- Testing the the middleware interface directly without shoryuken/SQS interaction 
- Testing the the complete shoryuken middleware chain with mocking api requests to SQS
- Testing the the complete shoryuken middleware chain using a mocked SQS service in Docker

Currently, I started with the second approach. But since docker compose is used for the test suite, one might prefer the last one.

PR #674, opened 2020-01-08
Add support for shoryuken (AWS SQS)

mikker:
Continued from elastic/apm-agent-ruby#656


mikker:
CI error is only a GH timeout.

PR #676, opened 2020-01-08
Add support for Sneakers

mikker:
Continued from elastic/apm-agent-ruby#599 

Updating the implementation and spec a little bit. I don't feel we need to include version checking in the spec unless explicitly testing with different versions.

@redwrx has an open PR here (https://github.com/jondot/sneakers/pull/414) to pass the worker itself to middleware. Until that is merged I don't think there's a way to get the worker name for the transaction names. Too bad, and we fall back on the queue name.

@estolfo, how do you feel about marking Sneakers support as _experimental_ until a way to get the worker name is provided? That way we won't need to do the whole version dance around breaking changes when we change the names of all Sneakers transactions?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,7 @@ gem 'shoryuken', require: nil
 gem 'sidekiq', require: nil
 gem 'simplecov', require: false, group: :test
 gem 'simplecov-cobertura', require: false, group: :test
+gem 'sneakers', require: nil
 gem 'yard', require: nil
 gem 'yarjuf'
 
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,7 @@ gem 'shoryuken', require: nil
 gem 'sidekiq', require: nil
 gem 'simplecov', require: false, group: :test
 gem 'simplecov-cobertura', require: false, group: :test
-gem 'sneakers', require: nil
+gem "sneakers", "~> 2.12.0", require: nil
 gem 'yard', require: nil
 gem 'yarjuf'
 
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,7 @@ gem 'shoryuken', require: nil
 gem 'sidekiq', require: nil
 gem 'simplecov', require: false, group: :test
 gem 'simplecov-cobertura', require: false, group: :test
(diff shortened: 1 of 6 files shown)


estolfo:
I think there's no harm in making it experimental for now. But I would also write a comment on the Sneakers PR you referenced to let them know you're hoping that it'll be merged soon. That seemed to work for this https://github.com/jondot/sneakers/issues/367

Gemfile
```diff
@@ -29,6 +29,7 @@ gem 'shoryuken', require: nil
 gem 'sidekiq', require: nil
 gem 'simplecov', require: false, group: :test
 gem 'simplecov-cobertura', require: false, group: :test
+gem "sneakers", "~> 2.12.0", require: nil
```
estolfo: We could also make this `2.12` so that we'll just test against the latest in the 2.x series. Do you see a reason why we should only test `2.12.x`?

docs/supported-technologies.asciidoc
```diff
@@ -85,3 +85,4 @@ We automatically instrument background processing using:
 - DelayedJob
 - Sidekiq
 - Shoryuken
+- Sneakers (v2.1.2+) (Experimental, see {pull}676[#676])
```
estolfo: This should be `2.12.0`, right?

Gemfile
mikker:
```suggestion
gem "sneakers", "~> 2.12", require: nil
```

docs/supported-technologies.asciidoc
mikker:
```suggestion
- Sneakers (v2.12.0+) (Experimental, see {pull}676[#676])
```

Gemfile
```diff
@@ -29,6 +29,7 @@ gem 'shoryuken', require: nil
 gem 'sidekiq', require: nil
 gem 'simplecov', require: false, group: :test
 gem 'simplecov-cobertura', require: false, group: :test
+gem "sneakers", "~> 2.12", require: nil
```
estolfo:
```suggestion
gem 'sneakers', '~> 2.12', require: nil
```

mikker: Damn you Rubocop! 😅

PR #683, opened 2020-01-13
ci(jenkins): even load distribution for the timeouts

v1v:

Run downstream jobs with some random sleep.

## Why is it important?

Let's see if this particular change reduces the issues with the tiemout when there are a bunch of builds asking for git checkouts

## Related issues
closes #ISSUE

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -94,8 +94,9 @@ pipeline {
             script {
               def ruby = readYaml(file: '.ci/.jenkins_ruby.yml')
               def testTasks = [:]
+              def i = 0
               ruby['RUBY_VERSION'].each{ rubyVersion ->
-                testTasks[rubyVersion] = { runJob(rubyVersion) }
+                testTasks[rubyVersion] = { runJob(rubyVersion, i++) }
               }
               parallel(testTasks)
               }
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -285,14 +286,15 @@ def runBenchmark(version){
   }
 }
 
-def runJob(rubyVersion){
+def runJob(String rubyVersion, int sleep = 1){
   try {
+    def quiet = (sleep * randomNumber(min: 1, max: 5)) + 5
```
v1v: There were different options:
- Uniform sleep.
- Exponential sleep.
- Random sleep.

.ci/Jenkinsfile
```diff
@@ -94,8 +94,9 @@ pipeline {
             script {
               def ruby = readYaml(file: '.ci/.jenkins_ruby.yml')
               def testTasks = [:]
+              def i = 0
               ruby['RUBY_VERSION'].each{ rubyVersion ->
```
mdelapenya: You can use `eachWithIndex` http://docs.groovy-lang.org/latest/html/groovy-jdk/java/lang/Object.html#eachWithIndex(groovy.lang.Closure)

kuisathaverat: I'd recommend to not use it, this is not groovy, it is groovy-ish, you have to try to use as lees as possible grovy extensions and keep the code on basics groovy features to avoid problems with the script security plugin and the DSL implementation, also it probably isn't implemented on the DSL plugin.

v1v:
I'm not fully sure this particular PR will do any improvement thought... 

I did manage to monitor there were about 43 builds in a 4 hours window and about 3 of them did fail with the timeout, ~8%, that's quite a lot. 

My assumption is that this is related to the `overall` number of github connections from the whole GCP infrastructure. If the GCP is a multitenant space then we don't know what others are doing, in fact, I do know other teams are using a retry up to 9 times with some sleep. So if the threshold is the total number of connections then we cannot do anything with this PR.

v1v:
Current failure related to the well-known timeout:
- https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-ruby%2Fapm-agent-ruby-downstream/detail/PR-683/41/pipeline

So I'll proceed with the merge

PR #685, opened 2020-01-13
Handle case when headers are nil in response

estolfo:
We got this error in a test:

```ruby
undefined method `each_with_object' for nil:NilClass
["./lib/elastic_apm/context/response.rb:24:in `headers='",...
```

So we must check if `headers` is `nil` before processing them.


mikker:
The bug was released in 3.4.0. We should put out a patch release with this.

PR #688, opened 2020-01-14
Minimize the amount of Ruby 2.3 test branches

mikker:
Ruby 2.3 has reached EOL, so we're minimizing the count of framework combinations under test. Support will officially be dropped in next major release.

### DIFF
--- a/.ci/.jenkins_exclude.yml
+++ b/.ci/.jenkins_exclude.yml
@@ -1,4 +1,27 @@
 exclude:
+  # Ruby 2.3
+  # Only includes rails-5.2, sinatra-2.0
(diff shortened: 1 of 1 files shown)


.ci/.jenkins_exclude.yml
```diff
@@ -1,4 +1,27 @@
 exclude:
+  # Ruby 2.3
+  # Only includes rails-5.2, sinatra-2.0
```
estolfo: thanks for this comment, it's hard to see what _is_ included

mikker: Definitely. This format is very hard to wrap ones head around.

mikker:
An actual green build 🌳

PR #690, opened 2020-01-17
ci(jenkins): rebuild downstream if timeout issue

v1v:

Enable to rebuild one more time in case of a timeout issue.

## Why is it important?

Add CI resilience to any kind of environmental issue.


v1v:
This is not required anymore, as the retry for the checkout is not handled with our shared library

PR #694, opened 2020-01-21
Support TraceContext and tracestate

### LINKED ISSUE #621: Fully support TraceContext
See https://github.com/elastic/apm/issues/71

mikker:
Closes #621.
See https://github.com/elastic/apm/issues/71 for details.

### DIFF
--- a/lib/elastic_apm/middleware.rb
+++ b/lib/elastic_apm/middleware.rb
@@ -57,7 +57,11 @@ module ElasticAPM
 
     def trace_context(env)
       return unless (header = get_traceparent_header(env))
-      TraceContext.parse(header)
+
+      TraceContext.parse(header).tap do |tc|
+        next unless (header = env['HTTP_TRACESTATE'])
+        tc.tracestate = TraceContext::Tracestate.parse(header)
+      end
     rescue TraceContext::InvalidTraceparentHeader
       warn "Couldn't parse invalid traceparent header: #{header.inspect}"
       nil
--- a/lib/elastic_apm/middleware.rb
+++ b/lib/elastic_apm/middleware.rb
@@ -56,26 +56,9 @@ module ElasticAPM
     end
 
     def trace_context(env)
-      return unless (header = get_traceparent_header(env))
-
-      TraceContext.parse(header).tap do |tc|
-        next unless (header = env['HTTP_TRACESTATE'])
-        tc.tracestate = TraceContext::Tracestate.parse(header)
-      end
-    rescue TraceContext::InvalidTraceparentHeader
-      warn "Couldn't parse invalid traceparent header: #{header.inspect}"
-      nil
-    end
-
-    def get_traceparent_header(env)
-      if (header = env['HTTP_ELASTIC_APM_TRACEPARENT'])
-        return header
-      end
-
-      if (header = env['HTTP_TRACEPARENT'])
-        return header
-      end
-
+      TraceContext.parse(env: env)
+    rescue TraceContext::InvalidTraceparentHeader => e
+      warn e.message
       nil
     end
 
--- a/lib/elastic_apm/middleware.rb
+++ b/lib/elastic_apm/middleware.rb
@@ -56,10 +56,22 @@ module ElasticAPM
     end
 
     def trace_context(env)
-      return unless (header = env['HTTP_ELASTIC_APM_TRACEPARENT'])
+      return unless (header = get_traceparent_header(env))
       TraceContext.parse(header)
     rescue TraceContext::InvalidTraceparentHeader
       warn "Couldn't parse invalid traceparent header: #{header.inspect}"
+      nil
+    end
+
+    def get_traceparent_header(env)
+      if (header = env['HTTP_ELASTIC_APM_TRACEPARENT'])
+        return header
+      end
+
+      if (header = env['HTTP_TRACEPARENT'])
+        return header
+      end
+
       nil
     end
 
--- a/lib/elastic_apm/trace_context.rb
+++ b/lib/elastic_apm/trace_context.rb
@@ -35,6 +35,7 @@ module ElasticAPM
     attr_accessor :version, :id, :trace_id, :parent_id, :recorded, :tracestate
 
     alias :recorded? :recorded
+
     def self.parse(header)
       raise InvalidTraceparentHeader unless header.length == 55
       raise InvalidTraceparentHeader unless header[0..1] == VERSION
--- a/lib/elastic_apm/trace_context.rb
+++ b/lib/elastic_apm/trace_context.rb
@@ -1,83 +1,83 @@
 # frozen_string_literal: true
 
 require 'elastic_apm/trace_context/tracestate'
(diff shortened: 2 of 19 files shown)


mikker:
Spec error is one of the usual unrelated jruby ones.

lib/elastic_apm/middleware.rb
```diff
@@ -56,13 +56,29 @@ def start_transaction(env)
...
     def trace_context(env)
-      return unless (header = env['HTTP_ELASTIC_APM_TRACEPARENT'])
-      TraceContext.parse(header)
+      return unless (header = get_traceparent_header(env))
+
+      TraceContext.parse(header).tap do |tc|
+        next unless (header = env['HTTP_TRACESTATE'])
+        tc.tracestate = TraceContext::Tracestate.parse(header)
+      end
     rescue TraceContext::InvalidTraceparentHeader
```
estolfo: What about errors that might be raised from parsing the `tracestate` header? Those are just raised?

I also find it confusing that there is a `TraceContext::Tracestate` object but calling `parse` on `TraceContext` parses the `traceparent` headers. Does the term "Trace Context" only apply to `traceparent` headers? If not, it would make more sense to have a `TraceContext::TraceParent` object.

In any case, there are two `parse` calls in this method: `TraceContext.parse(header)` . and `TraceContext::Tracestate.parse(header)`. Because only the first one can raise the `TraceContext::InvalidTraceparentHeader` error, it makes more sense to me to have the flow be:

parse the traceparent: `TraceContext.parse(header)`
catch the error: `rescue TraceContext::InvalidTraceparentHeader`
parse the tracestate: `TraceContext::Tracestate.parse(header)`

But maybe we should define an error that can be raised by parsing the `tracestate` and we should catch it.

lib/elastic_apm/trace_context/tracestate.rb
```diff
@@ -0,0 +1,26 @@
...
+      end
+
+      attr_accessor :values
+
+      def self.parse(header)
+        # HTTP allows multiple headers with the same name, eg. multiple
+        # Set-Cookie headers per response.
+        # Rack handles this by joining the headers under the same key, separated
+        # by newlines, see https://www.rubydoc.info/github/rack/rack/file/SPEC
+        new(String(header).split("\n"))
```
estolfo: just curious why you used `String(..)` when `to_s` seems more commonly used in Ruby?

spec/elastic_apm/trace_context_spec.rb
```diff
@@ -111,6 +111,68 @@ module ElasticAPM
...
+      subject do
+        described_class.new.tap do |tp|
+          tp.trace_id = '1' * 32
+          tp.id = '2' * 16
+          tp.flags = '00000011'
+        end
+      end
+
+      context 'when prefixed disable', :intercept do
+        it 'yields twice' do
```
estolfo: What does it mean "yields twice"? `calls` is validated to be 1 below...?

lib/elastic_apm/middleware.rb
```diff
@@ -56,13 +56,29 @@ def start_transaction(env)
     end
 
     def trace_context(env)
-      return unless (header = env['HTTP_ELASTIC_APM_TRACEPARENT'])
-      TraceContext.parse(header)
+      return unless (header = get_traceparent_header(env))
```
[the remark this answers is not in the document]
mikker: I thought about this but I don't see how that would ever happen, if distributed tracing should still work. Without the `traceparent` ids, how would you know what trace the `state` belonged to?

spec/elastic_apm/trace_context_spec.rb
mikker: It was a lazy description. Updated it.

lib/elastic_apm/trace_context/tracestate.rb
mikker: I think it comes from how I often do `Array(value)` on passed arguments if I want to make sure it's an array.

Instead of doing `ArgumentError`, just turn it into a list.

In this case I don't think there's a difference in how `String(...)` and `.to_s` work(?). Happy to change it if you think this looks weird.

lib/elastic_apm/middleware.rb
mikker: I see your point 👍

mikker: If you just think of `TraceContext` as the `Traceparent` object, it makes more sense?

mikker: Re: raising. `InvalidTraceparentHeader` is an exception we raise ourselves if the header doesn't match the strictly defined shape. I don't think we need strict parsing for the `Tracestate` header as it's more loosely and the only thing we do with it (for now) is passing it on as is.

mikker:
I pulled out `Traceparent` into its own class. `TraceContext` is referred to quite a few places and is sort of in a gray area public api-wise. Hence there's some redirection from context to -parent.

lib/elastic_apm/trace_context/tracestate.rb
estolfo: I don't think there's a difference. I don't mind either way, I was just wondering if there was a benefit I was missing.

lib/elastic_apm/trace_context.rb
```diff
@@ -1,82 +1,107 @@
...
+      traceparent.parent_id
     end
 
     def ensure_parent_id
-      @parent_id ||= hex(ID_LENGTH)
+      traceparent.ensure_parent_id
+    end
+
+    def recorded?
+      traceparent.recorded?
```
estolfo: Each of these methods could also be forwarded to `traceparent` to make the class smaller : )

spec/elastic_apm/trace_context/traceparent_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+        end
+
+        its(:traceparent) { is_expected.to be_a TraceContext::Traceparent }
+      end
+    end
+  end
+
+  RSpec.describe TraceContext::Traceparent do
+    describe '.new' do
+      let(:transaction) { Transaction.new }
```
estolfo: I don't think this is used in the test.

lib/elastic_apm/trace_context.rb
```diff
@@ -1,82 +1,107 @@
...
+          return header
+        end
 
-    def hex(len)
-      SecureRandom.hex(len)
+        if (header = env['HTTP_TRACEPARENT'])
+          return header
+        end
+
+        nil
```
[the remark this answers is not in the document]
mikker: It's so funny how you loose the big picture when changing just a little bit again and again 😅

spec/elastic_apm/spies/net_http_spec.rb
```diff
@@ -56,11 +56,12 @@ module ElasticAPM
       expect(http.status_code).to match('200')
     end
 
-    it 'adds traceparent header' do
+    it 'adds both traceparent headers' do
```
mikker:
```suggestion
    it 'adds both TraceContext headers' do
```

PR #695, opened 2020-01-23
Rename and deprecate disableD_instrumentations

### LINKED ISSUE #658: Rename DISABLED_INSTRUMENTATIONS to DISABLE_… to align with other agents
Added a _d_ by mistake. I say we add the correct one alongside the current one, soft-deprecate the misspelled version with a deprecation warning and redirect its value.

mikker:
Closes #658.

At first I added the deprecation warning to the reader method too, but figured one warning on app boot was enough.

### DIFF
--- a/lib/elastic_apm/config.rb
+++ b/lib/elastic_apm/config.rb
@@ -37,7 +37,7 @@ module ElasticAPM
     option :disable_metrics,                   type: :list,   default: [],      converter: WildcardPatternList.new
     option :disable_send,                      type: :bool,   default: false
     option :disable_start_message,             type: :bool,   default: false
-    option :disabled_instrumentations,         type: :list,   default: %w[json]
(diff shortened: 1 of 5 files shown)


lib/elastic_apm/config.rb
```diff
@@ -156,6 +156,16 @@ def collect_metrics?
...
+    def disabled_instrumentations
+      disable_instrumentations
+    end
+
+    def disabled_instrumentations=(value)
+      warn '[DEPRECATED] The option disabled_instrumentations has been ' \
+        'renamed to disable_instrumentations to align with other agents.'
+      self.disable_instrumentations = value
+    end
+
```
estolfo: Can't you just use the `Deprecations` module here?

mikker: I could perhaps but I wanted a special warning message.

PR #697, opened 2020-01-24
Enable docker image cache in the CI workers

v1v:

Build docker images and cache them in the CI workers using the `.ci/.jenkins_ruby.yml` as the reference to the docker images to be built

## Why is it important?

Speed up the CI overall execution time

## Related issues
closes #ISSUE

## How to test this

```bash
hub checkout pr 697
sh -x .ci/packer_cache.sh
```
_NOTE_: `/usr/local/bin/bash_standard_lib.sh` is required to exist. That particular requirement is already in place for the CI workers.

## Question
- Using a docker volume when running the docker compose could help with the caching? Or that's something already solved when building the docker image?

### DIFF
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+source /usr/local/bin/bash_standard_lib.sh
+
+grep "-" .ci/.jenkins_ruby.yml | cut -d'-' -f2- | \
+while read -r version;
+do
+    (retry 2 ./spec/scripts/docker_build.sh "${version}") || echo "Error building ${version} Docker image, we continue"
+done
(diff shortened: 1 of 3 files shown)


kuisathaverat:
I'd preferred to build the image in a job that we can have a report of the result and pull the image to our namespace in the docker registry, as we make with other images, then we can make the pull of the image directly to the packer cache.

v1v:
> I'd preferred to build the image in a job that we can have a report of the result and pull the image to our namespace in the docker registry, as we make with other images, then we can make the pull of the image directly to the packer cache.

I was tempted to move in that direction, as we do have in other places, before using this approach, but I thought it won't be ideal as contributors don't have access to our docker registry. I know we do use it even with that particular use case.


I'll raise another PR with the suggested approach then we can analyze if there are any blockers.

WDYT?

v1v:
I'll close this PR as there is an ongoing process to cache the docker images in another pipeline.

PR #681, opened 2020-01-10
Add thread-safety stress test for Metrics

mikker:
... doesn't fail. Can you spot anything to improve, @estolfo?

PR #699, opened 2020-01-27
Adjust metric collect method and run multi-thread tests

estolfo:
This PR adjusts the `#collect` method in the `Metric` class so that reading the `@value` and then resetting it is atomic. We were missing some counts because the `@value` variable was being updated by another thread between when it was cached and reset.

Continues https://github.com/elastic/apm-agent-ruby/pull/681


estolfo:
@mikker I copied the tests from your PR here: #681
What do you want to leave out of this one?

mikker:
So good 👍 

We probably need a changelog entry too.

PR #711, opened 2020-02-13
adding missing command `install` for gem installation

insukcho:

Fixing `elastic-apm` installation guide to add missing command `install`

## Why is it important?

It will be not working when following the guide


insukcho:
Sorry for the noise. I'm not that familiar with rub and gemfile and it seems I was totally wrong.

PR #712, opened 2020-02-13
add missing command install

insukcho:

Fixing `elastic-apm` installation guide to add missing command install

## Why is it important?

The user will be not available to install `elastic-apm` with following this guide


insukcho:
Sorry for the noise. I'm not that familiar with rub and gemfile and it seems I was totally wrong.

PR #715, opened 2020-02-14
[docs] Add EOL table to documentation

bmorelli25:

This PR defines end of life and end of maintenance dates for the Ruby Agent. EOL dates were determined by going through old releases, finding the GA release date of the major Ruby Agent versions, and adding 18 months.

## Why is it important?

We love all our products, but sometimes we must say goodbye to a release so that we can continue moving forward on future development and innovation.

## Checklist

~~- [ ] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).~~
~~- [ ] My code follows the style guidelines of this project (See `.rubocop.yml`)~~
~~ [ ] I have added tests that prove my fix is effective or that my feature works~~
~~- [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-ruby/blob/master/CONTRIBUTING.md#testing) pass locally with my changes~~
~~- [ ] I have updated [CHANGELOG.asciidoc](CHANGELOG.asciidoc)~~
~~- [ ] I have updated [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)~~
~~- [ ] Added an API method or config option? Document in which version this will be introduced~~

## Related issues

For elastic/apm#173.

## Screenshots

Screen Shot 2020-02-14 at 10 59 05 AM


estolfo:
It'll be great to have this info documented, thank you!
What is the difference between EOL and _maintained until_? What would it mean if the EOL date is reached before the next major version? Also, does _maintained_ mean bug fixes and security patches?

bmorelli25:
> What is the difference between EOL and maintained until? What would it mean if the EOL date is reached before the next major version? Also, does maintained mean bug fixes and security patches?

This comment should answer your questions: https://github.com/elastic/apm/issues/173#issuecomment-558541355. 

> What would it mean if the EOL date is reached before the next major version?

That's an interesting question... and one that the [stack maintenance policy](https://www.elastic.co/support/eol) doesn't address (because they have a set release schedule). I guess it means you just need to release a major version more often 😉. I suppose we could always just extend EOL by 6 months if that were to happen?

bmorelli25:
Does it make more sense to wait to define EOL for the current major until after a new major has been released?

mikker:
Yes, good point 👍
Let's keep the current without one.

bmorelli25:
So. I had the EOL table wrong in the first version. EOL should be for minors as well -- not just majors. This new table has more realistic dates, and avoids the problem of hitting EOL before the next major.

mikker:
Thank you!

I take it we don't necessarily have to do anything when we eventually hit the dates? If stuff still works, it still just works, right? But then when things eventually fail, we _have warned_ folks that it might happen?

bmorelli25:
> I take it we don't necessarily have to do anything when we eventually hit the dates? If stuff still works, it still just works, right? But then when things eventually fail, we have warned folks that it might happen?

Correct!

PR #719, opened 2020-02-20
Move framework specific requires to spies

mikker:
Our current solution depends on the `elastic-apm` gem to be loaded _after_ Grape or Sinatra.

This PR uses the already present mechanism of hooking into require to make sure to load regardless of gem order.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -59,7 +59,7 @@ end
 frameworks_versions.each do |framework, version|
   case version
   when 'master'
-    gem framework, github: GITHUB_REPOS.fetch[framework]
+    gem framework, github: GITHUB_REPOS.fetch(framework)
   when /.+/
     gem framework, "~> #{version}.0"
   else
--- a/lib/elastic_apm/spies/grape.rb
+++ b/lib/elastic_apm/spies/grape.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module ElasticAPM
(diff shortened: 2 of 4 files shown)


estolfo:
Looks good 👍
Do we have to make any changes in the tests?

mikker:
My hope was that it would _Just Work™_ but let's see.

estolfo:
Some more comments:
I think we have to add `grape` to [here](https://github.com/elastic/apm-agent-ruby/blob/master/lib/elastic_apm/config.rb#L112-L127)
We can also remove [these](https://github.com/elastic/apm-agent-ruby/blob/master/lib/elastic_apm.rb#L27-L28) lines.

mikker:
On second thought, this doesn't cut it. As the spies aren't required before here https://github.com/mikker/apm-agent-ruby/blob/af2357eb09ef68e5469facdf50752a2639a85316/lib/elastic_apm/agent.rb#L103-L106 `ElasticAPM::Grape` won't be available before starting the agent.

So, do we add `require 'elastic_apm/grape'` to the installation instructions (maybe, probably, breaking change?) or keep it all as is?

I think I'd say option 2.

estolfo:
I would hesitate to add a breaking change for something no one has complained about. So maybe we should just leave it as-is?

mikker:
Wholeheartedly agree, closing.

PR #721, opened 2020-02-20
Add GraphQL support

mikker:
According to spec (still experimental): https://github.com/elastic/apm/pull/208

This PR adds basic support for tracing GraphQL based APIs.


spec/integration/graphql_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+            json: {
+              error: { message: e.message, backtrace: e.backtrace },
+              data: {}
+            }
+          )
+        end
+      end
+      # rubocop:enable Style/ClassAndModuleChildren
+
+      MockIntake.stub!
```
mikker: @estolfo I have to say, after the dust has settled on the spec suite metrics woes, I like the old `MockIntake` api better than the new, (imo) convoluted setup 😐

estolfo: I don't mind what we use, as long as it works 😄

estolfo: ps what is the "old API"? I lost track of the iterations.

mikker: `MockIntake` _is_ the old API. _All the equivalent things_ we do in the Rails spec is the new API.

.rubocop.yml
```diff
@@ -98,5 +98,12 @@ Style/SpecialGlobalVars:
...
-  Enabled: false
+Style/HashEachMethods:
+  Enabled: true
+
+Style/HashTransformKeys:
+  Enabled: true
+
+Style/HashTransformValues:
+  Enabled: true
+
```
mikker: Apparently new configs added that need to be explicitly enabled/disabled. Weird Rubocop.

spec/integration/graphql_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+    include Rack::Test::Methods
+
+    def setup_database
+      ActiveRecord::Base.logger = Logger.new(SpecLogger)
+      ActiveRecord::Base.logger = Logger.new(nil)
+
+      ActiveRecord::Base.establish_connection(
+        adapter: 'sqlite3',
+        database: '/tmp/graphql.sqlite3'
+      )
```
mikker: @estolfo perhaps we ought to add something like this to the generic Rails spec too?

spec/support/with_env.rb
```diff
@@ -6,7 +6,7 @@ def with_env(env)
       current[key] = ENV.key?(key) ? ENV[value] : :__missing
     end
 
-    env.keys.each { |key| ENV[key] = env[key] }
+    env.each_key { |key| ENV[key] = env[key] }
```
mikker: Rubocop complained.

lib/elastic_apm/graphql.rb
```diff
@@ -0,0 +1,71 @@
...
+
+module ElasticAPM
+  # @api private
+  module GraphQL
+    TYPE = 'app'
+    SUBTYPE = 'graphql'
+
+    PREFIX = 'GraphQL: '
+    UNNAMED = '[unnamed]'
+    MORE_THAN_FIVE = '[more-than-five-queries]'
```
estolfo: It seems a bit strange to hardcode the query limit in the transaction name. I looked at the spec in the apm repo and the limit 5 seems to be arbitrary. What if that number changes? Would it be worth changing this to "exceeds-max-number-of-queries" in case that is configurable at some point? Or is it unlikely that number will change?

lib/elastic_apm/graphql.rb
```diff
@@ -0,0 +1,71 @@
...
+      'parse' => 'graphql.parse',
+      'validate' => 'graphql.validate',
+      'analyze_multiplex' => 'graphql.analyze_multiplex',
+      'analyze_query' => 'graphql.analyze_query',
+      'execute_multiplex' => 'graphql.execute_multiplex',
+      'execute_query' => 'graphql.execute_query',
+      'execute_query_lazy' => 'graphql.execute_query_lazy',
+      # "execute_field" => "graphql.execute_field",
+      # "execute_field_lazy" => "graphql.execute_field_lazy",
+      # "authorized" => "graphql.authorized",
```
estolfo: Why are these commented out?

spec/integration/graphql_spec.rb
estolfo: The `setup_database` method?

mikker: Sure, yeah. Was mostly thinking about just having it include calling out to AR too. We're integration testing that in our own suite right now.

lib/elastic_apm/graphql.rb
```diff
@@ -0,0 +1,71 @@
...
+        ) { yield }
+
+      if key == 'execute_multiplex'
+        transaction.name = "#{PREFIX}#{concat_names(results)}"
+      end
+
+      results
+    end
+
+    class << self
```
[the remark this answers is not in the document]
mikker: Uhm, I only had to do the `class << self` dance because I wanted `private`. I suppose there are other ways. This is on purpose, though :)

lib/elastic_apm/graphql.rb
mikker: That's a good point 👍

lib/elastic_apm/graphql.rb
mikker: They are available but left out deliberately. I thought it would be nice to include them in this form to 1) show what we skip and 2) be easy to find if we want to stop skipping them.

estolfo: Are all of these names common across all GraphQL implementations in different languages? Maybe it's worth noting in the spec which ones we include and exclude just so there's some documentation somewhere.

mikker: I don't think they are necessarily the same, no. There's already a note (sort of) about this: https://github.com/elastic/apm/pull/208/files#diff-9c3af4b12c4c30fd39c2bcf34e7bc375R146

estolfo: Ok, so perhaps the spec isn't the right place to document which are included and excluded. But I still don't know myself why they are commented out - would you mind explaining here or putting a comment in the source?

lib/elastic_apm/graphql.rb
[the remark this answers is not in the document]
estolfo: I was asking because you could just put the `class << self` on line 30 and then be able to write `def trace`

spec/integration/graphql_spec.rb
estolfo: Sure, that works.

lib/elastic_apm/graphql.rb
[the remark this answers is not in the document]
mikker: Yes, I know. I somehow like this better. I can change it if you'd like?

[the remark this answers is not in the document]
estolfo: No problem, was just trying to be _extra efficient_ and save 5 characters ;)

lib/elastic_apm/graphql.rb
```diff
@@ -0,0 +1,73 @@
...
+      private
+
+      def query_name(query)
+        query.operation_name || UNNAMED
+      end
+
+      def concat_names(results)
+        if results.length > 5
```
estolfo: I think this number (5) should be a constant

lib/elastic_apm/graphql.rb
```diff
@@ -0,0 +1,74 @@
...
+  # @api private
+  module GraphQL
+    TYPE = 'app'
+    SUBTYPE = 'graphql'
+
+    PREFIX = 'GraphQL: '
+    UNNAMED = '[unnamed]'
+    MULTIPLE_QUERIES = '[multiple-queries]'
+    CONCATENATOR = '+'
+    MAX_NUMBER_OF_QUERIES_FOR_NAME = 5
```
estolfo: yaaas 😎

PR #722, opened 2020-02-24
Destination address port

### LINKED ISSUE #720: Add span.context.destination.address and span.context.destination.port
When available, add `address` and `port` to `span.context.destination`

Related: #613

estolfo:
This PR adds `span.context.destination.address` and `span.context.destination.port` when they are available. The spies for which it was possible to get this info were:
- `Faraday`
- `NetHTTP`
- `HTTP`

The following don't expose this info in the spied-on method:
- `Mongo`
- `Sequel`
- `ActiveRecord`
- `Elasticsearch`

I've also updated the `SpanSerializer` to add these new fields.

**Note there is no test for IPv6 for Faraday because the `get` method fails when an IPv6 uri is provided. See [this](https://github.com/lostisland/faraday/issues/589) issue.

Closes #720

### DIFF
--- a/lib/elastic_apm/grpc.rb
+++ b/lib/elastic_apm/grpc.rb
@@ -5,7 +5,8 @@ module ElasticAPM
   class GRPC
     # @api private
     class ClientInterceptor < ::GRPC::ClientInterceptor
-      TYPE = 'external.grpc'
+      TYPE = 'external'
+      SUBTYPE = 'grpc'
 
       # rubocop:disable Lint/UnusedMethodArgument
       def request_response(request:, call:, method:, metadata:)
@@ -13,7 +14,23 @@ module ElasticAPM
         if (trace_context = transaction.trace_context)
           trace_context.apply_headers { |k, v| metadata[k.downcase] = v }
         end
-        ElasticAPM.with_span(method, TYPE) do
+
+        peer = call.instance_variable_get(:@wrapped)&.peer
+        context =
+          if peer
+            destination = ElasticAPM::Span::Context::Destination.new(
+              type: TYPE,
+              name: SUBTYPE,
+              resource: peer
+            )
+            ElasticAPM::Span::Context.new(destination: destination)
+          end
+
+        ElasticAPM.with_span(
+          method, TYPE,
+          subtype: SUBTYPE,
+          context: context
+        ) do
           yield
         end
       end
--- a/lib/elastic_apm/grpc.rb
+++ b/lib/elastic_apm/grpc.rb
@@ -18,10 +18,13 @@ module ElasticAPM
         peer = call.instance_variable_get(:@wrapped)&.peer
         context =
           if peer
+            split_peer = URI.split(peer)
             destination = ElasticAPM::Span::Context::Destination.new(
               type: TYPE,
               name: SUBTYPE,
-              resource: peer
+              resource: peer,
+              address: split_peer[0],
+              port: split_peer[6]
             )
             ElasticAPM::Span::Context.new(destination: destination)
           end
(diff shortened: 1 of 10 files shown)


lib/elastic_apm/span/context/destination.rb
```diff
@@ -5,21 +5,37 @@ class Span
...
+        attr_reader(
+          :name,
+          :resource,
+          :type,
+          :address,
+          :port
+        )
 
-        def self.from_uri(uri_or_str, type: 'external')
+        def self.from_uri(uri_or_str, type: 'external', port: nil)
```
estolfo: I added this argument, in case we are able to pass in a default port at some point.

lib/elastic_apm/transport/serializers/span_serializer.rb
```diff
@@ -78,7 +78,10 @@ def build_destination(destination)
                 resource: keyword_field(destination.resource),
                 type: keyword_field(destination.type)
               }
-            }
+            }.tap do |dest|
+              dest[:address] = keyword_field(destination.address)
+              dest[:port] = keyword_field(destination.port.to_s)
```
estolfo: Is this right? It should be a string?

mikker: The JSON spec knows 😅

mikker:
This is good!

It's deliberate, btw, that ports are only added to the `resource` for HTTP libs as the design doc says to only add them to those (for now). @ron will know more, I seem to recall that the reason is that it's harder than perhaps first assumed to tell individual db hosts apart, so we put them in the same bucket on service maps.

lib/elastic_apm/transport/serializers/span_serializer.rb
estolfo: it's an integer so I've updated it

spec/elastic_apm/grpc_spec.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 if !defined?(JRUBY_VERSION) && RUBY_VERSION < '2.7'
-require 'grpc'
+  require 'grpc'
```
mikker:
```suggestion
  require 'grpc'
  
```

lib/elastic_apm/grpc.rb
```diff
@@ -5,15 +5,35 @@ module ElasticAPM
...
+            split_peer = URI.split(peer)
+            destination = ElasticAPM::Span::Context::Destination.new(
+              type: TYPE,
+              name: SUBTYPE,
+              resource: peer,
+              address: split_peer[0],
+              port: split_peer[6]
+            )
+            ElasticAPM::Span::Context.new(destination: destination)
+          end
```
mikker: I think it would be nice to extract this into its own method. This one's getting a bit long and hard to read.

lib/elastic_apm/transport/serializers/span_serializer.rb
```diff
@@ -78,7 +78,10 @@ def build_destination(destination)
                 resource: keyword_field(destination.resource),
                 type: keyword_field(destination.type)
               }
-            }
+            }.tap do |dest|
+              dest[:address] = keyword_field(destination.address)
+              dest[:port] = destination.port
+            end
```
mikker: Why not just use regular `key: value` here instead of the `tap`?

spec/elastic_apm/spies/http_spec.rb
```diff
@@ -53,6 +53,27 @@ module ElasticAPM
       expect(destination.name).to match('http://example.com')
       expect(destination.resource).to match('example.com:80')
       expect(destination.type).to match('external')
+      expect(destination.address).to match('example.com')
+      expect(destination.port).to match(80)
+    end
+
+    it 'adds destination information with IPv6' do
```
mikker: 👍

lib/elastic_apm/transport/serializers/span_serializer.rb
estolfo: because that would make too much sense!

PR #726, opened 2020-02-28
Handle case when Bundler is defined but there's no Gemfile

estolfo:
I noticed this minor issue when testing Resque with the agent. Bundler was defined but my resque example directory didn't contain a Gemfile. I thus got this error when `Bundler.bundle_path` was called.

`Bundler::GemfileNotFound: Could not locate Gemfile`

This change checks if `Bundler.default_bundle_dir` is defined before calling `Bundler.bundle_path`.

### DIFF
--- a/lib/elastic_apm/stacktrace_builder.rb
+++ b/lib/elastic_apm/stacktrace_builder.rb
@@ -12,8 +12,12 @@ module ElasticAPM
     RUBY_VERS_REGEX = %r{ruby(/gems)?[-/](\d+\.)+\d}.freeze
     JRUBY_ORG_REGEX = %r{org/jruby}.freeze
 
-    GEMS_PATH = (defined?(Bundler) && Bundler.default_bundle_dir) ?
-                  Bundler.bundle_path.to_s : Gem.dir
+    GEMS_PATH =
+      if defined?(Bundler) && Bundler.default_bundle_dir
+        Bundler.bundle_path.to_s
+      else
+        Gem.dir
+      end
 
     def initialize(config)
       @config = config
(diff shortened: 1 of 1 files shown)


lib/elastic_apm/stacktrace_builder.rb
```diff
@@ -12,7 +12,8 @@ class StacktraceBuilder
     RUBY_VERS_REGEX = %r{ruby(/gems)?[-/](\d+\.)+\d}.freeze
     JRUBY_ORG_REGEX = %r{org/jruby}.freeze
 
-    GEMS_PATH = defined?(Bundler) ? Bundler.bundle_path.to_s : Gem.dir
+    GEMS_PATH = (defined?(Bundler) && Bundler.default_bundle_dir) ?
+                  Bundler.bundle_path.to_s : Gem.dir
```
mikker: Suggestion for readability

```suggestion
    GEMS_PATH =
      if defined?(Bundler) && Bundler.default_bundle_dir
        Bundler.bundle_path.to_s
      else
        Gem.dir
      end
```

estolfo:
I can imagine someone not using a Gemfile if they are testing (as I was) but yes, I agree that it's not the most common use case.

lib/elastic_apm/stacktrace_builder.rb
mikker: I think you force pushed over this?

estolfo: No, not on purpose. I rebased on master and forced pushed, not realizing that the committed suggestion hadn't been pulled down locally.

mikker: That was my guess. Classic 😊

PR #729, opened 2020-03-03
Add Resque instrumentation

### LINKED ISSUE #75: Add Resque instrumentation and error reporting

estolfo:
Closes #75

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -44,6 +44,7 @@ endif::[]
 
 - Support for gRPC using the `grpc` gem (Experimental) {pull}669[#669]
 - Add `span.context.destination.address` and `span.context.destination.port` when available. {pull}722[#722]
+- Add support for Resque. {pull}729[#729]
 
 [float]
 ===== Changed
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -43,6 +43,18 @@ Stop the currently running agent. Use this inside `at_exit` in your
 <<getting-started-rack,Rack app>> to gracefully shut down.
 
 [float]
+[[api-agent-restart]]
+==== `ElasticAPM.restart`
+
+If the agent is already running, this method will stop and start the agent.
+
+If the agent is not already running, this method will start the agent.
+
+A config can be passed to the method that will be used to start the agent. If the agent
+is already running and no config is passed to the `#restart` method, the running agent's
+config will be used.
+
+[float]
 [[api-agent-running]]
 ==== `ElasticAPM.running?`
 
--- a/docs/supported-technologies.asciidoc
+++ b/docs/supported-technologies.asciidoc
@@ -86,6 +86,31 @@ We automatically instrument background processing using:
 - Sidekiq
 - Shoryuken
 - Sneakers (v2.12.0+) (Experimental, see https://github.com/elastic/apm-agent-ruby/pull/676[#676])
+- Resque (v2.0.0+)
+
+[float]
+[[supported-technologies-resque]]
+=== Resque
+
+You should require `elastic_apm/resque` in the Rakefile where the Resque rake tasks are required.
+For example, in your `Rakefile`:
+
+[source,ruby]
+----
+require 'resque'
+require 'elastic_apm'
+require 'elastic_apm/resque'
+----
+
+When you start Resque, you should see a series of messages like the following in the Resque logs:
+
+[source,ruby]
+----
+I, [XXX #81227]  INFO -- : Starting worker main
+D, [XXX #81227] DEBUG -- : Registered signals
+I, [XXX #81227]  INFO -- : Running before_first_fork hooks
+D, [XXX #81227] DEBUG -- : Starting ElasticAPM agent
+----
 
 [float]
 [[supported-technologies-grpc]]
--- a/docs/supported-technologies.asciidoc
+++ b/docs/supported-technologies.asciidoc
@@ -92,8 +92,9 @@ We automatically instrument background processing using:
 [[supported-technologies-resque]]
 === Resque
 
-You should require `elastic_apm/resque` in the Rakefile where the Resque rake tasks are required.
-For example, in your `Rakefile`:
+To make the agent work with Resque, you need to require `elastic_apm/resque` before you boot your Resque worker process.
+
+For example in your `Rakefile`:
 
 [source,ruby]
 ----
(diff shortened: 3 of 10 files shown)


CHANGELOG.asciidoc
```diff
@@ -44,6 +44,7 @@ endif::[]
 
 - Support for gRPC using the `grpc` gem (Experimental) {pull}669[#669]
 - Add `span.context.destination.address` and `span.context.destination.port` when available. {pull}722[#722]
+- Add support for instrumenting `Resque`. {pull}729[#729]
```
mikker:
```suggestion
- Add support for Resque. {pull}729[#729]
```

Not really _code_.

docs/api.asciidoc
```diff
@@ -42,6 +42,16 @@ If you're not using Rails, see <<getting-started-rack,Getting started with Rack>
 Stop the currently running agent. Use this inside `at_exit` in your
 <<getting-started-rack,Rack app>> to gracefully shut down.
 
+[float]
+[[api-agent-restart]]
+==== `ElasticAPM.restart`
+
+If the agent is already running, this method will stop and start the agent.
```
mikker:
```suggestion
If the agent is already running, this method will stop and start the agent.

```

docs/api.asciidoc
```diff
@@ -42,6 +42,16 @@ If you're not using Rails, see <<getting-started-rack,Getting started with Rack>
...
+If the agent is not already running, this method will start the agent.
```
mikker:
```suggestion
If the agent is not already running, this method will start the agent.

```

docs/supported-technologies.asciidoc
```diff
@@ -86,6 +86,33 @@ We automatically instrument background processing using:
...
 - Sneakers (v2.12.0+) (Experimental, see https://github.com/elastic/apm-agent-ruby/pull/676[#676])
+- Resque (v2.0.0+)
+
+[float]
+[[supported-technologies-resque]]
+=== Resque
+
+You should require `elastic_apm/resque` in the Rakefile where the Resque rake tasks are required.
+Doing this will set up a hook to automatically start the ElasticAPM agent after Resque forks for the
+first time.
```
mikker:
```suggestion
```

I don't think we need to explain _why_ here.

docs/supported-technologies.asciidoc
```diff
@@ -86,6 +86,33 @@ We automatically instrument background processing using:
...
+For example, in your `Rakefile`:
+
+[source,ruby]
+----
+require 'resque'
+require 'elastic_apm'
+require 'elastic_apm/resque'
+----
+
+When you start Resque, you should see a series of messages like the following in the Resque logs:
```
mikker: 👌

lib/elastic_apm.rb
```diff
@@ -46,6 +46,16 @@ def stop
       Agent.stop
     end
 
+    # Restarts the ElasticAPM Agent using the same config or a new
+    # one, if it is provided.
+    # Starts the agent if it is not running.
+    # Stop ands starts the agent if it is running.
```
mikker:
```suggestion
    # Stops and starts the agent if it is running.
```

spec/elastic_apm/resque_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'resque'
+
+RSpec.describe Resque do
+  it 'defines a before_first_fork hook' do
+    hooks_count = ::Resque.instance_variable_get(:@hooks).size
+    require 'elastic_apm/resque'
+    expect(::Resque.instance_variable_get(:@hooks).size).to eq(hooks_count + 1)
+  end
```
mikker: Not sure I see the need for this spec.

spec/elastic_apm_spec.rb
```diff
@@ -13,6 +13,22 @@
       ElasticAPM.stop
       expect(ElasticAPM::Agent).to_not be_running
     end
+
+    it 'restarts the agent', :mock_intake do
```
mikker:
```ruby
describe ".restart" do
  it "starts if stopped" # ...
  it "restarts if running"
end
```

Might provide more helpful error messages.

docs/supported-technologies.asciidoc
estolfo: I think we should tell users that the agent will automatically start upon at the first fork though.

spec/elastic_apm/resque_spec.rb
estolfo: I just wasn't comfortable not testing the ElasticAPM::Resque module at all (or only indirectly in the resque spy spec). Do you think we should just remove this spec entirely?

docs/supported-technologies.asciidoc
mikker: We're already saying _Do this and Resque works_. I don't feel like we need to explain _why_ it does or _how_ it does so. It's all there in the code if anyone's curious.

estolfo: Also, they need to know that the agent will not automatically start if they don't include `elastic_apm/resque` in the `Rakefile`.

mikker: The instructions are needed, not disagreeing.

```
To make the agent work with Resque, you need to require `elastic_apm/resque` before you boot your Resque worker process.

For example in your `Rakefile`:

# ... blah blah
```

I think something like that is sufficient. People don't notice the details anyway, so better get straight to the code example 😅

estolfo: ok👍🏼

spec/elastic_apm/resque_spec.rb
mikker: Yes. I think this is better tested at an integration(ish) level.

estolfo:
Forgot about error handling 😬

lib/elastic_apm/spies/resque.rb
```diff
@@ -0,0 +1,43 @@
...
+          ElasticAPM.restart
+        end
+      end
+
+      def install_perform_spy
+        ::Resque::Job.class_eval do
+          alias :perform_without_elastic_apm :perform
+
+          def perform
+            name = @payload && @payload['class'].name
```
estolfo: Figured out a way to get the Job class name to add to the transaction.

mikker: Wonderful!

Issue #714, opened 2020-02-14
[apm-ci] Release automation failed when syncing up 3.x branch


v1v:
rake release caused 

```
12:33:15  + rake release
12:33:15  elastic-apm 3.5.0 built to pkg/elastic-apm-3.5.0.gem.
12:33:15  Tag v3.5.0 has already been created.
12:33:17  Pushed elastic-apm 3.5.0 to rubygems.org
12:33:17  Switched to a new branch '3.x'
12:33:17  .git/rebase-apply/patch:4191: trailing whitespace.
12:33:17    
12:33:17  .git/rebase-apply/patch:4343: trailing whitespace.
12:33:17  Please view the {apm-overview-ref-v}/index.html[APM Overview] for details on how these components work together. 
12:33:17  .git/rebase-apply/patch:4125: new blank line at EOF.
12:33:17  +
12:33:17  .git/rebase-apply/patch:4302: new blank line at EOF.
12:33:17  +
12:33:17  .git/rebase-apply/patch:20279: new blank line at EOF.
12:33:17  +
12:33:17  warning: 5 lines add whitespace errors.
12:33:17  error: Failed to merge in the changes.
12:33:17  hint: Use 'git am --show-current-patch' to see the failed patch
```

See [here](https://apm-ci.elastic.co/job/apm-agent-ruby/job/apm-agent-ruby-mbp/view/tags/job/v3.5.0/1/execution/node/275/log/)

Let's see what's going on with the CI environment.

PR #732, opened 2020-03-04
ci(jenkins): decouple post-release steps in the rake release step

v1v:

- Update docs
- Create rake task to syncup the branches when a release happens
- Notify if the post-release step failed and keep moving forward. 
- Send an email if the syncup failed.

 
## Why is it important?

Avoid fail fast when the syncup between merges failed, this will help to release the new version of the opbeans.

## Related issues

Closes https://github.com/elastic/apm-agent-ruby/issues/714

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,4 +48,5 @@ To release a new version:
     1. Tag the current commit as new version.
     2. Push the tag to GitHub.
     3. Build the gem and upload to Rubygems (local user needs to be signed in and authorized.)
-    4. Update `2.x` branch to be at released commit and push it to GitHub.
+1. Run `rake syncup`. This will...
+    1. Update `3.x` branch to be at released commit and push it to GitHub.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -48,4 +48,5 @@ To release a new version:
     1. Tag the current commit as new version.
     2. Push the tag to GitHub.
     3. Build the gem and upload to Rubygems (local user needs to be signed in and authorized.)
-    4. Update `2.x` branch to be at released commit and push it to GitHub.
+1. Run `rake syncup`. This will...
```
mikker:
```suggestion
1. Run `rake release:update_branch`. This will...
```

Rakefile
```diff
@@ -2,7 +2,10 @@
 
 require 'bundler/gem_tasks'
 
-Rake::Task[:release].enhance do
+desc """Post release action:
+Update `3.x` branch to be at released commit and push it to GitHub.
+"""
+task :syncup => :environment do
```
mikker:
```suggestion
namespace :release do
  task :update_branch do
```

mikker: I don't think `:environment` is needed here or if we even have such a task in the agent?

mikker: (This suggestion also needs an update to the indentation and one more `end` after the task.)

v1v: I actually used my web search skills and found an example for creating a task in rake, and unfortunately the example was with the `:environment`, thanks for the review :)

PR #737, opened 2020-03-16
Compare metricset counts only to number of successful requests

estolfo:
A really small change to the rails metrics stress test to compare the expected metricsets counts to the number of successful requests.
This would avoid test failures like this that are not indicative of an agent issue: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-ruby%2Fapm-agent-ruby-downstream/detail/master/2933/pipeline

I also removed the expectation on the request status and `env` variable, as those don't tell us anything about the agent functionality.


mikker:
Maybe we should add a short, one-line comment about why we do this, lest we forget.

mikker:
I am very sure that I will forget at least 😆

estolfo:
of course, no problem!

PR #740, opened 2020-03-18
Use config from the singleton agent instead of holding a reference

### LINKED ISSUE #725: Ensure that all config options are dynamic
In preparation for more options being available for edit in the UI, ensure that all config options in the agent can be modified dynamically. Also ensure that transactions and spans get a snapshot of the config that they reference during their lifetime.
See https://github.com/elastic/apm/issues/213

estolfo:
This sets up the codebase to allow all/any options to be dynamically changed.

Closes #725

### DIFF
--- a/lib/elastic_apm/error_builder.rb
+++ b/lib/elastic_apm/error_builder.rb
@@ -3,16 +3,13 @@
 module ElasticAPM
   # @api private
   class ErrorBuilder
-    def initialize(agent)
-      @agent = agent
-    end
 
     def build_exception(exception, context: nil, handled: true)
       error = Error.new context: context || Context.new
       error.exception =
         Error::Exception.from_exception(exception, handled: handled)
 
-      Util.reverse_merge!(error.context.labels, @agent.config.default_labels)
+      Util.reverse_merge!(error.context.labels, config.default_labels)
 
       if exception.backtrace
         add_stacktrace error, :exception, exception.backtrace
@@ -40,7 +37,7 @@ module ElasticAPM
 
     def add_stacktrace(error, kind, backtrace)
       stacktrace =
-        @agent.stacktrace_builder.build(backtrace, type: :error)
+          ElasticAPM.agent.stacktrace_builder.build(backtrace, type: :error)
       return unless stacktrace
 
       case kind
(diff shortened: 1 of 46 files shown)


lib/elastic_apm/error_builder.rb
```diff
@@ -40,7 +37,7 @@ def build_log(message, context: nil, backtrace: nil, **attrs)
 
     def add_stacktrace(error, kind, backtrace)
       stacktrace =
-        @agent.stacktrace_builder.build(backtrace, type: :error)
+          ElasticAPM.agent.stacktrace_builder.build(backtrace, type: :error)
```
mikker: Just a thought: Should we do a brief check for the agent not being `nil` at the point of initialising any class that depends on it being present, like this one?

Like,

```ruby
def initialize(...)
  raise AgentRequired unless ElasticAPM.agent
  
  # ...
end
```

As a safety measure for ourselves and, hopefully, catching cases where the agent is missing as early as possible.

estolfo: Yes, I'm thinking I need to be more defensive if there is no agent when the config is accessed in general. For example, having a method on `ElasticAPM` that takes a config option name so that such an error can be raised if there is no agent and thus no config.

```ruby
module ElasticAPM
    def get_config(option)
      raise AgentRequired unless agent
      Agent.config.send(option)
    end
```

instead of just calling `ElasticAPM.config.option_name` everywhere. I would be afraid we'd get a `no method on NilClass` error at some point.

estolfo: What do you think of that?

mikker:
```ruby
module ElasticAPM
    def config
      raise AgentRequired unless agent
      Agent.config
    end
end
```

Would essentially be the same but allow us to use ordinary method chaining like `ElasticAPM.config.option` without the `send`.

estolfo: Yeah, but what is the config is `nil` for some reason?

mikker: The we could also raise.

```
module ElasticAPM
  def config
    raise AgentRequired unless agent
    raise RunTimeError, 'mysteriously missing Config' unless agent.config
    Agent.config
  end
end
```

estolfo:
Replaced by #741

PR #741, opened 2020-03-19
Allow some config options to be dynamic

### LINKED ISSUE #725: Ensure that all config options are dynamic

estolfo:
Another approach than that in #740

Using `ElasticAPM.config` only from objects referring to dynamic options (i.e. remote config options). Other objects will take the config options they need in `#initialize`. 

These are the remote config options in 7.7 and where they are used in the agent:
**`active`**: `Agent.start`
**`api_request_size`**: `Connection`
**`api_request_time`**: `Connection`
**`capture_body`**: `ContextBuilder`
**`capture_headers`**: `ContextBuilder`
**`log_level`**: `Logging`
**`span_frames_min_duration`**: `Subscriber`, `ElasticAPM.start_span`, `Instrumenter#start_span`, `Span#long_enough_for_stacktrace?`
**`transaction_max_spans`**: `Transaction#inc_started_spans!`
**`transaction_sample_rate`**: `Instrumenter#random_sample?`

This describes each class and the rationale behind changing/not changing it:

**`ContextBuilder`**: This object is created in `Agent#initialize` and accepts `config` as an arg. The only config options it needs are dynamic ones: `capture_body` and `capture_headers`.
Changes:
- removed `config` being passed into `ContextBuilder#initialize`
- made references to `config` via `ElasticAPM.config`

**`ErrorBuilder`**: This object is created in `Agent#initialize` and was accepting the agent as an arg. It needs the `config` for the `default_labels`.
Changes:
- pass in `config` and set `@default_labels` using it.
- pass in `stacktrace_builder` and set it to an instance variable instead of holding a `@agent` instance variable

**`Instrumenter`**: This object is created in `Agent#initialize` and accepts `config` as an arg.
Changes:
- Removed `#initialize` arg and reference to `config` because the only need for it is via a `transaction` object

**`Metadata`**: This object is created in `Connection` and `UserAgent` and accepts `config` as an arg
Changes:
- No changes to method signature, the other objects created in `#initialize` need a `config`

**`Set`**: This object is created in many places, as there are other classes inheriting from it. It accepts `config` as an arg to `#initialize`.
Changes:
- Removed instance variable `@config`
- Only needs to know disabled metrics, so get the value from the `config` in `#initialize`

**`Normalizers`**: Leave this as-is. The `ActionView` normalizer needs a value from the `config`, so we keep an instance variable reference to it.
Changes: NONE

**`Subscriber`**: Previously had an instance variable reference to `@agent` so `current_span` and `current_transaction` could be evaluated.
Changes:
- Pass in config, as it is needed to initialize Normalizers in `Subscriber#initialize`
- Remove reference to `@agent` and get `current_span` and `current_transaction` via `ElasticAPM` instead

**`UserAgent`**: No change. `config` must be passed into `User#initialize` so that `Metadata` can be created.

**`ServiceInfo`**: This object needs to extract many values from the config. It also needs to get the `environment` value from the `config`, which is dynamic.
Changes:
- Removed instance variable `@config` and extract all necessary values from the `config` in `ServiceInfo#initialize`
- Define a method `environment` that gets the `config` via `ElasticAPM` and get the `environment` from there so it's dynamic.

**`SystemInfo`**: The only value needed by this object is `hostname`
Changes:
- Removed instance variable `@config` reference
- Pass in `hostname` value explicitly

**`ProcessInfo`**:
Changes:
- Removed instance variable `@config` reference, the object doesn't need `config` at all

**`BreakdownSet`**: `config` is passed to `BreakdownSet#initialize`. No reference is kept, it is used to disable breakdown metrics
Changes: NONE

**`Set`**: This object only needs to know what metrics are disabled from the config.
Changes:
- Removed instance variable `@config` reference and instead save the values of `config.disable_metrics` in `Set#initialize`

**`ActionView`**: This object needs the `@config.__view_paths` value
Changes: NONE

**`Base`**: This object needs to keep a reference to the `@config` so that Workers can be created.
Changes: NONE

**`Connection`**: This class needs `config` to be passed to `Connection#initialize` so that the `metadata` object can be created with it. It doesn't need to keep a `@config` instance variable because other config options needed are dynamic.
Changes:
- Removed `@config` instance variable
- Refer to `ElasticAPM.config` for dynamic values `api_request_time` and `api_request_size`

**`Filters`**: `config` is passed to `Filters#initialize` because it is needed to created `SecretsFilter`
Changes: NONE

**`SecretsFilter`**: `config` is needed for the `custom_key_filters` and `santize_field_names` values.
Changes:
- Removed `@config` instance variable and just extract the needed values in `SecretsFilter#initialize`

**`Headers`**: `config` is needed to instantiate `UserAgent` and so that headers can be merged.
Changes: NONE

**`Serializers`**: No serializers need a value from the `config`
Changes:
- Removed `config` from `#initialize` method signature 

**`Logging`**: The only reference to `config` from this module is to get the `log_level`. 
Changes:
- Reference `config` via `ElasticAPM.config` because `log_level` is a dynamic option.

**`CPUMemSet`**: There was a `attr_reader` for `config`
Changes:
- Remove `attr_reader` for `config`

**`Worker`**: This class itself doesn't need a reference to `config`. It was used for the `Logging` module. The `config` is passed to `Worker#initialize` because `Connection#initialize` needs it.
Changes:
- Removed `@config` instance variable and just pass the `config` to `Connection#initialize` in `Worker#initialize`

**`Http`**: This class needs a reference to `config` so that a client can be created.
Changes: NONE

Closes #725

### DIFF
--- a/lib/elastic_apm.rb
+++ b/lib/elastic_apm.rb
@@ -32,6 +32,8 @@ require 'elastic_apm/grpc' if defined?(::GRPC)
 # ElasticAPM
 module ElasticAPM
   class << self
+    class AgentRequired < RuntimeError; end
+
     ### Life cycle
 
     # Starts the ElasticAPM Agent
@@ -65,6 +67,16 @@ module ElasticAPM
     # @return [Agent] Currently running [Agent] if any
     def agent
       Agent.instance
+    end
+
+    # Returns the config of the running agent.
+    #
+    # @return [Config] The config of the running agent.
+    # Returns nil if the agent is not running.
+    def config
+      raise AgentRequired unless agent
+      raise RuntimeError, 'Config for agent is missing' unless Agent.config
+      Agent.config
     end
 
     ### Metrics
--- a/lib/elastic_apm/agent.rb
+++ b/lib/elastic_apm/agent.rb
@@ -42,7 +42,11 @@ module ElasticAPM
           return
         end
 
-        @instance = new(config).start
+        # It's important that the @instance variable is set before
+        # calling #start. Objects rely on the config being available
+        # via ElasticAPM.agent.config while starting.
+        @instance = new(config)
+        @instance.start
       end
     end
 
@@ -59,9 +63,13 @@ module ElasticAPM
       !!@instance
     end
 
+    def self.config
+      @instance&.config
+    end
+
     def initialize(config)
       @stacktrace_builder = StacktraceBuilder.new(config)
-      @context_builder = ContextBuilder.new(config)
+      @context_builder = ContextBuilder.new
       @error_builder = ErrorBuilder.new(self)
 
       @central_config = CentralConfig.new(config)
--- a/lib/elastic_apm/agent.rb
+++ b/lib/elastic_apm/agent.rb
@@ -76,7 +76,6 @@ module ElasticAPM
       @transport = Transport::Base.new(config)
       @metrics = Metrics.new(config) { |event| enqueue event }
       @instrumenter = Instrumenter.new(
-        config,
         metrics: metrics,
         stacktrace_builder: stacktrace_builder
       ) { |event| enqueue event }
--- a/lib/elastic_apm/agent.rb
+++ b/lib/elastic_apm/agent.rb
@@ -70,7 +70,10 @@ module ElasticAPM
     def initialize(config)
       @stacktrace_builder = StacktraceBuilder.new(config)
       @context_builder = ContextBuilder.new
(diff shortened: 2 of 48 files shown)


estolfo:
This is too complicated 😅
Closing in favor of https://github.com/elastic/apm-agent-ruby/pull/747

PR #747, opened 2020-04-01
Allow Config values to be dynamic

estolfo:
This is another approach to supporting dynamic config options. 
It swaps out the `config`'s `options` object for a new one when options are updated.

Closes https://github.com/elastic/apm-agent-ruby/issues/725


mikker:
Very elegant, pragmatic solution ❤️

Issue #756, opened 2020-04-06
[CI] Coverage report not found


kuisathaverat:
Builds are falling to archive the coverage report with the following errors

https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-ruby%2Fapm-agent-ruby-downstream/detail/PR-755/2/pipeline

```
[2020-04-06T11:16:39.187Z] Coverage report generated for RSpec to /app/coverage/coverage.xml
```

```
[2020-04-06T11:16:42.678Z] Archiving artifacts
[2020-04-06T11:16:42.833Z] 'coverage/matrix_results/' doesn't match anything, but '' does. Perhaps that's what you mean?
No artifacts found that match the file pattern "coverage/matrix_results/". Configuration error?
```

cc @cachedout

cachedout:
This was resolved with a simple rebase. Closing for now. If this becomes a persistent problem, we can re-open but this isn't expected to be an ongoing issue.

PR #757, opened 2020-04-06
ci(jenkins): merge branch in downstream jobs

v1v:
## What

Downstream jobs need to merge their changes with the target branch otherwise the upstream won't be aligned with the downstream jobs.

## Why is it important?

Upstream and downstream jobs don't use the same version of the source code.

## Related issues

Relates to #756

### DIFF
--- a/.ci/jobs/defaults.yml
+++ b/.ci/jobs/defaults.yml
@@ -17,7 +17,7 @@
     days-to-keep: '30'
     concurrent: true
     node: linux
-    periodic-folder-trigger: 1d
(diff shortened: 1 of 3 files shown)


.ci/jobs/defaults.yml
```diff
@@ -17,7 +17,7 @@
     days-to-keep: '30'
     concurrent: true
     node: linux
-    periodic-folder-trigger: 1d
+    periodic-folder-trigger: 1w
```
v1v, line 20: Unrelated to this PR

Issue #623, opened 2019-11-21
Add enabled & recording options


mikker:
See https://github.com/elastic/apm/issues/92

Should probably replace `disabled` and `disable_send` for example.

mikker:
Blocked by the proposed change not being final yet.

PR #760, opened 2020-04-09
docs: add dynamic badge to configuration options

bmorelli25:
## What does this pull request do?

This PR adds the `Dynamic` badge to supported configurations. I defined "supported configurations" as options that are supported by the [Apm app](https://github.com/elastic/kibana/blob/master/x-pack/plugins/apm/common/agent_configuration/setting_definitions/general_settings.ts). 

- [x] `api_request_size`
- [x] `api_request_time`
- [x] `capture_body`
- [x] `capture_headers`
- [x] `log_level`
- [x] `span_frames_min_duration`
- [x] `transaction_max_spans`
- [x] `transaction_sample_rate`

There is a discrepancy between the list above, and the list created by @estolfo in https://github.com/elastic/apm-agent-ruby/pull/741: `recording` doesn't seem to be a valid option yet.

## Documentation preview

http://apm-agent-ruby_760.docs-preview.app.elstc.co/diff

## Screenshot

Screen Shot 2020-04-08 at 4 56 48 PM

## Related issues

For https://github.com/elastic/apm/issues/232
For https://github.com/elastic/observability-dev/issues/639

PR #762, opened 2020-04-09
Add enabled config option replacing active

estolfo:
Deprecate the `active` config option and replace it with `enabled`. 

#623

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -178,7 +178,7 @@ WARNING: Api keys only provide any real security if your APM server uses TLS.
 
 [float]
 [[config-active]]
-==== `active`
+==== `active` deprecated:[3.7.0,See <<enabled>> instead.]
 |============
 | Environment          | `Config` key | Default
 | `ELASTIC_APM_ACTIVE` | `active`     | `true`
@@ -186,6 +186,8 @@ WARNING: Api keys only provide any real security if your APM server uses TLS.
 
 Whether or not to start the agent.
 If `active` is `false`, `ElasticAPM.start` will do nothing and all calls to the root API will return `nil`.
+
+NOTE: This option is being removed. See <<enabled>> instead.
 
 [float]
 [[config-api-buffer-size]]
@@ -407,6 +409,18 @@ Elastic APM automatically instruments select third party libraries.
 Use this option to disable any of these.
 
 Get an array of enabled instrumentations with `ElasticAPM.agent.config.enabled_instrumentations`.
+
+[float]
+[[config-enabled]]
+==== `enabled`
+[options="header"]
+|============
+| Environment          | `Config` key | Default
+| `ELASTIC_APM_ENABLED` | `enabled`     | `true`
+|============
+
+Whether or not to start the agent.
+If `enabled` is `false`, `ElasticAPM.start` will do nothing and all calls to the root API will return `nil`.
 
 [float]
 [[config-environment]]
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -51,6 +51,11 @@ endif::[]
 - Ensure that the log level is updated for the config's logger when value is changed {pull}755[#755]
 - Set config `false` values to `false`, not `nil` {pull}761[#761]
 
+[float]
+===== Added
+
+- Add enabled config option, replacing active {pull}669[#669]
+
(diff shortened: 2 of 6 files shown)


docs/configuration.asciidoc
```diff
@@ -383,6 +385,18 @@ Use this option to disable any of these.
 
 Get an array of enabled instrumentations with `ElasticAPM.agent.config.enabled_instrumentations`.
 
+[float]
+[[config-enabled]]
+==== `enabled`
+[options="header"]
```
bmorelli25, line 391: Adds a `dynamic` badge like we've done for the other settings: https://github.com/elastic/apm-agent-ruby/pull/760

```suggestion
[options="header"]

<<dynamic-configuration, image:./images/dynamic-config.svg[] >>

```

estolfo: No, only `recording` is dynamic.
It says [here](https://github.com/elastic/apm/issues/92#issuecomment-519752096) about the `enabled` configuration:
```
This can only be set locally to the agent (e.g. in code, config file, environment variable), and is interpreted once at agent initialisation time; once disabled, disabled for the process's lifetime.
```

matschaffer:
Will `config.elastic_apm.enabled` be supported as an attr writer?

Was thinking it'd be nice to make it dynamic based on the presence of the token & url env vars.

e.g., https://github.com/Safecast/safecastapi/pull/670#pullrequestreview-407079971

estolfo:
According to the definitions [here](https://github.com/elastic/apm/issues/92#issuecomment-519752096) only the `recording` option is dynamic. Can you use that instead?
If you want to enabled/disable the agent based on the presence of token and url env vars, can't you interpret that before starting the agent and set `ELASTIC_APM_ENABLED`? Why does the `enabled` setting have to be dynamic?

matschaffer:
As mentioned in https://github.com/Safecast/safecastapi/pull/670#issue-413832508 having the gem installed without env vars set (eg., for local dev) led to a lot of log noise.

matschaffer:
I guess we could also move the gem to a production group in the gemfile. Any idea on what’s “usual”? 

For new relic we have it available and the token is configured via env var. if that’s nil the newrelic gem doesn’t log any errors. The we can set the key in deployed envs to get reporting. Would be nice to have a similar flow with elastic apm.

estolfo:
It sounds like what you've proposed-- putting the gem in the `production` group in the Gemfile-- would resolve this. As mention in [this comment](https://github.com/Safecast/safecastapi/pull/670#issuecomment-624963244) it sounds like there's no need to instrument the app in an environment other than production, so that solution seems to make the most sense.

Alternatively, you could set the elastic apm log level to `FATAL` via an env variable if you want to avoid the error messages when developing locally.

estolfo:
Hi @matschaffer I just realized that maybe you were asking to change the `enabled` attribute on the agent `config` object. Were you asking to do something like the following *before* the agent is started:
```ruby
config.enabled = false
```
or were you looking to change the `enabled` value while the agent is running?

matschaffer:
I'm not clear on where agent start happens in the rails lifecycle, but I don't need to switch it dynamically. It'll be either on or off for the life of the process.

I ended up doing https://github.com/Safecast/safecastapi/pull/677 which uses the gemfile in order to avoid using what _appears_ to be a deprecated config setter.

https://github.com/Safecast/safecastapi/pull/677/files#diff-b1fe55db50c712fef0673345e5b9c0d9L13

matschaffer:
Fwiw, the configuration was a little surprising/confusing compared to newrelic.

With newrelic all the config is always in place (I think it even provides a dev view though I don't use it much). But data is only sent if `NEW_RELIC_LICENSE_KEY` is exported to the env (https://github.com/Safecast/safecastapi/blob/master/config/newrelic.yml#L15)

Was trying to figure out how to do something like that with elastic APM (only sending data if `ELASTIC_APM_SECRET_TOKEN` is set, assuming `ELASTIC_APM_SERVER_URL` will also be set.)

estolfo:
yes, `config.elastic_apm.enabled` would work in the next release :)

Thanks for the feedback regarding comparison between new relic and elastic APM. I do think, however, the best and simplest solution is what you've gone with-- using the groups in the Gemfile as there is no point in loading the gem if it is not intended to be used.

matschaffer:
Agreed. Thanks for helping get me pointed in a good direction.

PR #770, opened 2020-04-21
ci(jenkins): refactor with withGitRelease step

v1v:

Use specific pipeline step to configure the git workspace for the CI

## Why is it important?

Simplify the scripts in the repos to be used across other projects.

## Checklist
- [x] Validate the https does work rather than the sshagent for this particular use case

### Tests
- https://github.com/elastic/apm-agent-ruby/pull/774 was created based on this changes but only running until the dry-run step.

Worked as expected without pushing things

image

Pushing to GitHub verified in my local jenkins environment

image


v1v:
Test failure unrelated to this particular PR

PR #780, opened 2020-04-28
ci(jenkins): avoid converge if only docs

v1v:

Fix a corner case with PRs that only affect docs

## Why is it important?

To avoid failing the build 

## Related issues

See https://github.com/elastic/apm-agent-ruby/pull/778

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -273,6 +273,10 @@ pipeline {
       }
     }
     stage('Coverage converge') {
+      when {
+        beforeAgent true
+        expression { return env.ONLY_DOCS == "false" }
+      }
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -273,6 +273,10 @@ pipeline {
       }
     }
     stage('Coverage converge') {
+      when {
+        beforeAgent true
+        expression { return env.ONLY_DOCS == "false" }
```
v1v, line 278: I was initially tempted to use `expression { return rubyDownstreamJobs.size() > 0 }` but I guess we don't want to as if it's empty then something really bad has happened with the pipeline

Issue #724, opened 2020-02-25
Investigate support for web servers with master/worker architecture


estolfo:
Web servers such as `puma` in "clustered mode" and `unicorn` fork workers from the main (master) process. This leads to undesired behavior, for example: https://github.com/elastic/apm-agent-ruby/issues/696

We should come up with a way to detect forking and adapt.

mikker:
Could potentially also be the missing link to make Resque integration easier. Saying that with only a hunch 😅

estolfo:
For research: https://github.com/elastic/apm-agent-python/pull/636

vidok:
I also think that this method 
https://github.com/elastic/apm-agent-ruby/blob/master/lib/elastic_apm/agent.rb#L59 doesn't answer the question. It just checks an object, I assume it should be something like 
```Ruby
def running?
  central_config.running? &&\
  metrics.running? &&\ 
  instrumenter.running? &&\
  transport.running?
end
```
It would be check everything that should be up and running.

vidok:
Anyway, we should find a simple way to start threads in child processes because `fork` doesn't copy running threads.

estolfo:
yes, indeed. Thanks for your feedback. I'll ping you when we start working on this!

vidok:
@estolfo I would be happy to help. I already started working on it :). I'm also okay if you implement this. 

Thank you.

estolfo:
We always welcome pull requests! Please do open one if you find a solution :)

mikker:
> Anyway, we should find a simple way to start threads in child processes because fork doesn't copy running threads.

Definitely. Restarting the whole agent is also not very expensive and perhaps that could simplify things.

phlegx:
Hi Guys! Is someone already working on this? Will it be in 7.7?

estolfo:
@phlegx Yes, we are working on this and hope to have it in the next release. In the meantime, can you tell us more about your setup? What web server are you using and with what configuration?

Depending on the web server you are using, you might be able to resolve any issues by using the `ElasticAPM.restart` method in an after fork hook. [That is what we do for `Resque`](https://github.com/elastic/apm-agent-ruby/blob/460c14201bbaf71500897449d88d56dfabb3b239/lib/elastic_apm/spies/resque.rb#L16-L18), for example.

phlegx:
Hi @estolfo! Thanks a lot for the info. 

[...]

PR #783, opened 2020-04-30
Detect forking and restart all tasks/background threads

estolfo:
Todo:

- [x] Write tests
- [x] Call `detect_forking!` in each method that creates an event
- [x] Document changes in PR description
- [x] Update Resque implementation
- [x] Show findings from Resque tests and tests with Puma
- [x] Benchmark

[Puma test](https://gist.github.com/estolfo/55f5aae66caff4a2aa81c7cba3543718)
[Transport test](https://gist.github.com/estolfo/3285809388b94b77e83169be9ae80088)

**Transport**: 
- The previous solution for handling forking didn't work because in some cases, the process would be forked, events created, and then the forked process would be terminated before the watcher had a chance to bring up the workers. This was the case with a library like Resque that has potentially short-lived forks.
- We can't just stop and start again because the StopMessage will then be the first message processed when the transport is restarted.
- We also don't need a watcher mutex anymore because methods called on the watcher object won't come from different threads. Previously, `ensure_watcher_running` was called via `#submit`, which could be called from separate threads.

CentralConfig:
- It's "safe" to stop and start it again because the `scheduled_task` will be cancel, then recreated.

Instrumenter:
- There are no background threads, so it's fine to stop and start again.

Metrics:
- As the code comment says, ideally we'd be able to check if the timer task died and restart it but the `Concurrent::TimerTask` API does not have a way to detect if the internal thread has died. So we stop and start `Metrics` again. `@sets` is memoized in `#start` so that it's not recreated when `Metrics` is stopped and started again.

Benchmarks:
```
(master) $ FRAMEWORK=rails-6.0 bench/benchmark.rb
-- create_table(:posts, {:force=>true})
   -> 0.0080s
                      user     system      total        real
with agent:      21.338486   3.983200  25.321686 ( 24.874574)
without agent:   12.976541   2.314433  15.290974 ( 15.626493)

(forking-2) $ FRAMEWORK=rails-6.0 bench/benchmark.rb
-- create_table(:posts, {:force=>true})
   -> 0.0182s
                      user     system      total        real
with agent:      20.871092   3.956547  24.827639 ( 24.374119)
without agent:   11.765376   2.073324  13.838700 ( 13.869019)
```

### DIFF
--- a/lib/elastic_apm/agent.rb
+++ b/lib/elastic_apm/agent.rb
@@ -270,6 +270,9 @@ module ElasticAPM
 
     def detect_forking!
       return if @pid == Process.pid
+
+      config.logger.debug "Detected forking,
+        restarting threads in process [PID:#{Process.pid}]"
       central_config.handle_forking!
       transport.handle_forking!
       instrumenter.handle_forking!
--- a/lib/elastic_apm/agent.rb
+++ b/lib/elastic_apm/agent.rb
@@ -273,7 +273,7 @@ module ElasticAPM
       central_config.handle_forking!
       transport.handle_forking!
       instrumenter.handle_forking!
-      #metrics.handle_forking!
+      metrics.handle_forking!
       @pid = Process.pid
     end
   end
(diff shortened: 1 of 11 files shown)


estolfo:
thanks @simitt !

mikker:
Great work! ❤️

estolfo:
Resolve https://github.com/elastic/apm-agent-ruby/issues/724

PR #784, opened 2020-05-03
Put env var config for rails on the main page

matschaffer:

Updates Rails getting started docs.

## Why is it important?

I finally set up elastic apm for safecast.org.

When doing so I landed on https://www.elastic.co/guide/en/apm/agent/ruby/current/getting-started-rails.html pushes you toward `config/elastic_apm.yml` but I don’t think people really want to check in their token if they can avoid it.

I tried env substitution in the file, but that led to CI errors (https://app.circleci.com/pipelines/github/Safecast/safecastapi/415/workflows/33c23466-cca2-4266-84e6-c4b877c4b0ee/jobs/897).

I thought it would be nice if `ELASTIC_APM_SERVER_URL` and `ELASTIC_APM_SECRET_TOKEN` were front-and-center.

### DIFF
--- a/docs/getting-started-rails.asciidoc
+++ b/docs/getting-started-rails.asciidoc
@@ -25,4 +25,6 @@ server_url: http://localhost:8200
 secret_token: ''
 ----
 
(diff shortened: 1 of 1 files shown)


estolfo:
Thanks @matschaffer! I don't see a reason why we wouldn't want to include this.

PR #786, opened 2020-05-04
ci: aggregate downstream test results

v1v:
##  What does this pull request do?

Aggregate the downstream test results in the upstream pipeline

## Why is it important?

Enable GitHub PR comments from the parenstream pipeline with the test results. Otherwise when using parenstream/downstream pattern the test results are not reported.

## Tests

Tests are reported in the parentstream, for instance

image

image

And therefore those details are also reported in the GitHub PR comment https://github.com/elastic/apm-agent-ruby/pull/786#issuecomment-623651784

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -340,8 +340,7 @@ def runBenchmark(version){
 }
 
 def runJob(String rubyVersion, int sleep = 1){
-  def buildFailed = false
-  def buildObject, ret
+  def buildObject, ret, buildNumber, buildName
   def branch = env.CHANGE_ID?.trim() ? env.CHANGE_TARGET : env.BRANCH_NAME
   def jobName = "${env.DOWNSTREAM_PIPELINE}/${env.BRANCH_NAME}"
   try {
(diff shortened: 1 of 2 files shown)


.ci/Jenkinsfile
```diff
@@ -295,7 +295,9 @@ pipeline {
 
 def unpack_artifacts(builds) {
     builds.values().each() { build ->
-      copyArtifacts(projectName: "apm-agent-ruby/apm-agent-ruby-downstream/${env.BRANCH_NAME}", selector: specific(buildNumber: build.number.toString()))
+      copyArtifacts(projectName: "apm-agent-ruby/apm-agent-ruby-downstream/${env.BRANCH_NAME}", 
+                    filter: 'coverage/**/*.*',
+                    selector: specific(buildNumber: build.number.toString()))
```
v1v, line 300: Let's only care about the coverage ones as that's the aim of this particular section. This will help to avoid duplicating the test results when aggregating the dowsntream test output that happens later on.

.ci/Jenkinsfile
```diff
@@ -337,25 +339,42 @@ def runBenchmark(version){
...
+    buildNumber = buildObject.number.toString()
   } catch(e) {
     rubyTasksFailed = true
+    buildFailed = true
+    buildNumber = buildObject.number.toString()
+    log(level: 'INFO', text: "Job failed '${buildObject?.name}' - ${buildNumber}")
     buildObject = e
-    warnError('Downstream Failed') {
-      error("Downstream job for '${rubyVersion}' failed")
-    }
```
v1v, line 356: It has been moved otherwise the finally section didn't happen for some reason

.ci/downstreamTests.groovy
```diff
@@ -117,9 +117,9 @@ class RubyParallelTaskGenerator extends DefaultParallelTaskGenerator {
           saveResult(x, y, 0)
           steps.error("${label} tests failed : ${e.toString()}\n")
         } finally {
-          steps.junit(allowEmptyResults: true,
-            keepLongStdio: true,
-            testResults: "**/spec/junit-reports/**/ruby-agent-junit.xml")
+          def testResultsPattern = '**/spec/junit-reports/**/ruby-agent-junit.xml'
+          steps.junit(testResults: testResultsPattern, allowEmptyResults: true, keepLongStdio: true)
+          steps.archiveArtifacts(artifacts: testResultsPattern, allowEmptyArchive: true, onlyIfSuccessful: false)
```
v1v, line 122: This is required to be consumed from the parentstream

.ci/Jenkinsfile
```diff
@@ -337,25 +339,42 @@ def runBenchmark(version){
...
   } finally {
+    dir(rubyVersion) {
+      copyArtifacts(projectName: jobName, filter: '**/ruby-agent-junit.xml', selector: specific(buildNumber: buildNumber))
+    }
```
v1v, line 366: Using the rubyVersion to help with the categorisation of the tests, so the same stage will run for different downstreams, so this will enhance the test reporting:

image

.ci/Jenkinsfile
```diff
@@ -337,25 +339,42 @@ def runBenchmark(version){
...
+    // Aggregate test results from downstream jobs.
+    junit(testResults: "${rubyVersion}/**/ruby-agent-junit.xml", allowEmptyResults: true, keepLongStdio: true)
+    archiveArtifacts(artifacts: "${rubyVersion}/**", allowEmptyArchive: true, onlyIfSuccessful: false)
```
v1v, line 369: Only for debugging purposes, let's remove it

```suggestion
```

v1v:
After a further analysis, we won't be using this approach and likely moving to traditional stages rather than downstream/parentstream design as this particular test aggregation is not supported and there are some design flags that could cause some unknowns.

PR #789, opened 2020-05-13
Elasticsearch client integration: Use body in db statement

picandocodigo:
I have been working on an example app integrating this agent and the [Elasticsearch Ruby client](github.com/elastic/elasticsearch-ruby). Looking at the information on the requests on the APM Server, I noticed some more information could be provided.

The apm client uses the elasticsearch client's `perform_request`, and passes its parameters as statements to APM. However some endpoints don't use parameters and use body instead, or use both.

This PR makes it so that the Elasticsearch spy sends both parameters and body to the APM server on each request. So if there's only a parameter, the information will remain the same. But if there's a body, json will be passed in with a `params` and a `body` key with the corresponding information. This could be improved to pass in just the body if there's only a body and not use these keys, more tests for when having both, etc. But before working further on this code, I wanted to check in if this is even relevant or useful for the agent. Not including the body could be a decision made by design, since it might be too much information, although it does seem like the server trims the information to how much it can show.

Here's a screenshot of how the body would look in a bulk example:
image

Let me know what you think, thanks!

## Why is it important?

It could be useful to see the request bodies in the APM Server, but it could also be too much information depending on what kind of request it is (e.g. a bulk request with many records).

## Checklist

- [x] New and existing [**unit** tests](https://github.com/elastic/apm-agent-ruby/blob/master/CONTRIBUTING.md#testing) pass locally with my changes

### DIFF
--- a/lib/elastic_apm/spies/elasticsearch.rb
+++ b/lib/elastic_apm/spies/elasticsearch.rb
@@ -26,6 +26,10 @@ module ElasticAPM
       TYPE = 'db'
       SUBTYPE = 'elasticsearch'
 
+      def self.sanitizer
+        @sanitizer ||= ElasticAPM::Transport::Filters::HashSanitizer.new
+      end
+
       def install
         ::Elasticsearch::Transport::Client.class_eval do
           alias perform_request_without_apm perform_request
@@ -38,7 +42,9 @@ module ElasticAPM
 
             if ElasticAPM.agent.config.capture_elasticsearch_queries
               unless args[1].nil? || args[1].empty?
-                statement << { body: args[1] }
+                statement << {
+                  body: ElasticAPM::Spies::ElasticsearchSpy.sanitizer.strip_from!(args[1])
+                }
               end
             end
 
--- a/lib/elastic_apm/spies/elasticsearch.rb
+++ b/lib/elastic_apm/spies/elasticsearch.rb
@@ -34,12 +34,12 @@ module ElasticAPM
             name = format(NAME_FORMAT, method, path)
             statement = []
 
-            unless args[0].nil?
-              statement << { params: args[0] }
-            end
+            statement << { params: args&.[](0) }
 
-            unless args[1].nil? || args[1].empty?
-              statement << { body: args[1] }
+            if ElasticAPM.agent.config.capture_elasticsearch_queries
+              unless args[1].nil? || args[1].empty?
+                statement << { body: args[1] }
+              end
             end
 
             context = Span::Context.new(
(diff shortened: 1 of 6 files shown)


Mpdreamz:
Awesome I am also including this information. 

https://github.com/elastic/apm-agent-dotnet/pull/329#issuecomment-616718098

The .net apm client integration also has spans for sending/receiving, I want to offload request/response to those and include information about the state of the client on the main span. (dead node count, open memory buffer count). The latter are great to send as metrics to the APM server too.

lib/elastic_apm/spies/elasticsearch.rb
```diff
@@ -32,10 +32,21 @@ def install
...
+
+            statement << { params: args&.[](0) }
+
+            if ElasticAPM.agent.config.capture_elasticsearch_queries
+              @sanitizer = ElasticAPM::Transport::Filters::HashSanitizer.new
+              unless args[1].nil? || args[1].empty?
+                statement << {
+                  body: @sanitizer.strip_from!(args[1])
+                }
+              end
```
mikker, line 45: We can't set instance vars inside patched objects – worst case we override and break something 😅

```suggestion
              unless args[1].nil? || args[1].empty?
                body = 
                  ElasticAPM::Transport::Filters::HashSanitizer.new
                  .strip_from!(args[1])
                statement << { body: body }
              end
```

We should probably save the sanitizer in `ElasticAPM::Spies::Elasticsearch` instead so as to not allocate a new one for each and every ES query.

mikker: https://github.com/mikker/apm-agent-ruby/blob/master/lib/elastic_apm/spies/sequel.rb#L12-L14

picandocodigo: @mikker right! I followed the sequel pattern and saved the sanitizer in the Spy now, much better :smile:

lib/elastic_apm/spies/elasticsearch.rb
```diff
@@ -26,16 +26,31 @@ class ElasticsearchSpy
...
+
+            statement << { params: args&.[](0) }
+
+            if ElasticAPM.agent.config.capture_elasticsearch_queries
+              sanitizer = ElasticAPM::Spies::ElasticsearchSpy.sanitizer
+              unless args[1].nil? || args[1].empty?
+                statement << {
+                  body: sanitizer.strip_from!(args[1])
+                }
+              end
```
mikker, line 49:
```suggestion
              unless args[1].nil? || args[1].empty?
                statement << {
                  body: ElasticAPM::Spies::ElasticsearchSpy.sanitizer.strip_from!(args[1])
                }
              end
```

mikker: Need to use it too 😉

mikker:
Good work! Thanks

picandocodigo:
@mikker thanks for the review! :tada:

Issue #754, opened 2020-04-06
Incompatibilities with the OpenTracing interface


eric-christian:
Hello,

We are currently in the process of moving from jaeger to APM and encountered some problems with our implementation of the OpenTracing interface and the APM opentracing-bridge.

Before i continue with the points we found, i would like to ask a question: What is the current state of the OpenTracing bridge feature? Is it still maintained? Is it worse to continue investing time? :)

So far we found the following points:
1. Added FORMAT_TEXT_MAP for injection/extraction
2. `start_active_span { /code block/ }`, when a code block is given, returns the result of the code block, not the scope
3. The methods `start_span` and `start_active_span` have to take the parameter `tags`, not `labels`
4. `span.set_tag` must be implemented and has to return self
5. `span.finish` - the parameter `end_time` is not deprecated. Also we had issues with the default `clock_end: Util.monotonic_micros`. When a span is started, the default start_time is set as `start_time = Util.micros(Time.now)`. Combining `Util.micros(Time.now)` with `Util.monotonic_micros` resulted in durations that looked like this big numbers `1586156283628015`. We didn't go into more details here and simply stopped mixing them together.
6. The parameter `child_of` is supposed to be the parent span or parent span_context of the newly created span. The extracted span_context from ElasticAPM is already changed to represent a child of the once injected context.
7. `tracer.extract` has to return a `OpenTracing::SpanContext`, not an `ElasticAPM::TraceContext`.

We were able to fix all of these points, at least to our understanding.
Here is our fork https://github.com/eric-christian/apm-agent-ruby/commits/master
Is a PR welcome?

What we didn't check is the handling of any of the OpenTracing and ElasticAPM specific tags or log keys. We'll do this at a later point in time.

## Expected behavior
We expected to be able to simply switch from one tracer implementation to another. That is the reason why we started to implement the interface of OpenTracing and not the interface of a specific tracer.

## Environment

- **OS**: OS X
- **Ruby version**: 2.6.3
- **APM Server version**: 7.6.2
- **Agent version**: 3.6.0
- **OpenTracing version**: 0.5.0

Best regards,
Eric

estolfo:
Hi @eric-christian! Thanks for opening this issue and for the detailed description. We'll investigate your points, but in the meantime, we certainly do welcome PRs! Feel free to open one with the work you've done.
Thanks and we'll get back to you soon.

estolfo:
[...]

PR #791, opened 2020-05-13
Update Opentracing bridge

estolfo:
Resolves https://github.com/elastic/apm-agent-ruby/issues/754

As pointed out in the issue, there are a number of incompatibilities with the OpenTracing interface. They are addressed in this PR.

Thanks for @eric-christian for both pointing out these issues and his work to resolve them.

- [x] Added FORMAT_TEXT_MAP for injection/extraction
- [x] `start_active_span` { /code block/ }, when a code block is given, returns the result of the code block, not the scope
- [x] The methods `start_span` and `start_active_span` have to take the parameter `tags`, not labels. **There's no need to support both `labels` and `tags` on this method, as it was an API bug to have removed `tags` and replaced it with `labels`.**
- [x] `span.set_tag` must be implemented and has to return self. **I've removed the `set_label` method, as it should never have replaced `set_tag`.**
- [x] `span.finish` - the parameter end_time is not deprecated. Also we had issues with the default clock_end: Util.monotonic_micros. When a span is started, the default start_time is set as start_time = Util.micros(Time.now). Combining Util.micros(Time.now) with Util.monotonic_micros resulted in durations that looked like this big numbers 1586156283628015. We didn't go into more details here and simply stopped mixing them together. **`clock_end` is removed because it was an API bug to deprecate `end_time` in favor of it. The OpenTracing::Span#finish API uses a `Time` object as the `end_time` value so we should do the conversion ourselves. Also, `start_span` requires `start_time` to be a `Time` and we do the conversion, so the same should be done in `#finish`**
- [x] The parameter `child_of` is supposed to be the parent span or parent span_context of the newly created span. The extracted span_context from ElasticAPM is already changed to represent a child of the once injected context.
- [x] `tracer.extract` has to return a `OpenTracing::SpanContext`, not an `ElasticAPM::TraceContext`.

### DIFF
--- a/lib/elastic_apm/opentracing.rb
+++ b/lib/elastic_apm/opentracing.rb
@@ -85,12 +85,7 @@ module ElasticAPM
       def finish(clock_end: Util.monotonic_micros, end_time: nil)
         return unless (agent = ElasticAPM.agent)
 
-        if end_time
-          warn '[ElasticAPM] DEPRECATED: Setting a custom end time as a ' \
-            '`Time` is deprecated. Use `clock_end:` and monotonic time instead.'
-          clock_end = end_time
-        end
-
+        clock_end = Util.micros(end_time) if end_time
         elastic_span.done clock_end: clock_end
 
         case elastic_span
--- a/lib/elastic_apm/opentracing.rb
+++ b/lib/elastic_apm/opentracing.rb
@@ -116,6 +116,8 @@ module ElasticAPM
 
     # @api private
     class SpanContext
+      extend Forwardable
+
       def initialize(trace_context:, baggage: nil)
         if baggage
           ElasticAPM.agent.config.logger.warn(
@@ -127,6 +129,7 @@ module ElasticAPM
       end
 
       attr_accessor :trace_context
+      def_delegators :trace_context, :trace_id, :id, :parent_id
 
       def self.from_header(header)
         return unless header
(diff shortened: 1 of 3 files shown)


lib/elastic_apm/opentracing.rb
```diff
@@ -293,7 +308,7 @@ def start_span(
 
       def inject(span_context, format, carrier)
         case format
-        when ::OpenTracing::FORMAT_RACK
+        when ::OpenTracing::FORMAT_RACK, ::OpenTracing::FORMAT_TEXT_MAP
           carrier['elastic-apm-traceparent'] =
             span_context.traceparent.to_header
```
eric-christian, line 313: We had the situation that span_context was a `OpenTracing::SpanContext.NOOP_INSTANCE` which does not know about `traceparent`. It probably derived from a situation where `start_span` returned a `OpenTracing::Span::NOOP_INSTANCE`.

```suggestion
            span_context.traceparent.to_header if span_context.respond_to?(:traceparent)
```

estolfo: thanks!

mikker:
@estolfo I think this has been verified to work (https://github.com/elastic/apm-agent-ruby/issues/754#issuecomment-642121080) – do you need a review from me?

estolfo:
@mikker I think the changes should be good, as @eric-christian said they work. Are you ok with me merging them or do you want to do a review?

mikker:
Remember a changelog entry!

PR #801, opened 2020-05-20
Do not use URI.build as it disallows _ in hostname

hqm42:
URI.build throws an InvalidComponentError if the host contains an undescrore.

When making requests to hosts with an underscore in its name (e.g. myapp_mysql) the apm spy throws an exception.

> This is just a quickfix. Please try to find a better solution to prevent this error.


mikker:
Thank you for reporting this 👍
I fixed it in a separate PR as you said this was just a temp quickfix.

PR #802, opened 2020-05-20
ci: run ITs and master tests in parallel

v1v:

Run some other tests in parallel to report the status in advance. ITs reporting is kept as used to be.

## Why is it important?

Flakiness might skip the run of the ITs for the given changes, so let's honor them and run them and wait for their outcome.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -95,14 +95,14 @@ pipeline {
     Execute unit tests.
     */
     stage('Tests') {
+      when {
+        beforeAgent true
+        expression { return env.ONLY_DOCS == "false" }
+      }
       failFast false
       parallel {
         stage('Tests') {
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -94,56 +94,58 @@ pipeline {
...
+              expression { return env.ONLY_DOCS == "false" }
+              anyOf {
+                changeRequest()
+                expression { return !params.Run_As_Master_Branch }
+              }
+            }
+          }
+          steps {
+            withGithubNotify(context: 'Integration Tests') {
+              build(job: env.ITS_PIPELINE, propagate: false, wait: true,
```
v1v, line 142:
```suggestion
              build(job: env.ITS_PIPELINE, propagate: true, wait: true,
```

PR #807, opened 2020-05-26
GraphQL: Get operation name from selected_operation

### LINKED ISSUE #779: GraphQL Tracer: transactions are all unnamed
## Describe the bug
All transactions that are recorded with the GraphQL tracer are bucketed into `GraphQL: [unnamed]`.

## Steps to reproduce
Fire any query to the backend such as `query HelloWorld { viewer { user { id } } }`

## Expected behavior
Should see a transaction `GraphQL: HelloWorld`

## Actual behavior
Should see a transaction `GraphQL: [unnamed]`

## Environment
- **OS**: Linux
- **Ruby version**: 2.5.7
- **Framework and version**: 5.1.6.2
- **APM Server version**: ?
- **Agent version**: 3.6.0

## Additional context
I don't think this is the correct way to extract the operation name: https://github.com/elastic/apm-agent-ruby/blob/master/lib/elastic_apm/graphql.rb#L60

If you look at any of the other custom tracers, such as Datadog, https://github.com/rmosolgo/graphql-ruby/blob/1a9a20f3da629e63ea8e5ee8400be82218f9edc3/lib/graphql/tracing/data_dog_tracing.rb#L32, the correct method is `query.selected_operation_name` as indicated by https://github.com/rmosolgo/graphql-ruby/blob/1a9a20f3da629e63ea8e5ee8400be82218f9edc3/lib/graphql/query.rb#L61

mikker:
Fixes #779, hopefully

### DIFF
--- a/lib/elastic_apm/graphql.rb
+++ b/lib/elastic_apm/graphql.rb
@@ -74,7 +74,7 @@ module ElasticAPM
       private
 
       def query_name(query)
-        query.operation_name || UNNAMED
+        query.selected_operation_name || UNNAMED
       end
 
       def concat_names(results)


mikker:
@rtymchyk can you test if this change fixes your unnamed operations?

rtymchyk:
> @rtymchyk can you test if this change fixes your unnamed operations?

Yeah I can probably check sometime today and get back to you!

rtymchyk:
@mikker I've installed the gem as so: `gem 'elastic-apm', git: 'https://github.com/mikker/apm-agent-ruby.git', branch: 'graphql-op-name'` but unfortunately still seeing all the queries as `GraphQL: [unnamed]`. Let me know if there's any other information I can provide.

mikker:
Could you by any chance provide an example request? That would make this much easier 😊

rtymchyk:
@mikker As I started digging into this, I realized that this is maybe a bug at a different level (with another third party library). Our web client fires graph queries to a gateway (Apollo), which then disperses to downstream services, one of which is running this APM agent, but looking at the logs it looks like the gateway may be stripping operation names.

Sorry for the waste of time.

 I _think_ this PR may still be worthwhile since it seems like `selected_operation_name ` is how other tracers operate, but I'll need to figure out the other issue before I can test this PR out again.

mikker:
No worries! Glad you figured it out 😊 

I’ll leave the agent as is for now. If it ain’t broke…, you know. Thanks for using ElasticAPM!

Issue #811, opened 2020-05-27
Patching `Kernel#require` considered harmful


v-kolesnikov:
### Objective

elastic-apm pathes `Kernel#require` that is bad itself. Moreover it breaks Sidekiq spy. 

Script to reproduce a bug:

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'elastic-apm', '3.6.0', require: false
  gem 'sidekiq', require: false
end

require 'elastic_apm'
ElasticAPM.start(
  service_name: 'Lobster',
  environment: 'development',
  framework_name: 'Rack',
  framework_version: ''
)

require 'sidekiq'
```

```shell
$ ruby server.rb
Failed hooking into 'sidekiq'. Please report this at github.com/elastic/apm-agent-ruby
/Users/b.white/.rvm/gems/ruby-2.7.1/gems/elastic-apm-3.6.0/lib/elastic_apm/spies/sidekiq.rb:52:in `install_processor'
/Users/b.white/.rvm/gems/ruby-2.7.1/gems/elastic-apm-3.6.0/lib/elastic_apm/spies/sidekiq.rb:78:in `install'
/Users/b.white/.rvm/rubies/ruby-2.7.1/lib/ruby/2.7.0/forwardable.rb:235:in `install'
/Users/b.white/.rvm/gems/ruby-2.7.1/gems/elastic-apm-3.6.0/lib/elastic_apm/spies.rb:53:in `hook_into'
/Users/b.white/.rvm/gems/ruby-2.7.1/gems/elastic-apm-3.6.0/lib/elastic_apm/spies.rb:76:in `require'
/Users/b.white/.rvm/gems/ruby-2.7.1/gems/sidekiq-6.0.7/lib/sidekiq/api.rb:3:in `<top (required)>'
/Users/b.white/.rvm/gems/ruby-2.7.1/gems/elastic-apm-3.6.0/lib/elastic_apm/spies.rb:73:in `require'
/Users/b.white/.rvm/gems/ruby-2.7.1/gems/elastic-apm-3.6.0/lib/elastic_apm/spies.rb:73:in `require'
/Users/b.white/.rvm/gems/ruby-2.7.1/gems/sidekiq-6.0.7/lib/sidekiq/scheduled.rb:5:in `<top (required)>'
[...]

Issue #800, opened 2020-05-20
Provide an opt-out for patching require


mikker:
_Putting this here to not forget._

Perhaps we want to provide a way to opt out of how we patch `require`? For the users who want to be in full control.

estolfo:
Related #811

PR #812, opened 2020-05-28
Add special setting to disable patching Kernel#require

mikker:
As discussed in #811 and #800.

This isn't an option in `Config` as it will have to be set when the process boots.

Had to remove `-r spec_helper` from `.rspec` to be able to switch the setting on before the test. That forced me to add it again to all the specs that were relying on it already being there.


spec/integration/skip_require_patch_spec.rb
```diff
@@ -17,7 +17,14 @@
 
 # frozen_string_literal: true
 
-module ElasticAPM
-  RSpec.describe Context::Request do
+# Deliberately setting this before requiring spec_helper.rb so it's set
+# before we require the agent
+ENV['ELASTIC_APM_SKIP_REQUIRE_PATCH'] = '1'
```
estolfo, line 22: Do you want to cache the original value of the env variable and then set it again after this test? I know it's not really necessary but it would just be so that this test is completely contained.

mikker: As you mention; the integration tests are run in their own processes and setting it like this doesn't `export` it to the outside `env`. I don't want to add a _fix_ for something that isn't in fact broken.

estolfo: It would be an improvement, not a fix :)
I am just always wary of making changes to something global in a single test.

docs/configuration.asciidoc
```diff
@@ -890,3 +890,11 @@ The unit is provided as suffix directly after the number, without any separation
  * `gb` (gigabytes)
 
 NOTE: we use the power-of-two sizing convention, e.g. `1 kilobyte == 1024 bytes`
+
+[float]
+[[special-configuration]]
+=== Special configuration
+
+Elastic APM patches `Kernel#require` to auto-detect and instrument supported third party libraries. It does so with the utmost care but for the purists this may be unacceptable.
```
estolfo, line 898: hmm saying that it might be unacceptable for purists sounds a tad snarky. How about something more neutral like "sometimes this conflicts with other libraries so we provide a setting to turn this functionality off"

mikker: You're right 👍

Issue #803, opened 2020-05-22
MongoDB and Elasticsearch performance affected by elastic-apm gem


lgiacomini92:
## Describe the bug
This issue is similar to #105.

Our application had several performance issues right after we added the `elastic-apm` gem into the project.
We tried to set the attribute `transaction_sample_rate: 0.1` at `config/elastic-apm.yml`, but the problem has not been solved.

Our application is running on AWS and ELK is on GCP, because of that we are using a VPN to connect both. We are also using Datadog agent in this project.

**MongoDB metrics:**
![image (80)](https://user-images.githubusercontent.com/6638683/82674067-252c5800-9c19-11ea-9826-d3a01035c697.png)

**Elasticsearch metrics:**
![image (79)](https://user-images.githubusercontent.com/6638683/82674074-26f61b80-9c19-11ea-9f9e-7c539503f328.png)

## Steps to reproduce
Add the `elastic-apm` gem into the project.

## Expected behavior
It does not affect MongoDB or Elasticsearch performance.

## Environment
- **Docker image**: `ruby:2.5.7`
- **OS**: Linux Debian
- **Ruby version**: 2.5.7
- **Framework and version**: Rails 5.2.4
- **APM Server version**: 7.7
- **Agent version**: 3.6.0

mikker:
Hi @lgiacomini92! Thank you for reporting this. This is obviously not intended.

Could you try maybe adding the gem, but skipping the ES and Mongo spies? That could perhaps tell us whether the spies themselves are the problem.

```yaml
# config/elastic_apm.yml
disable_instrumentations: ['json', 'elasticsearch', 'mongo_db'] # json is off by default
```

lgiacomini92:
Hi @mikker.
We did this test by skipping the ES and Mongo spies and the results were the same, both had affected performance after a few hours.

**MongoDB:**
![image (90)](https://user-images.githubusercontent.com/6638683/83020443-dad31e80-9ffe-11ea-9122-c5e835fb7195.png)

**Elasticsearch:**
![image (89)](https://user-images.githubusercontent.com/6638683/83020433-d870c480-9ffe-11ea-8d1d-d8c0fa00963a.png)

mikker:
How do you mean _after a few hours_? Does the overhead show up as soon as you run with the agent or do you have to run the process for some time before it shows up?

lgiacomini92:
The application has to run the process for some time before it shows up.
[...]

PR #817, opened 2020-06-12
Mongo spy: Remove event before checking for span

mikker:
Re #803, this ensures we're not keeping all events around.


mikker:
That _"Build Succeeded"_ 💚

mikker:
Merging this as it could be a _potential_ memory leak. No confirmed cases, no alarm.

PR #825, opened 2020-06-25
Set view paths from ActionController::Base (#822)

f-mer:
## What does this pull request do?

Resolves #822 

## Checklist

- [ ] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/). 
- [x] My code follows the style guidelines of this project (See `.rubocop.yml`)
- [x] I have rebased my changes on top of the latest master branch

- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing [**unit** tests](https://github.com/elastic/apm-agent-ruby/blob/master/CONTRIBUTING.md#testing) pass locally with my changes

- [ ] I have made corresponding changes to the documentation
- [ ] I have updated [CHANGELOG.asciidoc](CHANGELOG.asciidoc)
- [ ] I have updated [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)
- [ ] Added an API method or config option? Document in which version this will be introduced

PR #839, opened 2020-08-03
Get view paths from ActionController::Base

mikker:
See https://github.com/elastic/apm-agent-ruby/pull/825


mikker:
Test failures are only on rails `master`

PR #858, opened 2020-08-20
Add sampling info to Tracestate headers

mikker:
See https://github.com/elastic/apm/pull/307 (WIP)

### DIFF
--- a/lib/elastic_apm/trace_context/tracestate.rb
+++ b/lib/elastic_apm/trace_context/tracestate.rb
@@ -24,73 +24,103 @@ module ElasticAPM
       # @api private
       class Entry
         def initialize(key, value)
-          @key, @value = key, value
-          parse! if key == 'es'
+          @key = key
+          @value = value
         end
 
-        attr_reader :key, :values
-
-        def set(k, v)
-          if key != 'es'
-            raise ArgumentError,
-              'trying to set a value in a non-Elastic tracestate entry'
-          end
-
-          @values[k.to_s] = v.to_s
-        end
-
-        def get(k)
-          values[k.to_s]
-        end
-
-        def value
-          return @value unless values
-          values.map { |(k, v)| "#{k}:#{v}" }.join(';')
-        end
+        attr_reader :key, :value
 
         def to_s
           "#{key}=#{value}"
         end
+      end
+
+      class EsEntry
+        ASSIGN = ':'
+        SPLIT = ';'
+
+        SHORT_TO_LONG = { 's' => 'sample_rate' }
+        LONG_TO_SHORT = { 'sample_rate' => 's' }
+
+        def initialize(values = nil)
+          parse(values)
+        end
+
+        attr_reader :sample_rate
+
+        def key
+          'es'
+        end
+
+        def value
+          LONG_TO_SHORT.map do |l, s|
+            "#{s}#{ASSIGN}#{send(l)}"
+          end.join(SPLIT)
+        end
+
+        def empty?
+          !sample_rate
+        end
+
+        def sample_rate=(val)
+          float = Float(val).round(3)
+
+          return nil unless (0.0..1.0).include?(float)
+
+          @sample_rate = float
+        rescue ArgumentError => e
+          nil
+        end
+
+        def to_s
+          return nil if empty?
+
+          "es=#{value}"
+        end
 
         private
 
-        def parse!
-          @values = Hash[value.split(';').map { |kv| kv.split(':') }]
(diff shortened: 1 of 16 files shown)


spec/elastic_apm/trace_context_spec.rb
```diff
@@ -148,7 +150,7 @@ module ElasticAPM
           calls = {}
           block = ->(k, v) { calls[k] = v }
 
-          subject.tracestate = TraceContext::Tracestate.parse('a=b')
+          subject.tracestate = TraceContext::Tracestate.parse('es=s:1.0,b=na')
```
axw, line 153: not the end of the world, but it would be nice to format in the minimal representation: `1` instead of `1.0`

lib/elastic_apm/transaction.rb
```diff
@@ -52,13 +53,19 @@ def initialize(
...
       @context = context || Context.new # TODO: Lazy generate this?
       if @default_labels
         Util.reverse_merge!(@context.labels, @default_labels)
       end
 
-      @trace_context = trace_context || TraceContext.new(recorded: sampled)
+      unless (@trace_context = trace_context)
+        @trace_context = TraceContext.new(
+          traceparent: TraceContext::Traceparent.new(recorded: sampled),
+          tracestate: TraceContext::Tracestate.new(sample_rate: sample_rate)
```
axw, line 66: I think you're missing the a special case, setting sample_rate=0 when the transaction is not sampled

axw (commented):
Looks good overall. What happens if the `es` vendor value is invalid?
Is there a test that covers the case where a downstream transaction receives `traceparent`, and either `tracestate` is missing or doesn't have a valid `es` entry? In that case the transaction should have no `sample_rate` field.

spec/elastic_apm/trace_context_spec.rb
mikker: I ended up not bothering because the potential savings are miniscule ¯\_(ツ)_/¯

mikker:
> What happens if the es vendor value is invalid?

Not sure but depends on what we mean by _invalid_? Not matching the `k:v` pattern or including an illegal utf char or something?

lib/elastic_apm/transaction.rb
mikker: Done

axw:
>> What happens if the es vendor value is invalid?
>
> Not sure but depends on what we mean by invalid? Not matching the k:v pattern or including an illegal utf char or something?

Yes (all of the above :))

Here's a few test cases which I think we should consider, to be defensive:

- "es=foo"
- "es=s:foo"
- "es=s:1.5" (`s` should be in the range 0..1)

spec/elastic_apm/trace_context/tracestate_spec.rb
```diff
@@ -22,31 +22,69 @@
...
       end
 
-      context 'with mixed' do
-        subject { described_class.parse("a=b,c=d\ne=f") }
-        its(:to_header) { is_expected.to eq 'a=b,c=d,e=f' }
+      context 'with bad values' do
+        [
+          ['es=xyz', nil],
+          ['es=s:foo', nil],
+          ['es=s:1.5', nil]
```
mikker, line 60: @axw Do you feel like I should log these failures to parse?

axw: I wouldn't, or at least if you do make it trace level. Otherwise agent logs could be flooded by sending a lot of bogus requests.

axw (approved):
LGTM. I feel like we might all benefit from a Gherkin spec for this, WDYT?

simitt (approved):
Didn't follow semantics to closely, but LGTM from a ruby perspective.

lib/elastic_apm/trace_context/tracestate.rb
```diff
@@ -21,22 +21,127 @@ module ElasticAPM
...
-        values.join(',')
+        return "" unless entries.any?
+
+        entries.values.map(&:to_s).join(',')
+      end
+
+      private
+
+      def es_entry
+        # laze generate this so we only add it if necessary
```
mikker, line 131:
```suggestion
        # lazy generate this so we only add it if necessary
```

mikker:
@axw 

> I feel like we might all benefit from a Gherkin spec for this, WDYT?

I agree. As this would add significant complexity (incoming and outgoing reqs) to the currently very simple gherkin specs (spec, really), I'm going to add it in a separate PR.

PR #861, opened 2020-08-25
Disable flaky specs

mikker:
We need to get back to a green build.


mikker:
jenkins, run the tests

PR #880, opened 2020-10-16
Implement sample_rate precision spec

### LINKED ISSUE #868: Implement sample_rate precision spec
You can view the spec changes [here](https://github.com/elastic/apm/pull/335).

estolfo:
This PR adds a `PrecisionValidator` in the Utils that can ensure a certain number of digits of precision of a number between 0 and 1. It optionally ensures a minimum.

The `PrecisionValidator` is used
- by a `Precision` Converter in the config when `transaction_sample_rate` is loaded
- by a `Tracestate` when `sample_rate` is set.

Closes #868

### DIFF
--- a/lib/elastic_apm/config/precision.rb
+++ b/lib/elastic_apm/config/precision.rb
@@ -1,0 +1,31 @@
+# Licensed to Elasticsearch B.V. under one or more contributor
+# license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright
+# ownership. Elasticsearch B.V. licenses this file to you under
+# the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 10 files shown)


lib/elastic_apm/config/precision.rb
```diff
@@ -0,0 +1,31 @@
...
+# under the License.
+
+# frozen_string_literal: true
+
+require 'elastic_apm/util/precision_validator'
+
+module ElasticAPM
+  class Config
+    # @api private
+    class Precision
```
mikker, line 25: Can we come up with a more descriptive name maybe? `RoundedFloat` or who knows.

mikker (approved):
Only one small comment, otherwise LGTM!

lib/elastic_apm/config/precision.rb
estolfo: ok done!

PR #882, opened 2020-10-19
Fix Kubernetes metadata serialization

mikker:


mikker:
Seems the MongoDB integration (hosted) test is failing on JRuby. I don't even want to try to debug why that is. Should we just skip it somehow, @estolfo?

estolfo:
@mikker I don't see that it failed in this build https://apm-ci.elastic.co/blue/organizations/jenkins/apm-integration-tests-selector-mbp/detail/master/11014/pipeline
Am I missing something?

PR #883, opened 2020-11-02
Add transaction and span outcome

### LINKED ISSUE #852: Add outcome to transactions and spans
See https://github.com/elastic/apm/pull/299

estolfo:
Closes #852 

Todo:

- [ ] grpc
- [x] elasticsearch
- [x] sequel
- [x] Should the `http_status_code` argument be converted to an `Integer` or required to be an `Integer` in the above methods?


lib/elastic_apm.rb
```diff
@@ -381,5 +391,29 @@ def add_filter(key, callback = nil, &block)
...
     end
+
+    def set_transaction_outcome(result: nil, http_status_code: nil)
+      return unless transaction = current_transaction
+      return transaction.outcome = result if result
+      return unless http_status_code
+      code = http_status_code.to_i
+      if code < 500
+        transaction.outcome = 'success'
+      elsif code >= 500
```
mikker, line 402: It's never _neither_ above _and_ below 500, so this will do:

```ruby
transaction.outcome = http_status_code.to_i < 500 ? 'success' : 'failure'
```

Or if you prefer:

```ruby
transaction.outcome =
  if http_status_code.to_i < 500
    'success'
  else 
    'failure'
  end
```

lib/elastic_apm.rb
```diff
@@ -381,5 +391,29 @@ def add_filter(key, callback = nil, &block)
 
       agent&.add_filter(key, block || callback)
     end
+
+    def set_transaction_outcome(result: nil, http_status_code: nil)
+      return unless transaction = current_transaction
+      return transaction.outcome = result if result
+      return unless http_status_code
```
mikker, line 398: I think this is a bit hard to read. I think a few extra newlines or a comment would help us in the future.

If both `result` and `http_status_code` are `nil`, why even call the method? Is that instead an ArgumentError? What if both are passed?

I think I feel the API of the method is a bit vague.

lib/elastic_apm.rb
```diff
@@ -381,5 +391,29 @@ def add_filter(key, callback = nil, &block)
 
       agent&.add_filter(key, block || callback)
     end
+
+    def set_transaction_outcome(result: nil, http_status_code: nil)
```
mikker, line 395: I'm not sure why we want a public(?) API for this?

As far as I can see, we already have direct access to either `transaction` or `span` where the outcome is set, so couldn't it instead just be set as a simple prop?

lib/elastic_apm.rb
mikker: Also, the knowledge of HTTP status codes might not tie directly to the concept of _outcome_, hence it might not even belong here, but in the surrounding context ie. the middleware.

mikker (changes requested):
I'm a little bit confused by this approach.

How about something like

```ruby

class Transaction
  SUCCESS = 'success'
  FAILURE = 'failure'

  # ...

  def success!
    self.outcome = SUCCESS
  end

  def failure!
    self.outcome = FAILURE
  end
end
```

This ties the details of how to signify the two values into the class itself. That way we don't have to throw around the magic strings `success` and `failure` in the whole lib.

mikker:
Could even be something like 

```ruby
class Transaction
  class Outcome
    def initialize(value)
      @value = value
    end

    def self.success
      new('success')
    end

    def self.failure
      new('failure')
    end
    
    def self.from_status_code(code)
      new(code >= 500 ? 'failure' : 'success')
    end
  end

  def success!
    self.outcome = Outcome.success
  end

  def failure!
    self.outcome = Outcome.failure
  end
end
```

Such OO, very message passing 😅

lib/elastic_apm/span.rb
```diff
@@ -25,6 +25,15 @@ class Span
     extend Forwardable
     include ChildDurations::Methods
 
+    class Outcome
```
mikker, line 28:
```suggestion
    @api private
    class Outcome
```

lib/elastic_apm/transaction.rb
```diff
@@ -20,6 +20,16 @@
 module ElasticAPM
   # @api private
   class Transaction
+
+    class Outcome
```
mikker, line 24:
```suggestion
    @api private
    class Outcome
```

lib/elastic_apm/span.rb
mikker: Prevents YARD from generating api docs

mikker (commented):
Thanks!

lib/elastic_apm.rb
```diff
@@ -183,7 +183,14 @@ def with_transaction(
             context: context,
             trace_context: trace_context
           )
-        yield transaction
+        result = yield transaction
+        transaction&.outcome =
+          Transaction::Outcome::SUCCESS unless transaction&.outcome
```
mikker, line 188:
```suggestion
        transaction&.outcome ||= Transaction::Outcome::SUCCESS
```

Is perhaps simpler?

lib/elastic_apm/span.rb
```diff
@@ -25,6 +25,16 @@ class Span
     extend Forwardable
     include ChildDurations::Methods
 
+    # @api private
+    class Outcome
+      FAILURE = "failure"
+      SUCCESS = "success"
+
+      def self.from_status(code)
```
mikker, line 33:
```suggestion
      def self.from_http_status(code)
```

Although that might be obvious it doesn't hurt to be explicit

lib/elastic_apm/spies/delayed_job.rb
```diff
@@ -42,9 +42,11 @@ def self.invoke_job(job, *args, &block)
         transaction = ElasticAPM.start_transaction(job_name, TYPE)
         job.invoke_job_without_apm(*args, &block)
         transaction&.done 'success'
+        transaction&.outcome = Transaction::Outcome::SUCCESS
```
mikker, line 45:
```suggestion
        transaction&.outcome ||= Transaction::Outcome::SUCCESS
```

lib/elastic_apm/spies/delayed_job.rb
```diff
@@ -42,9 +42,11 @@ def self.invoke_job(job, *args, &block)
...
       rescue ::Exception => e
         ElasticAPM.report(e, handled: false)
         transaction&.done 'error'
+        transaction&.outcome = Transaction::Outcome::FAILURE
```
mikker, line 49:
```suggestion
        transaction&.outcome ||= Transaction::Outcome::FAILURE
```

mikker (commented):
Generally looks good!

We're a bit inconsistent about whether or not we allow users to set the outcome themselves. Sometimes we take it into account, sometimes only when `'success'`, sometimes not.

Is there a reason we couldn't _always_ let the user set it themselves? What does the spec suggest?

estolfo:
Originally, I only took into account whether `outcome` was already set in `with_transaction` and `with_span` because the user would run their own code and could determine themselves what was a success or failure. I don't think it's necessary to check whether the outcome is already set for spans or transactions we create ourselves in the spies. Unless you disagree?

mikker (approved):
🚀

PR #888, opened 2020-11-10
Update dynamo db test

estolfo:
There are two more operations available via the latest `aws-sdk-dynamodb` gem that were missing from the required params hash. This was causing the test to fail as it iterated over `::Aws::DynamoDB::Client.api.operation_names`


estolfo:
closing in favor of #889

Issue #753, opened 2020-04-06
Conflict with the Airbrake gem


paul-mesnilgrente:
## Describe the bug

I upgraded to the latest version of Airbrake (10.0.2) and Elastic-APM (3.6.0) and I started to have a `SystemStackError` in my controller. After some time I found that the stack was too deep. Logging the backtrace showed:

```
"/app/vendor/cache/ruby/2.6.0/gems/elastic-apm-3.6.0/lib/elastic_apm/spies/net_http.rb:72:in `block in request'"
"/app/vendor/cache/ruby/2.6.0/gems/elastic-apm-3.6.0/lib/elastic_apm.rb:273:in `with_span'"
"/app/vendor/cache/ruby/2.6.0/gems/elastic-apm-3.6.0/lib/elastic_apm/spies/net_http.rb:62:in `request'"
"/app/vendor/cache/ruby/2.6.0/gems/airbrake-10.0.2/lib/airbrake/rails/net_http.rb:11:in `block in request'"
"/app/vendor/cache/ruby/2.6.0/gems/airbrake-10.0.2/lib/airbrake/rack.rb:26:in `block in capture_timing'"
"/app/vendor/cache/ruby/2.6.0/gems/airbrake-ruby-4.13.3/lib/airbrake-ruby/timed_trace.rb:29:in `span'"
"/app/vendor/cache/ruby/2.6.0/gems/airbrake-ruby-4.13.3/lib/airbrake-ruby/timed_trace.rb:18:in `block in span'"
"/app/vendor/cache/ruby/2.6.0/gems/airbrake-ruby-4.13.3/lib/airbrake-ruby/timed_trace.rb:18:in `tap'"
"/app/vendor/cache/ruby/2.6.0/gems/airbrake-ruby-4.13.3/lib/airbrake-ruby/timed_trace.rb:18:in `span'"
"/app/vendor/cache/ruby/2.6.0/gems/airbrake-10.0.2/lib/airbrake/rack.rb:25:in `capture_timing'"
"/app/vendor/cache/ruby/2.6.0/gems/airbrake-10.0.2/lib/airbrake/rails/net_http.rb:10:in `request'"

"/app/vendor/cache/ruby/2.6.0/gems/elastic-apm-3.6.0/lib/elastic_apm/spies/net_http.rb:72:in `block in request'"
[...]

PR #890, opened 2020-11-11
Use Module#prepend where possible

mikker:
This leaves us with only the module based APIs still using `class_eval`, specifically `delayed_job` and `sucker_punch`. The problem being that at the time of hooking up the integration, the target modules have already been included and so prepending doesn't modify the including classes.

I propose we leave them as is for now and change them in a future version if we come up with a scheme or they introduce some sort of plugin api that could help us avoid `class_eval`.

This PR is a breaking change not because it removes any features but because it might interfere with other libraries still using class_eval and thereby possibly introduce the infinite loop when trying to look up stacktraces after an exception([#753 ](https://github.com/elastic/apm-agent-ruby/issues/753) for example).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,8 @@ gem 'pry'
 gem 'rack-test'
 gem 'rspec', '~> 3'
 gem 'rspec-its'
+gem 'rubocop', require: false
+gem 'rubocop-performance', require: false
 gem 'timecop'
 gem 'webmock'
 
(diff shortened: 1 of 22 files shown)


mikker:
😅 https://github.com/elastic/apm-agent-ruby/blob/9557a3c79dd9ded69cbc706bbffb47d7593cdf0e/lib/elastic_apm/span_helpers.rb#L39-L49

estolfo (approved):
Looks good-- seems a lot more straightforward than I thought it would be.
You mentioned wanting to change the name of _Spies_, do you still want to? We could call them `extensions`, as they are each in a module `Ext`. Or `integrations`, as I think you suggested.

mikker:
I sort of still do. We'd have to do it in 2 steps though as we'd be renaming some config options too. So step 1 is soft deprecation, then remove in major.

estolfo:
ok, I guess it's not urgent then and doesn't necessarily have to be part of this PR.

PR #891, opened 2020-11-11
Fix RuboCop complaints

estolfo:
These changes fix all the RuboCop complaints. Points that still need to be discussed/decided:

- [x] Do we want to enforce the use of single quotes for String literals? We use double quotes often and I've changed all of them to single quotes in this PR.
- [x] We use `rescue` with no specified exception class many times and I think it's often necessary. So should we disabled the `Style/RescueStandardError` cop?
- [x] I changed the `TargetRubyVersion` in `rubocop.yml` because the latest version of the gem doesn't support 2.3. So we can't run rubocop for ruby 2.3 and maybe it'll allow syntax that is not valid for 2.3. Though I think the only syntax introduced in 2.4 is [multiple assignment in conditionals](https://rubyreferences.github.io/rubychanges/2.4.html#multiple-assignment-allowed-in-conditional-expression). UPDATE: `Regexp.match?` is new in ruby 2.4. So I'll have to undo all those changes from `=~` to `match?` and the tests will fail on ruby-2.3 right now.


mikker:
* I think I like `"` the most. Always a hassle when you start typing something and come upon a `'`, like `isn't`, or realize you want to `#{}` something, and have to go back and change the leading quote.
* Agree about the `rescue` rule, let's disable.
* `Regexp#match` without the `?` is in < 2.4 fwiw

.rubocop.yml
```diff
@@ -24,6 +24,12 @@ Layout/EmptyLineAfterGuardClause:
 Layout/FirstParameterIndentation:
   Enabled: false
 
+Layout/EmptyLinesAroundAttributeAccessor:
+  Enabled: true
```
mikker, line 28: Don't all the rules come enabled? Do we need to explicitly enable these?

estolfo: I would think so but I got this message for the cops that I added to the file: `The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file.`

mikker: Apparently so then 😊

estolfo: ah, I see you can add something to the file indicating you want to opt-in to all new cops.

lib/elastic_apm/span.rb
```diff
@@ -50,7 +50,7 @@ def initialize(
       context: nil,
       stacktrace_builder: nil,
       sync: nil,
-      sample_rate: nil
+      _sample_rate: nil
```
mikker, line 53: If we don't use this, should we drop it?

spec/elastic_apm/transport/base_spec.rb
```diff
@@ -31,6 +31,7 @@ module Transport
       end
 
       describe '#start' do
+        pending
       end
```
mikker, line 36:
```suggestion
```

spec/elastic_apm/transport/base_spec.rb
```diff
@@ -56,7 +57,9 @@ module Transport
       describe '#submit' do
         before do
           # Avoid emptying the queue
+          # rubocop:disable Lint/EmptyBlock
           allow(subject).to receive(:ensure_watcher_running) {}
+          # rubocop:enable Lint/EmptyBlock
```
mikker, line 62:
```suggestion
          allow(subject).to receive(:ensure_watcher_running) { nil }
```

🤷‍♂️

mikker (approved):
Looks great! Do you want to re-enable it on CI in this PR or a separate one?

lib/elastic_apm/span.rb
estolfo: sure🤷🏻‍♀️

PR #896, opened 2020-11-17
Handle invalid utf-8 byte sequences in sql summarizer and DB statement

### LINKED ISSUE #895: ElasticAPM::Sql module breaks when a sql string contains a binary column value
## Describe the bug
I add apm-agent-ruby into my rails application. In most cases, apm-agent can correctly capture SQL of ActiveRecord. However, when a SQL string contains a  binary column value,  apm-agent raises an error (ArgumentError: invalid byte sequence in UTF-8), which breaks my application.

## Steps to reproduce
1. an example raw SQL string
```
sql_str =<<EOF
 INSERT INTO "checksums" ("checksum") VALUES ( ",&\xB4kh\xFF\xC6\x8F\xF9\x9BE<\x1D0A4\x13B-pd\x83\xBF\xA0\xF9\x8A^\x88bf\xE7\xAE")
EOF
```

2. apm-agent tries to capture this SQL string

```
# https://github.com/elastic/apm-agent-ruby/blob/a958ef697bd93eed2c242aa92e65bb290713b128/lib/elastic_apm/sql/tokenizer.rb#L46

# same as the following code

SPACE = /[[:space:]]+/.freeze
scanner = StringScanner.new(input)
scanner = scanner.Scan(SPACE)  # raise a error

```

## Expected behavior

apm-agent correctly captures that raw SQL string 

## Current behavior

apm-agent raises an error 
```
ArgumentError: invalid byte sequence in UTF-8
  from elastic_apm/sql/tokenizer.rb:47:in `skip'
  from elastic_apm/sql/tokenizer.rb:47:in `scan'
  from elastic_apm/sql/signature.rb:44:in `parse'
  from elastic_apm/sql/signature.rb:57:in `parse'
```

## Environment

- **OS**: linux
- **Ruby version**: 2.5.0
- **Framework and version**: e.g. Rails 5.0.2
- **Agent version**: 3.11.0

estolfo:
Fixes #895 

- Enforce utf-8 encoding of sql strings and replace invalid characters with �.
- Enforce utf-8 encoding of the DB statement and replace invalid characters with �. 

Note that we must do this when the `Sql::Signature` object is created, as opposed to in the `Sql::Tokenizer` because in case the string cannot be parsed, [the first character is returned via `@sql.split`](https://github.com/elastic/apm-agent-ruby/blob/v3.12.1/lib/elastic_apm/sql/signature.rb#L50). That will fail if the first character is an invalid utf-8 byte sequence.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -35,6 +35,16 @@ endif::[]
 [[release-notes-3.x]]
 === Ruby Agent version 3.x
 
+[float]
+[[unreleased]]
+==== Unreleased
+
+[float]
+===== Fixed
+
+- Handle invalid utf-8 byte sequences in sql summarizer and DB statement {pull}896[#896]
+
+
 [[release-notes-3.12.1]]
 ==== 3.12.1 (2020-11-16)
 
--- a/lib/elastic_apm/span/context/db.rb
+++ b/lib/elastic_apm/span/context/db.rb
@@ -30,7 +30,7 @@ module ElasticAPM
           rows_affected: nil
         )
           @instance = instance
-          @statement = statement.encode('utf-8', invalid: :replace, undef: :replace)
+          @statement = statement&.encode('utf-8', invalid: :replace, undef: :replace)
           @type = type
           @user = user
           @rows_affected = rows_affected
--- a/lib/elastic_apm/span/context/db.rb
+++ b/lib/elastic_apm/span/context/db.rb
@@ -30,7 +30,7 @@ module ElasticAPM
           rows_affected: nil
         )
           @instance = instance
-          @statement = statement
+          @statement = statement.encode('utf-8', invalid: :replace, undef: :replace)
           @type = type
           @user = user
           @rows_affected = rows_affected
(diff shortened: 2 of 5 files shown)


mikker (approved):
Great!

What about `statement`? Do we need to handle binary there too?

mikker:
(https://github.com/elastic/apm-agent-ruby/blob/4f745350eddbc400520b368f9ab164179277a6e9/lib/elastic_apm/span/context/db.rb#L33)

estolfo:
not unless we call methods on the string. Simply setting it and passing it along in the context won't matter. I'll double check though.

mikker:
We evetually convert it to JSON which will fail, I think. And we catch those failures, but maybe we want to do it upfront instead?

estolfo:
Yes, you're right. You'd get this with invalid utf-8 bytes when serializing to JSON:
`JSON::GeneratorError: source sequence is illegal/malformed utf-8`

mikker (approved):
LGTM 👍

mikker:
Remember changelog entry

PR #900, opened 2020-11-19
Remove deprecated options and settings

mikker:
In preparation for the upcoming major release, 4.0.


estolfo (changes requested):
[This test](https://github.com/elastic/apm-agent-ruby/blob/master/spec/integration/rails_spec.rb#L268-L269) is failing.
I believe you can just remove it as you removed the default filters in the `HashSanitizer`.
Other than that, this LGTM.

estolfo (approved):

PR #901, opened 2020-11-20
Remove support for ruby 2.3 and jruby-9.1

estolfo:
- Use `transform_values` instead of `each_with_object`
- Use `match?` instead of `=~`
- Remove ruby 2.3 from Jenkins test matrices
- Remove jruby-9.1 from Jenkins test matrices

Note that jruby-9.1 supports ruby 2.3 so we must remove support for that as well.

### DIFF
--- a/lib/elastic_apm/context/response.rb
+++ b/lib/elastic_apm/context/response.rb
@@ -38,9 +38,7 @@ module ElasticAPM
       attr_reader :headers
 
       def headers=(headers)
-        @headers = headers&.each_with_object({}) do |(k, v), hsh|
-          hsh[k] = v.to_s
-        end
+        @headers = headers&.transform_values(&:to_s)
       end
     end
   end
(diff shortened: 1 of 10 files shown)


lib/elastic_apm/context/response.rb
```diff
@@ -38,9 +38,7 @@ def initialize(
       attr_reader :headers
 
       def headers=(headers)
-        @headers = headers&.each_with_object({}) do |(k, v), hsh|
-          hsh[k] = v.to_s
-        end
+        @headers = headers&.transform_values(&:to_s)
```
mikker, line 41: I think `rubocop-performance` suggests using real blocks instead of symbol-to-proc?

mikker (approved):

lib/elastic_apm/context/response.rb
estolfo: The only thing I could find was [this](https://docs.rubocop.org/rubocop-performance/cops_performance.html#performancemethodobjectasblock), which I don't believe is the same thing. RuboCop doesn't complain about this line.

PR #902, opened 2020-11-23
Handle both int and string log level, and extra central config values

### LINKED ISSUE #878: Add log_level spec
Implementing elastic/apm#332

Note that the implementation of this spec also includes adding the option to [Kibana](https://github.com/elastic/kibana/blob/master/x-pack/plugins/apm/common/agent_configuration/setting_definitions/general_settings.ts).

estolfo:
Resolves #878


mikker (approved):
LGTM

PR #905, opened 2020-11-26
Update the Rails.start API

### LINKED ISSUE #565: Change ElasticAPM::Rails.start arguments
⚠️**Breaking change**

Follow the pattern of `ElasticAPM::Grape.start` and take the Rails app as an argument to `ElasticAPM::Rails.start`.
That way, we can centralize [these lines](https://github.com/elastic/apm-agent-ruby/blob/master/lib/elastic_apm/config.rb#L222-L238) into that method.

It would look like:
`ElasticAPM::Rails.start(RailsApp, config)`

estolfo:
Update the `ElasticAPM::Rails.start` API to accept the app object so that the config manipulations for Rails can be done there and not in the `Config`.

Closes #565


lib/elastic_apm/config.rb
```diff
@@ -182,8 +182,6 @@ def app=(app)
       case app_type?(app)
       when :sinatra
         set_sinatra(app)
-      when :rails
-        set_rails(app)
```
estolfo, line 186: @mikker can you think of a reason why we wouldn't want to remove these Rails checks in the Config?

lib/elastic_apm/railtie.rb
```diff
@@ -28,14 +28,15 @@ class Railtie < ::Rails::Railtie
     end
 
     initializer 'elastic_apm.initialize' do |app|
-      config = Config.new(app.config.elastic_apm.merge(app: app)).tap do |c|
+      app.config.elastic_apm[:logger] ||= ::Rails.logger
```
estolfo, line 31: We need to do this otherwise, a logger object will be created and set on the `config` during the initialization. Then we won't know if we can override it with the `::Rails::logger` or not.

lib/elastic_apm/config.rb
mikker: Could we remove the Sinatra ones too then, if we want to move it to the _starters_?

docs/api.asciidoc
```diff
@@ -249,7 +249,7 @@ the gem and using the `Railtie`.
 
 [source,ruby]
 ----
-ElasticAPM::Rails.start(server_url: 'http://localhost:8200')
+ElasticAPM::Rails.start(MyRailsApp.new, server_url: 'http://localhost:8200')
```
mikker, line 252: Aren't there a bunch of side effects from calling `new` on an app?

mikker: I'd be a little bit concerned about `new`ing a rails app. Too much magic going on 😅

estolfo: If a user doesn't use a Railtie, then how do they initialize a Rails app? A user would pass their app instance to the `ElasticAPM::Rails.start` method and however they instantiate it is up to them. Do you know of a better example of instantiating a rails app without a Railtie?

lib/elastic_apm/config.rb
estolfo: Yes, let's do that 👍🏼

docs/api.asciidoc
mikker: I might be misunderstanding, but I don't think it's our job to start nor instantiate users' apps?

If someone wants to use this API, I think they should do so in an initializer or their own custom railtie. And when they do, they'll have the application object at hand. I don't think we ought to encourage any other way to start the agent with Rails.

estolfo: Yes, we are saying the same thing. I was just trying to provide a simple example. So would this be better? The `myapp` object is whatever application object they have in an initializer or custom railtie.

`ElasticAPM::Rails.start(myapp, server_url: 'http://localhost:8200')`

mikker: I think so. Or maybe include a whole example of starting in an initializer (including a note that this is the hard way)

estolfo: ok I added an example initializer

docs/api.asciidoc
```diff
@@ -244,12 +244,16 @@ Returns the built context.
 [[rails-start]]
 === Rails
 
-Start the agent and hook into Rails manually. This is useful if you skip requiring
-the gem and using the `Railtie`.
+Start the agent and hook into Rails manually. If you skip requiring the gem,
+you won't have the `Railtie` and would have to define your own initializer.
+Note that this is not the typical way to setup and start the agent in a Rails app.
```
mikker, line 249:
```suggestion
Start the agent and hook into Rails manually. If you skip requiring the gem,
you will need to start the agent yourself in for example an initializer.
Note that this is not the typical way to setup and start the agent in a Rails app.
```

spec/elastic_apm/rails_spec.rb
```diff
@@ -21,10 +21,22 @@
...
+
+    after(:all) do
+      Object.send(:remove_const, :RailsTestApp)
+    end
+
     describe '.start' do
       it 'starts the agent' do
         begin
-          ElasticAPM::Rails.start({})
+          ElasticAPM::Rails.start(RailsTestApp::Application.new, {})
```
mikker, line 39: Can we move this to an int. test perhaps? Starting Rails will infect the `Rails` namespace for the whole process, so for cleanliness' sake it'd perhaps be beneficial to run it in its own process like the other int. tests.

estolfo:
@mikker One consequence of removing [these lines](https://github.com/elastic/apm-agent-ruby/pull/905/files#diff-d8cae6e87d9863ebefec225c4db719eb38339e0d77227c5f900026046f8df979L171-L178) is that `service_name` won't default to `ruby` for apps that are neither Sinatra nor Rails. I don't know how often that case was encountered, as `app` is not a real config value and unlikely to be provided by users. What do you think?

mikker:
I think actually it's somewhat common to set the name yourself.

We could default the name to `"ruby"` in the `Config` and then override it in the `.start` methods. However that would break for users who _want_ to name the service `"ruby"`. So, perhaps we could default it to `"__none__"` or something else that's very unlikely to be a real name, then set it to `"ruby"` after all the initialization and configuration is settled?

mikker:
I might be answering something else than what you asked.

`app:` is very likely to be provided for Sinatra apps, as it's the recommended way to setup and start the agent with Sinatra.

estolfo:
Sorry I wasn't clear-- I wasn't asking about whether `service_name` is set by the user. I meant the scenario when `app: SomeAppOtherThanRailsOrSinatra` is set by the user. In that case, (the default case in the case statement), the `service_name` is set to `ruby`. This case we don't handle with the removal of the lines I linked above.

estolfo:
But we're essentially removing the `app` config value with this and #906 so that scenario would become irrelevant.

estolfo:
Closing this as it needs more discussion. There isn't much benefit to changing the `Rails.start` interface.

PR #906, opened 2020-11-26
Only support auto Sinatra app configuration via ElasticAPM::Sinatra.start

estolfo:
Use `ElasticAPM::Sinatra.start(MySinatraApp, ...)` instead of `ElasticAPM.start(app: MySintraApp, ...)`


mikker (commented):
If we were _reeaally_ nice to folks, we'd start suggesting this new api in the 3.x branch with an upcoming removal notice, then remove it in 4.0 first?

estolfo:
Sure, we should probably do it for rails also. I can open a PR against master for that.

estolfo:
In the meantime, can we merge this into `next` if you approve of this PR? That way, I can update #905 to include these changes.

mikker (approved):
Yes!

PR #908, opened 2020-11-30
Rename server_ca_cert to server_ca_cert_file

### LINKED ISSUE #650: Rename server_ca_cert config to server_ca_cert_file
The Ruby agent is currently the only one defining this config attribute, [server_ca_cert](https://www.elastic.co/guide/en/apm/agent/ruby/current/configuration.html#config-ssl-ca-cert):

> The path to a custom CA certificate for connecting to APM Server.

The agents have some other config which accept file paths, such as `log_file` and `config_file`. We should be consistent to avoid confusion, so I suggest we rename this one to `server_ca_cert_file`.

We could either deprecate the existing config name, and or keep it but repurpose it to accept the CA cert contents _contents_. For backwards compatibility we would do this if we detect the cert PEM header (`-----BEGIN`), and otherwise treat it as a file path.

estolfo:
Resolves #650

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -49,6 +49,11 @@ endif::[]
 
 - Support both integer and string log levels, and extra central config values {pull}902[#902]
 
+[float]
+===== Changed
+
+- Rename server_ca_cert to server_ca_cert_file {pull}908[#908]
+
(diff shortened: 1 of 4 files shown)


mikker (approved):
💯

PR #910, opened 2020-12-01
Warn that starting the agent for a Sintra app using app config arg is deprecated

### LINKED ISSUE #909: Warn users of the removal of the config app arg in favor of Rails.start and Sinatra.start
Suggest the `Rails|Sinatra.start` API in the 3.x branch with a notice of upcoming removal in 4.0.

Update: We only document starting a rails app via `ElasticAPM::Rails.start(...)` or via the Railtie, so we don't have to create a deprecation warning. ([ref](https://www.elastic.co/guide/en/apm/agent/ruby/current/api.html#rails-start))
For Sinatra, we do document `ElasticAPM.start(app: MySinatraApp)` ([ref](https://www.elastic.co/guide/en/apm/agent/ruby/current/configuration.html#_sinatra_and_rack)) so we have to warn about the deprecation.

estolfo:
Resolves #909

### DIFF
--- a/lib/elastic_apm/config.rb
+++ b/lib/elastic_apm/config.rb
@@ -344,6 +344,11 @@ module ElasticAPM
     end
 
     def set_sinatra(app)
+      warn '[DEPRECATED] Starting the agent for a Sinatra app via ' \
+           'ElasticAPM.start(app: MySinatraApp, ...) is deprecated and ' \
+           'will be removed in version 4.0. ' \
+           'Please use the ElasticAPM::Sinatra.start(MySinatraApp, ...) ' \
+           'API instead.'
       self.service_name = format_name(service_name || app.to_s)
       self.framework_name = framework_name || 'Sinatra'
       self.framework_version = framework_version || ::Sinatra::VERSION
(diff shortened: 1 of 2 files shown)


estolfo:
I think I should warn about using the `app` config arg instead.

estolfo:
Actually, I think this makes sense. If we warn anytime `app` is set on a config, it will create noise for users who use a Railtie and who don't intentionally use the `app` option.

mikker:
Remind me what the benefit is here? Could we keep on supporting the old way? I just think, if there's no clear benefit in the end for the user, we are just being a nuisance by making them have to change their app with no clear benefit to them (nor to us, in the end)?

estolfo:
I was hoping to consolidate the "special" behavior for each applications type into the application's module. For example, I was trying to avoid having [these lines](https://github.com/elastic/apm-agent-ruby/blob/master/lib/elastic_apm/config.rb#L356-L365) in the config and instead have them in the `Rails` module. It also seems cleaner to suggest only one way to start the agent for a given app type (Sinatra, Rails, Grape)

estolfo:
Closing this as we probably aren't going to deprecate the app in the config

PR #923, opened 2021-01-05
Test ruby 3.0 on Jenkins

estolfo:
Issues with running test suite on ruby 3.0:

- `google-protobuf` cannot be install on ruby >= 2.8. The message is `google-protobuf-3.14.0-universal-darwin requires ruby version < 2.8.dev, >= 2.3, which is incompatible with the current version, ruby 3.0.0p0`

- The json validation spec is failing because the `json-schema` gem needs to update their use of `Kernel#open` to be `URI.open` instead [here](https://github.com/ruby-json-schema/json-schema/blob/v2.8.1/lib/json-schema/schema/reader.rb#L121). Reference [here](https://rubyreferences.github.io/rubychanges/2.7.html#network-and-web). I've opened an issue with their repo [here](https://github.com/ruby-json-schema/json-schema/issues/449).

General changes:
- Test ruby 3.0
- Test rails 5.2
- Don't test rails 5.0 (only 5.1 and 5.2)
- Test rails 6.1 in addition to 6.0
- Test latest grape (1.5)
- Test master of frameworks on ruby 2.7 and 3.0
- Don't pass a url to `JSON::Validator.validate!`, pass the schema directly

After this is merged, open an issue as a reminder to:
- When a grape version supporting ruby 3.0 is released, update the jenkins matrices to include that version
- When a `json-schema` version is released supporting ruby 3.0, possibly change the `json_schema_matcher` to pass a url to `JSON::Valdiator.validate!` again


estolfo:
Replaced by https://github.com/elastic/apm-agent-ruby/pull/925

PR #924, opened 2021-01-12
CI: Update tested ruby and framework versions

mikker:
Also adds Rails 6.1


estolfo:
As long as we are also making changes to the tested rails versions, might as well:

- Don't test rails 5.0 or 5.1, only 5.2
- Test latest grape (1.5)

PR #934, opened 2021-01-14
Add transaction and span outcome cucumber specs

estolfo:


mikker:
This actually took a lot less stepdefs than I was expecting. Good job!

mikker:
Is this blocked btw?

estolfo:
Yes, we mentioned it briefly at the agents meeting last week-- Sylvain will look over the the PR [here](https://github.com/elastic/apm/pull/402) and make some suggestions. So I'm waiting until that PR is settled, as this one may change as a result.

estolfo:
Closing in favor of #966

PR #935, opened 2021-01-15
Remove legacy SQL parser

mikker:
New one has been default for a while and there haven't been any complaints.


lib/elastic_apm/spies/sequel.rb
```diff
@@ -28,7 +28,7 @@ class SequelSpy
       ACTION = 'query'
 
       def self.summarizer
-        @summarizer ||= Sql.summarizer
+        @summarizer ||= Sql::Signature::Summarizer.new
```
estolfo, line 31: Do we even need this method now? We could just get rid of the `Sql.summarizer` method.

mikker: The memoization method? I guess not necessarily.

PR #939, opened 2021-01-26
fix: rails master has been renamed to main

mdelapenya:
## What does this pull request do?
It renames the `rails-master` variable to `rails-main`

## Why is it important?
The CI is broken for both rails 3.0 and 2.7, which are not excluded in the CI support matrix.

## Checklist

- [x] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/). 
- [ ] My code follows the style guidelines of this project (See `.rubocop.yml`)
- [x] I have rebased my changes on top of the latest master branch

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-ruby/blob/master/CONTRIBUTING.md#testing) pass locally with my changes

- [ ] I have made corresponding changes to the documentation
- [ ] I have updated [CHANGELOG.asciidoc](CHANGELOG.asciidoc)
- [ ] I have updated [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)
- [ ] Added an API method or config option? Document in which version this will be introduced

PR #940, opened 2021-01-26
fix: rails master has been renamed to main (cont.)

mikker:
Continued from #939

### DIFF
--- a/.ci/.jenkins_exclude.yml
+++ b/.ci/.jenkins_exclude.yml
@@ -14,7 +14,7 @@ exclude:
   - RUBY_VERSION: ruby:2.3
     FRAMEWORK: grape-1.5,sinatra-2.0,rails-6.0
   - RUBY_VERSION: ruby:2.3
-    FRAMEWORK: rails-master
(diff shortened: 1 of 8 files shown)


mdelapenya:
Oh, I sent a commit on top of mine, but it seems this one resolves it better. Will close my PR, thanks!

mdelapenya (approved):

mikker:
This fails while building a native package on some jruby images. Do they have `gcc` or similar available? `build-essential` or what have you?

mdelapenya:
> This fails while building a native package on some jruby images. Do they have `gcc` or similar available? `build-essential` or what have you?

The jruby images are built within this repository, in https://github.com/elastic/apm-agent-ruby/tree/master/.ci/docker/jruby, but I do not see any reference to `gcc` or `build-essential`. Do you think that updating them with your needs would fix this build problem? If it fixes it, we should push them again to the container registry.

mikker:
I don't know honestly, I just saw that building the native extension failed. I'll have to try it out locally.

PR #953, opened 2021-02-03
Fix missing ActiveRecord connection object in notification payload

taylorthurlow:

ActiveRecord's notification subscription system is leveraged in order to instrument SQL queries. This notification system passes a `payload` hash containing arbitrary notification information about the event that has occurred. `apm-agent-ruby` parses this notification payload and determines the adapter name by looking for the connection object stored in `payload[:connection]`.

This value is _not_ set in any ActiveRecord version before 6.0. In https://github.com/rails/rails/commit/6d1c9a9b23579d13b7cbdd3217ed2cfcc300e06e, it can be seen that the `connection` key is added to the payload, and this change is included in ActiveRecord 6.0 and newer. Earlier versions of ActiveRecord include a `:connection_id` payload key/value pair which includes the ActiveRecord connection object's `#object_id`.

This pull request adds functionality which loads the connection object in the event that `:connection_id` has a value, and `:connection` does not. Because we're loading the object from an `object_id`, proper error handling is included in the event that the object referred to by the `:connection_id` value has been garbage collected (`RangeError` will be raised). The nature of fetching an object by an `object_id` in this way is that we cannot guarantee that the same ID will still match to the object we expect (though we can be reasonably sure). To give ourselves a little more assurance I make sure the resulting object responds to `:adapter_name` as we expect.

It should also be noted that the `:connection_id` key/value pair was removed from the notification payload in https://github.com/rails/rails/commit/d140ab073f296a2a689a4dbd58cf9434c6b2a484#diff-460f4e7973c5dd945c51d24df5b0173961190d3645f4e2585fd3003fa1fc0ff7 (included in ActiveRecord releases 6.1.0 and later), and the included code change is robust enough to handle this key being absent.

## Why is it important?

Currently, if the `payload[:connection]` value is missing (which it will be in AR < 6.0), the fallback adapter value is the result of calling `::ActiveRecord::Base.connection_config[:adapter]`. This is fine if the connection in question uses the base `ActiveRecord::Base` class, but in the apps I'm working on, that isn't a safe assumption. Multiple database connections use multiple base classes which come from various Ruby gems which define their own base connection classes that inherit from `ActiveRecord::Base`.

This change makes things a little more flexible, and we can trust that ActiveRecord will provide us with the correct `object_id` reference to the connection.

One side-effect of this change would be that the aforementioned fallback adapter code path will probably never be reached anymore - The likelihood of the payload missing a value for both `:connection` and `:connection_id` is slim to none. That said, I don't necessarily see the harm in removing it.

## Checklist

- [X] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).
- [X] My code follows the style guidelines of this project (See `.rubocop.yml`)
- [X] I have rebased my changes on top of the latest master branch
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing [**unit** tests](https://github.com/elastic/apm-agent-ruby/blob/master/CONTRIBUTING.md#testing) pass locally with my changes
- [X] I have updated [CHANGELOG.asciidoc](CHANGELOG.asciidoc)

## Related issues

### DIFF
--- a/lib/elastic_apm/normalizers/rails/active_record.rb
+++ b/lib/elastic_apm/normalizers/rails/active_record.rb
@@ -57,22 +57,17 @@ module ElasticAPM
         private
 
         def subtype_for(payload)
-          if payload[:connection_id]
-            payload[:connection] ||= begin
-              loaded_object = ObjectSpace._id2ref(payload[:connection_id])
+          return cached_adapter_name(payload[:connection].adapter_name) if payload[:connection]
 
-              if loaded_object.respond_to?(:adapter_name)
-                loaded_object
-              end
+          if payload[:connection_id]
+            begin
+              loaded_object = ObjectSpace._id2ref(payload[:connection_id])
+              return cached_adapter_name(loaded_object.adapter_name) if loaded_object.respond_to?(:adapter_name)
             rescue RangeError # if connection object has somehow been garbage collected
-              nil
             end
           end
 
-          cached_adapter_name(
-            payload[:connection]&.adapter_name ||
-              ::ActiveRecord::Base.connection_config[:adapter]
-          )
+          cached_adapter_name(nil)
         end
 
         def summarize(sql)
--- a/lib/elastic_apm/normalizers/rails/active_record.rb
+++ b/lib/elastic_apm/normalizers/rails/active_record.rb
@@ -59,7 +59,7 @@ module ElasticAPM
         def subtype_for(payload)
           return cached_adapter_name(payload[:connection].adapter_name) if payload[:connection]
 
-          if payload[:connection_id]
+          if can_attempt_connection_id_lookup?(payload)
             begin
               loaded_object = ObjectSpace._id2ref(payload[:connection_id])
               return cached_adapter_name(loaded_object.adapter_name) if loaded_object.respond_to?(:adapter_name)
@@ -67,7 +67,7 @@ module ElasticAPM
             end
           end
 
-          cached_adapter_name(nil)
+          cached_adapter_name(::ActiveRecord::Base.connection_config[:adapter])
         end
 
         def summarize(sql)
@@ -76,10 +76,15 @@ module ElasticAPM
 
         def cached_adapter_name(adapter_name)
           return UNKNOWN if adapter_name.nil? || adapter_name.empty?
+
           @adapters[adapter_name] ||
             (@adapters[adapter_name] = adapter_name.downcase)
         rescue StandardError
           nil
+        end
+
+        def can_attempt_connection_id_lookup?(payload)
+          RUBY_ENGINE == "ruby" && payload[:connection_id] && ObjectSpace.respond_to?(:_id2ref)
         end
       end
     end
--- a/lib/elastic_apm/normalizers/rails/active_record.rb
+++ b/lib/elastic_apm/normalizers/rails/active_record.rb
@@ -57,6 +57,18 @@ module ElasticAPM
         private
 
         def subtype_for(payload)
+          if payload[:connection_id]
+            payload[:connection] ||= begin
+              loaded_object = ObjectSpace._id2ref(payload[:connection_id])
+
+              if loaded_object.respond_to?(:adapter_name)
+                loaded_object
+              end
+            rescue RangeError # if connection object has somehow been garbage collected
+              nil
+            end
+          end
+
           cached_adapter_name(
             payload[:connection]&.adapter_name ||
               ::ActiveRecord::Base.connection_config[:adapter]
--- a/spec/elastic_apm/normalizers/active_record_spec.rb
+++ b/spec/elastic_apm/normalizers/active_record_spec.rb
@@ -83,12 +83,6 @@ module ElasticAPM
               connection: double(adapter_name: 'MySQL')
             )
             expect(subtype).to eq 'mysql'
-
-            _name, _type, subtype, = normalize_payload(
-              sql: sql,
-              connection: double(adapter_name: 'Postgres')
-            )
-            expect(subtype).to eq 'postgres'
           end
           
           it 'resolves the connection_id object id to a connection if the full connection is missing' do
@@ -100,12 +94,6 @@ module ElasticAPM
               connection_id: double(adapter_name: 'MySQL').object_id
             )
             expect(subtype).to eq 'mysql'
-
-            _name, _type, subtype, = normalize_payload(
-              sql: sql,
-              connection_id: double(adapter_name: 'Postgres').object_id
-            )
-            expect(subtype).to eq 'postgres'
           end
           
           it 'handles a failure to load the connection object by connection_id' do
@@ -134,13 +122,6 @@ module ElasticAPM
               connection_id: double(adapter_name: 'wrong_db').object_id
             )
             expect(subtype).to eq 'mysql'
-
-            _name, _type, subtype, = normalize_payload(
-              sql: sql,
-              connection: double(adapter_name: 'Postgres'),
-              connection_id: double(adapter_name: 'wrong_db').object_id
-            )
-            expect(subtype).to eq 'postgres'
           end
 
           it 'uses ActiveRecord::Base when payload is not available' do
--- a/spec/elastic_apm/normalizers/active_record_spec.rb
+++ b/spec/elastic_apm/normalizers/active_record_spec.rb
@@ -32,26 +32,6 @@ module ElasticAPM
           expect(subject).to be_a SqlNormalizer
         end
 
-        describe '#initialize' do
-          unless defined? ::ActiveRecord::Base
-            class ::ActiveRecord
-              class Base; end
-            end
-          end
-
-          it 'knows the AR adapter' do
-            allow(::ActiveRecord::Base)
-              .to receive(:connection_config) { { adapter: 'MySQL' } }
-
-            subject = SqlNormalizer.new nil
-
-            _, _, subtype, =
-              subject.normalize(nil, nil, sql: 'DROP * FROM users')
-
-            expect(subtype).to eq 'mysql'
-          end
-        end
-
         describe '#normalize' do
           let(:key) { 'sql.active_record' }
           subject { SqlNormalizer.new nil }
@@ -61,12 +41,13 @@ module ElasticAPM
           end
 
           it 'normalizes queries' do
-            allow(::ActiveRecord::Base)
-              .to receive(:connection_config) { { adapter: nil } }
             sql = 'SELECT  "hotdogs".* FROM "hotdogs" ' \
               'WHERE "hotdogs"."topping" = $1 LIMIT 1'
 
-            name, type, subtype, action, context_ = normalize_payload(sql: sql)
+            name, type, subtype, action, context_ = normalize_payload(
+              sql: sql,
+              connection: double(adapter_name: '')
+            )
             expect(name).to eq 'SELECT FROM hotdogs'
             expect(type).to eq 'db'
             expect(subtype).to eq 'unknown'
@@ -84,7 +65,7 @@ module ElasticAPM
             )
             expect(subtype).to eq 'mysql'
           end
-          
+
           it 'resolves the connection_id object id to a connection if the full connection is missing' do
             sql = 'SELECT  "burgers".* FROM "burgers" ' \
               'WHERE "burgers"."cheese" = $1 LIMIT 1'
@@ -94,22 +75,6 @@ module ElasticAPM
               connection_id: double(adapter_name: 'MySQL').object_id
             )
             expect(subtype).to eq 'mysql'
-          end
-          
-          it 'handles a failure to load the connection object by connection_id' do
-            sql = 'SELECT  "burgers".* FROM "burgers" ' \
-              'WHERE "burgers"."cheese" = $1 LIMIT 1'
-
-            allow(::ActiveRecord::Base)
-              .to receive(:connection_config) { { adapter: 'Postgres' } }
-
-            _name, _type, subtype, = normalize_payload(
-              sql: sql,
-              connection_id: 999_999_999 # invalid object_id
-            )
-
-            expect(subtype).to eq 'postgres'
-            expect(::ActiveRecord::Base).to have_received(:connection_config)
           end
 
           it 'uses the connection from payload even if the connection_id is available' do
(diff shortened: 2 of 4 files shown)


spec/elastic_apm/normalizers/active_record_spec.rb
```diff
@@ -90,6 +90,58 @@ def normalize_payload(payload)
...
+          end
+
+          it 'uses the connection from payload even if the connection_id is available' do
+            sql = 'SELECT  "burgers".* FROM "burgers" ' \
+              'WHERE "burgers"."cheese" = $1 LIMIT 1'
+
+            _name, _type, subtype, = normalize_payload(
+              sql: sql,
+              connection: double(adapter_name: 'MySQL'),
+              connection_id: 'wrong object'.object_id
```
estolfo, line 134: It might make more sense for this to be `connection_id: double(adapter_name: 'wrong_db').object_id`, to be sure that we could successfully get an adapter name from the `connection_id:` field, if there were a bug in the code.

spec/elastic_apm/normalizers/active_record_spec.rb
```diff
@@ -90,6 +90,58 @@ def normalize_payload(payload)
...
+              sql: sql,
+              connection_id: double(adapter_name: 'MySQL').object_id
+            )
+            expect(subtype).to eq 'mysql'
+
+            _name, _type, subtype, = normalize_payload(
+              sql: sql,
+              connection_id: double(adapter_name: 'Postgres').object_id
+            )
+            expect(subtype).to eq 'postgres'
```
estolfo, line 108: What is the reason for testing this twice with a different `adapter_name`?

estolfo (changes requested):
Thank you so much for the thorough explanation and fantastic code changes + tests! I had a few suggestions/comments.

lib/elastic_apm/normalizers/rails/active_record.rb
```diff
@@ -57,6 +57,18 @@ def normalize(_transaction, _name, payload)
         private
 
         def subtype_for(payload)
+          if payload[:connection_id]
+            payload[:connection] ||= begin
```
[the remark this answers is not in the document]
taylorthurlow: I went back and forth on this idea. It seems like it actually is our own object, we subscribed to notifications from ActiveRecord and they sent us a copy. If I receive a magazine in the mail every month because I asked for it, it's my magazine. We read it and throw it away, don't we?

Maybe my analogy doesn't fit, but it doesn't make any difference to me in the end. I think I went this route because it required less code overall. I'll refactor this method to skip the mutation. I also assume, based on your suggestion, that you think we should remove the `::ActiveRecord::Base.connection_config[:adapter]` check? I mentioned that it was pretty much redundant, as I don't expect there to be a case where that code is reached anyways.

spec/elastic_apm/normalizers/active_record_spec.rb
taylorthurlow: Seems like a good change to me, thanks.

spec/elastic_apm/normalizers/active_record_spec.rb
taylorthurlow: Pretty much only because the comparable pre-existing test for using the `:connection` value checked both. The only assumption I might be able to make about the original tester's intent was to check that the adapter name gets downcased properly?

See this existing test on `master`: https://github.com/elastic/apm-agent-ruby/blob/master/spec/elastic_apm/normalizers/active_record_spec.rb#L77-L92

taylorthurlow:
Thanks for reviewing so quickly! Left my feedback where necessary, I'll make some of the changes now.

taylorthurlow:
Ok so I've pushed up another commit which addresses the payload mutation discussion. I removed the code path which uses `::ActiveRecord::Base` as a fallback, and the tests associated with it. I believe I found a problem with the test suite, though - I removed this bit of code now that the system under test doesn't refer to the class anymore:

https://github.com/elastic/apm-agent-ruby/blob/9e49beb4d41538abd25b8b00babf9cbb19989224/spec/elastic_apm/normalizers/active_record_spec.rb#L36-L40

This had the unintended impact of breaking the spec located at `spec/elastic_apm/spies/shoryuken_spec.rb`:

```
$> be rspec
[INFO] Skipping Sinatra.start spec
Run options: exclude {:type=>"json_schema"}

Randomized with seed 54416
...........................................................................................................................................................FF.......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................*..........................................................................................................

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) ElasticAPM::Metrics thread safety stress test handles multiple threads reporting and collecting at the same time
     # Temporarily skipped with xcontext
     # ./spec/elastic_apm/metrics_spec.rb:118


Failures:

  1) Spy: Shoryuken reports errors
     Failure/Error:
       allow(::ActiveRecord::Base)
         .to receive(:clear_active_connections!)

     NameError:
       uninitialized constant ActiveRecord
     # ./spec/elastic_apm/spies/shoryuken_spec.rb:68:in `block (2 levels) in <module:ElasticAPM>'
     # ./vendor/ruby/2.7.0/gems/webmock-3.11.2/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

  2) Spy: Shoryuken instruments jobs
     Failure/Error:
       allow(::ActiveRecord::Base)
         .to receive(:clear_active_connections!)

     NameError:
       uninitialized constant ActiveRecord
     # ./spec/elastic_apm/spies/shoryuken_spec.rb:68:in `block (2 levels) in <module:ElasticAPM>'
     # ./vendor/ruby/2.7.0/gems/webmock-3.11.2/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

Finished in 11.84 seconds (files took 1.75 seconds to load)
767 examples, 2 failures, 1 pending

Failed examples:

rspec ./spec/elastic_apm/spies/shoryuken_spec.rb:106 # Spy: Shoryuken reports errors
rspec ./spec/elastic_apm/spies/shoryuken_spec.rb:89 # Spy: Shoryuken instruments jobs

Randomized with seed 54416
```

The tests do run in random order so I imagine this could be an intermittent issue. I added the snippet above into this newly-failing test and everything is passing again.

Regarding the changes that I made to avoid the payload mutation, let me know if you think it's an appropriate solution. I think I arrived at an implementation that is reasonably easy to grep.

spec/elastic_apm/normalizers/active_record_spec.rb
estolfo: Thanks for the explanation. I don't know why exactly we have more than one check for downcasing the adapter name. I guess we can just skip doing a second check in your tests.

estolfo:
Sorry about that `ActiveRecord::Base` trouble. I looked into it and it seems that we were defining an `ActiveRecord` class and `ActiveRecord::Base` class outside of a `before` block in the `active_record_spec`, which meant that the classes were defined when the file was loaded. Regardless of what order the specs ran in, the classes would have been defined.

Anyway, what you've done-- putting the definition in a `before` block of the Shoryuken spec is a good solution so those classes are defined when the spec is run.

lib/elastic_apm/normalizers/rails/active_record.rb
```diff
@@ -57,10 +57,17 @@ def normalize(_transaction, _name, payload)
...
-            payload[:connection]&.adapter_name ||
-              ::ActiveRecord::Base.connection_config[:adapter]
-          )
+          return cached_adapter_name(payload[:connection].adapter_name) if payload[:connection]
+
+          if payload[:connection_id]
+            begin
+              loaded_object = ObjectSpace._id2ref(payload[:connection_id])
+              return cached_adapter_name(loaded_object.adapter_name) if loaded_object.respond_to?(:adapter_name)
+            rescue RangeError # if connection object has somehow been garbage collected
```
estolfo, line 66: We return `unknown` if we run into any trouble getting the connection object or its adapter name. So I think we should also return `unknown` in the case we can't get the connection object because it has been garbage collected.

estolfo: I think you also removed a test for this code path. Was that intentional?
https://github.com/elastic/apm-agent-ruby/pull/953/commits/c776010ff157e203cf5c128055366e9f8c43c525#diff-54b1044cee5cfa350d537561412b450f0858a900078239714e4f503306491251L108

estolfo (changes requested):

lib/elastic_apm/normalizers/rails/active_record.rb
taylorthurlow: > We return `unknown` if we run into any trouble getting the connection object or its adapter name. So I think we should also return `unknown` in the case we can't get the connection object because it has been garbage collected.

Unless I misunderstand the code path you're referring to, I believe this does what you want. the final non-early-return case in `#subtype_for` calls `cached_adapter_name(nil)` which will return the adapter name in the `UNKNOWN` constant. The only cases in which we won't reach this final return value are:

- If we have a `payload[:connection]` value and we grab it's adapter name
- If we have a `payload[:connection_id]` value, it loads an object correctly, _and_ that object responds to `:adapter_name`

Otherwise that final code path should be reached and we'll get `unknown` as the adapter name.

> I think you also removed a test for this code path. Was that intentional?
> [c776010#diff-54b1044cee5cfa350d537561412b450f0858a900078239714e4f503306491251L108](https://github.com/elastic/apm-agent-ruby/commit/c776010ff157e203cf5c128055366e9f8c43c525#diff-54b1044cee5cfa350d537561412b450f0858a900078239714e4f503306491251L108)

This one is sort of an oversight on my part. I'm still going to "remove" this test but I'll be replacing it with one or two that properly test the fallback to the `unknown` adapter.

taylorthurlow: Ok I pushed up the new tests. I don't see a way to effectively test the code path where the object_id lookup raises the `RangeError` due to the object having been garbage collected.

We'd somehow have to manually garbage collect an object, which, even if I knew how to do, I don't think would be a good idea. Do you have any suggestions, or are you ok with that code path remaining untested?

estolfo:
jenkins, please run the tests

estolfo (commented):
Thanks a lot for addressing my comments!

I think there might be an issue with using `ObjectSpace._id2ref` and I apologize for not realizing this earlier.
I was reading about the method and how it had changed in ruby 2.7 and I found [this issue](https://bugs.ruby-lang.org/issues/15711) in the Ruby tracker.
It says that the method will be deprecated soon. So this is a good solution now but unfortunately it’s not a long-term one.

Secondly, I read in Charles Nutter’s comments that JRuby only supports `_id2ref` with a command line flag. We officially support JRuby and we unfortunately can’t require our users to use a particular command line flag for the agent to work properly. I ran the tests on CI, and as expected, the test fails on JRuby. The method raises a RangeError if the command line flags are not used.
Here is the behavior in irb:

```
jruby-9.2.8.0 :002 > "my string".object_id
 => 2002 
jruby-9.2.8.0 :003 > ObjectSpace._id2ref(2002)
(irb):3: warning: ObjectSpace is disabled; _id2ref only supports immediates, pass -X+O to enable
Traceback (most recent call last):
        7: from /Users/emily/.rvm/rubies/jruby-9.2.8.0/bin/irb:13:in `<main>'
        6: from org/jruby/RubyKernel.java:1193:in `catch'
        5: from org/jruby/RubyKernel.java:1193:in `catch'
        4: from org/jruby/RubyKernel.java:1425:in `loop'
        3: from org/jruby/RubyKernel.java:1061:in `eval'
        2: from (irb):3:in `evaluate'
        1: from org/jruby/RubyObjectSpace.java:114:in `_id2ref'
RangeError (0x00000000000007d2 is not id value)
```


Considering `_id2ref` will eventually be deprecated and that we can’t require JRuby users to pass certain command line args in order for this code to work, it doesn’t seem to make sense to use this method. I’m open to ideas though— you could write the code such that it fails back to our previous behavior using `::ActiveRecord::Base.connection_config[:adapter]` for JRuby or when a RangeError is raised.

And finally, regarding the test for a RangeError, you can use `_id2ref` with 4 and it’ll always throw a RangeError.

Let me know what you think!

estolfo:
Hi @taylorthurlow, do you have any thoughts on my last comment?

taylorthurlow:
Hey, sorry for the delay.

> I think there might be an issue with using `ObjectSpace._id2ref` and I apologize for not realizing this earlier. I was reading about the method and how it had changed in ruby 2.7 and I found [this issue](https://bugs.ruby-lang.org/issues/15711) in the Ruby tracker. It says that the method will be deprecated soon. So this is a good solution now but unfortunately it’s not a long-term one.

Interesting discussion on this issue, I wasn't aware of the shortcomings of the method.

> Secondly, I read in Charles Nutter’s comments that JRuby only supports `_id2ref` with a command line flag. We officially support JRuby and we unfortunately can’t require our users to use a particular command line flag for the agent to work properly. I ran the tests on CI, and as expected, the test fails on JRuby. The method raises a RangeError if the command line flags are not used.

I hadn't thought about this, so good thing you brought it up!

> regarding the test for a RangeError, you can use `_id2ref` with 4 and it’ll always throw a RangeError.

Nice tip! Handy.

> Considering `_id2ref` will eventually be deprecated and that we can’t require JRuby users to pass certain command line args in order for this code to work, it doesn’t seem to make sense to use this method. I’m open to ideas though— you could write the code such that it fails back to our previous behavior using `::ActiveRecord::Base.connection_config[:adapter]` for JRuby or when a RangeError is raised.

So given this new information I see two ways forward.

To fix the issue I could do a check that we are using MRI and only attempt the use of `id2ref` if the method is defined. I could also optionally add a check for an ActiveRecord version `< 6`, while not strictly necessary, it would probably avoid potential deprecation warning messages in the future. As you mentioned the `ActiveRecord::Base` fallback would be added back, seeing as we can't assume this new functionality works in all cases.

Alternatively, maybe the best choice is to just continue with the subpar details in older ActiveRecord versions.

I'm OK either way, so let me know if you think the extra checks are worth it, and I can fix things up. Thanks for checking in again!

estolfo:
Hi @taylorthurlow, thanks for your response. Most of our Rails users are on MRI so I think that would be a reasonable check. Then, if `_id2ref` is not available, we can fall back to using `ActiveRecord::Base`. I don't think it's necessary to check the `ActiveRecord` version as the deprecation doesn't have a target date so it's probably not worth worrying about the deprecation warning for now. Thank you for your effort on this!

taylorthurlow:
jenkins, please run the tests

taylorthurlow:
Sorry for the delay @estolfo. I just pushed up changes which should achieve what we were looking for. I added a check for the current Ruby interpreter and added the `::ActiveRecord::Base` check back in. In order to properly test this stuff I added a little skip clause to the relevant tests which should mark the tests as pending/skipped for non-MRI environments. The skip message explains that they are MRI-only. Let me know if that solution is adequate or if you'd prefer a solution that pretends those tests don't exist rather than marking them as skipped.

estolfo:
jenkins, please run the tests

estolfo:
jenkins, please run the tests

estolfo:
Thank you so much for your work on this, @taylorthurlow!

taylorthurlow:
No problem, glad to contribute. Thanks for your detailed feedback.

PR #959, opened 2021-02-09
Remove support for legacy Traceparent args

estolfo:


lib/elastic_apm/opentracing.rb
```diff
@@ -129,8 +129,9 @@ def initialize(trace_context:, baggage: nil)
       def self.from_header(header)
         return unless header
 
-        trace_context = ElasticAPM::TraceContext.parse(header)
-        return unless trace_context
+        trace_context = ElasticAPM::TraceContext.new(
+          traceparent: TraceContext::Traceparent.parse(header)
```
mikker, line 133:
```suggestion
        trace_context = TraceContext.new(
          traceparent: TraceContext::Traceparent.parse(header)
```

I found it a bit confusing to leave out the topmost module in one and not the other.

mikker (approved):
Nice!

PR #960, opened 2021-02-10
skips spans for http calls missing an authority segment when the call is made using Net::HTTP or Faraday

CarlosRoque:
It allows for requests missing the authority segment to be gracefully handled by the spies. This fix is for net_http and Faraday. The fix forwards the call without AMP spans when the authority is missing.

## Why is it important?
There are some situations where a gem could use Faraday incorrectly by not providing domain name in their calls. This in turn, causes APM to fail with an error that the offending gem is not expecting. The unexpected error can cause the gems to completely fail without the user being able fix the issue by means of a workaround. The only solution would be to remove APM.

## Example
The following explanation is for the RestForce gem from SalesForce. 
The gem performs calls using Faraday, it automatically configures urls and paths.
When making a call for the first time, the gem makes a call without a configured host.
Without a valid host APM fails with an invalid url from the URI:HTTP class. 
It seems RestForce does not  handle this error correctly and halts any future chain of calls. 
There is nothing the user of this gem can do to fix the issue other than to remove APM.

## Checklist
- [x] New and existing [**unit** tests](https://github.com/elastic/apm-agent-ruby/blob/master/CONTRIBUTING.md#testing) pass locally with my changes
- [ ] ~~I have updated [CHANGELOG.asciidoc](CHANGELOG.asciidoc)~~

### DIFF
--- a/lib/elastic_apm/spies/faraday.rb
+++ b/lib/elastic_apm/spies/faraday.rb
@@ -55,6 +55,8 @@ module ElasticAPM
               tmp_request = build_request(method) do |req|
                 yield(req) if block_given?
               end
+
+              return run_request_without_apm(method, url, body, headers, &block) if tmp_request.path.nil?
               uri = URI(tmp_request.path)
             end
 
--- a/lib/elastic_apm/spies/net_http.rb
+++ b/lib/elastic_apm/spies/net_http.rb
@@ -65,6 +65,10 @@ module ElasticAPM
             method = req.method.to_s.upcase
             path, query = req.path.split('?')
 
+            if host.nil?
+              return request_without_apm(req, body, &block)
+            end
+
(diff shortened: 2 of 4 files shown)


mikker:
Hi @CarlosRoque! Thank you for reporting and contributing this. I think we might want to add something more in the line of a catch-all `begin/rescue` here, as there are potentially many ways, libraries could be _"mis-using"_ http?

CarlosRoque:
@mikker thanks for the reply. I have reviewed your proposed change, and that works fine for my particular issue. 
My reasoning behind my changes was not to create a span but I had alternatively considered just doing this in the faraday spy
`uri = URI(tmp_request.path.to_s)` either way I would be more than happy if any of the 3 possible solutions is implemented. right now I am not able to use APM on 3 different projects dues to this issue. Thanks again.

PR #962, opened 2021-02-11
Default to localhost when Host missing

mikker:
Some HTTP clients allow requests without specifying a `Host` to request them at.

This [must be sent with every request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host) but we'll just assume that the request is made at `localhost`.

Alternative solution to #960 that still captures spans. Thank you @CarlosRoque for finding and reporting this issue.

Closes #960

### DIFF
--- a/docs/supported-technologies.asciidoc
+++ b/docs/supported-technologies.asciidoc
@@ -77,8 +77,7 @@ requests using these libraries:
 - Http.rb (v0.6+)
 - Faraday (v0.2.1+)
 
-*Note:* Some libraries like Net/HTTP and Faraday can make requests without specifying a Host.
-When no Host is specified they, and the agent, will fall back to `localhost` as the Host.
+*Note:* These libraries usually assume `localhost` if no `Host` is specified, so the agent does as well. 
 
 [float]
 [[supported-technologies-backgroud-processing]]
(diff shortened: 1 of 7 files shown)


CarlosRoque:
@mikker thank you, I tested this change on my app and it fixed my [issue #960](https://github.com/elastic/apm-agent-ruby/pull/960)

mikker:
We decided to go with this with an added note about this behaviour in the docs.

CarlosRoque:
@mikker thanks!

docs/supported-technologies.asciidoc
```diff
@@ -77,6 +77,9 @@ requests using these libraries:
 - Http.rb (v0.6+)
 - Faraday (v0.2.1+)
 
+*Note:* Some libraries like Net/HTTP and Faraday can make requests without specifying a Host.
+When no Host is specified they, and the agent, will fall back to `localhost` as the Host.
+
```
mikker, line 82:
```suggestion
*Note:* These libraries usually assume `localhost` if no `Host` is specified, so the agent does as well. 

```

estolfo, line 82: I think you missed applying the suggested change.

estolfo (approved):

PR #971, opened 2021-02-19
Update outcome cucumber spec

mikker:
Updating #966 with the most recent update #969

Closes #966


estolfo:
I think this might need to be rebased, right? The default to localhost commit is in this PR.

mikker:
No idea why this is listing that `localhost` PR? It's been merged and this has been rebased?

PR #975, opened 2021-02-25
Set current_transaction when starting span in new thread and use current_span transaction in ErrorBuilder

### LINKED ISSUE #973: Include transaction details for errors which are reported during an async span
## Is your feature request related to a problem? Please describe.
Context: we are running a Rails application using ElasticAPM (see versions below)

We spawn multiple threads during an individual request into our system to do various units of work (mostly making asynchronous HTTP requests). For the execution of those threads we wrap the block in `ElasticAPM.with_span` and set the `parent` to be the `current_transaction` from the main thread. For the most part, this seems to work as expected in that we see the single transaction and the async spans show up in APM in Kibana. Where it seems to fall down is if we report an error during the execution of one of those threads. While `ElasticAPM.current_span` is correct and references the span we created by calling `with_span`, `ElasticAPM.current_transaction` is `nil`. When `ElasticAPM.current_transaction` is `nil` the `report` method will not add transactional context. Custom context does not work either. 

My inclination is that given how async spans work (they have a reference to their `parent` as well as the overarching `transaction`) and that `ElasticAPM.current_transaction` will be `nil` on those threads (as this is stored in thread local storage in `Instrumenter` and so a new thread does not have one) that it would be fair for the `ErrorBuilder` to attempt to pull a transaction from first `ElasticAPM.current_transaction` (as it does today) and then from `ElasticAPM.current_span&.transaction`. It feels like this aligns with how async spans should be working and would allow error reporting from threads which have started an async span to have context, but I am most likely missing something!

## Describe the solution you'd like
Either via the mechanism described above or some other way, we would like the ability to report errors on threads that can include transactional and custom context.

## Describe alternatives you've considered
We attempted to set the `current_transaction` at the beginning of our Thread (before calling `with_span`) by digging into the `instrumenter`:
```
ElasticAPM.agent.instrumenter.current_transaction = ElasticAPM.current_transaction
```
however this caused all sorts of problems.

Our current alternative would be to not have the errors correlate back to the transaction, though this feels as if we are not getting the most out of our ELK/APM stack.

## Additional context
rails 5.2.4.5
elastic-apm 3.5.0 (also tried on 3.13.0)

estolfo:
These changes set `ElasticAPM.current_transaction` when a new span in created in a new thread using a span's transaction attribute. It also uses `current_span`'s transaction in the `ErrorBuilder`.

I've made the following changes:

- The `current` variable on an instrumenter is itself now set as a Thread current variable, instead of the `spans` and `transaction` variables. This way, a newly spawned thread will have its own `current` variable to track its current spans and a current transaction, which may be different from the parent thread's.
- The error builder uses either the current transaction or the span's transaction object to build the error.
- Set the current transaction to the transaction attribute of newly created span in `instrumenter#start_span`.

Outstanding questions
- When a span is created in a new thread, should the `transaction` of the parent object be set as the new thread's `current_transaction`? For example:

```ruby
transaction = ElasticAPM.start_transaction

Thread.new do
  ElasticAPM.with_span('job 1', parent: transaction) do |span|
    # Does ElasticAPM.current_transaction here equal `transaction` above or is it nil?
  end
end

# or when a span is the parent

span = ElasticAPM.start_span
Thread.new do
  ElasticAPM.with_span('job 2', parent: span) do |span|
    # Does ElasticAPM.current_transaction here equal the `transaction` above or is it nil?
  end
end
```
Closes #973

### DIFF
--- a/lib/elastic_apm/error_builder.rb
+++ b/lib/elastic_apm/error_builder.rb
@@ -35,7 +35,9 @@ module ElasticAPM
         add_stacktrace error, :exception, exception.backtrace
       end
 
-      add_current_transaction_fields error, ElasticAPM.current_transaction
+      transaction = ElasticAPM.current_transaction ||
+        ElasticAPM.current_span&.transaction
+      add_current_transaction_fields error, transaction
 
       error
     end
--- a/lib/elastic_apm/instrumenter.rb
+++ b/lib/elastic_apm/instrumenter.rb
@@ -34,26 +34,11 @@ module ElasticAPM
     # @api private
     class Current
       def initialize
-        self.transaction = nil
-        self.spans = []
+        @transaction = nil
+        @spans = []
       end
 
-      def transaction
-        Thread.current[TRANSACTION_KEY]
-      end
-
-      def transaction=(transaction)
-        Thread.current[TRANSACTION_KEY] = transaction
-      end
-
-      def spans
-        Thread.current[SPAN_KEY] ||= []
-      end
-
-      def spans=(spans)
-        Thread.current[SPAN_KEY] ||= []
-        Thread.current[SPAN_KEY] = spans
-      end
+      attr_accessor :transaction, :spans
     end
 
     def initialize(config, metrics:, stacktrace_builder:, &enqueue)
@@ -61,8 +46,6 @@ module ElasticAPM
       @stacktrace_builder = stacktrace_builder
       @enqueue = enqueue
       @metrics = metrics
-
-      @current = Current.new
     end
 
     attr_reader :stacktrace_builder, :enqueue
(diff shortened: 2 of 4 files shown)


bscharm:
Thanks for taking this on so quickly!

I think our initial expectation was exactly what you described, that `ElasticAPM.current_transaction` within a `with_span` block would be aware of the `parent` transaction as opposed to returning `nil`. I wonder if that is indeed the case, does it make the change to `ErrorBuilder` redundant? Either way I do believe that `current_transaction` should be contextually representative of the span you are in.

bscharm:
We pulled down the fork and tested locally, and while errors are being reported correctly now we are seeing some of the same issues we saw when we tried to manually set the `current_transaction`- spans are being reported inconsistently (most of the time many of the spans seem to be dropped). If we remove the code to set the `current_transaction`  when calling `start_span` (`instrumenter.rb:201` on your branch) it at the very least seems to capture most spans. For the life of me I don't know why this is as the `current` should be entirely local to the thread, but I don't know much about how this all get's reported back to the APM server and if there may be an issue there? Does that behavior sound expected from your perspective?

estolfo:
I'm going to close this because it needs more investigation and a different approach.

PR #999, opened 2021-04-09
Azure: Add Storage Table spy

### LINKED ISSUE #947: [META 410] Instrumentation for CosmosDB
Provide instrumentation for Azure CosmosDB.

Meta issue: elastic/apm#410
SPEC: elastic/apm#411

### LINKED ISSUE #948: [META 410] Instrumentation for Azure Storage
Provide instrumentation for Azure Storage.

Meta issue: elastic/apm#410
SPEC: elastic/apm#411

### LINKED ISSUE #1022: [META 448] Auto-infer destination.service.resource and adapt public API
As described in https://github.com/elastic/apm/issues/448, the agent needs to apply the following changes to comply with the current SPEC for destination fields:

- [ ] Infer destination.service.resource based on other fields
- [ ] Add a public API to set destination.service.resource
- [ ] As destination.service.name and destination.service.type are not used anywhere in the product, I propose to deprecate and remove them also in the public API (if already added there)
- [ ] Add a public API to start an exit span. This auto-infers the destination.service.resource based on the subtype if not set otherwise. Avoids creating nested exit spans and makes the service map work with those custom spans by default, without having to set any destination fields.

mikker:
Closes #948
Closes #947 (Storage Table and CosmosDB use the same lib in Ruby)
Closes #1022 

**ALSO:**

This PR…

- Changes all `span.context.destination.service` fields per [this change](https://github.com/elastic/apm/blob/cb697900cc88fbb4b08e567a77cef68d63f44007/specs/agents/tracing-spans-destination.md#contextdestinationaddress) (only setting `service.resource`)
- Introduces `BasicObject` for making it easier to create _"dumb"_ objects that mostly hold values


gregkalapos:
We use a built-in library called Diagnostic Source to get info for Azure storage. For each operation we get a start and an end event - [here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.Azure.Storage/AzureBlobStorageDiagnosticListener.cs#L35) is the code for it.

> Are there common operation names? Or are they only defined by HTTP method? Do we call the spans by their HTTP operation rather than their Ruby method name then?

What we do here is that we just use the operation name from the name of the event that we get - So if the diagnostic source event is called `BlobBaseClient.Delete.Start` then it's a `Delete` operation which will be part of the span name and we also set `action` to `Delete` in this case. All that can be seen [in the linked code](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.Azure.Storage/AzureBlobStorageDiagnosticListener.cs#L45).

> How do we get region info? From the storage url?

I think we don't collect region info, but what we do is that we indeed pars the URL and get things like `ResourceName` - see [here](https://github.com/elastic/apm-agent-dotnet/blob/master/src/Elastic.Apm.Azure.Storage/AzureBlobStorageDiagnosticListener.cs#L192)

@russcam knows more about this.

gregkalapos:
Oh, and just to be on the same page :) 

Azure Storage and CosmosDb are different things - what I linked above is about Azure Storage.

gregkalapos:
Another update :) 

actually the diagnostic source way I describe above works only for newer libraries. @russcam works on a PR to parse out everything purely from the URLs which will work with older versions as well - I think this will be closer to what you can do in Ruby.

[Here](https://github.com/elastic/apm-agent-dotnet/pull/1258/files#diff-9bc163a38ea5559cda5763de23e2caf0fba0106775f34e418466f781c8fa92c0R24) you can see the URL parsing logic.

lib/elastic_apm/basic_object.rb
```diff
@@ -0,0 +1,74 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+# frozen_string_literal: true
+
+module ElasticAPM
```
mikker, line 20: I though it would be nice to have a common way to create all these little objects that mostly just correspond to API objects, like `{ cloud: region }` or `{ service: name }`.

This is already going a little bit into comlex territory adding `#empty?` and `optional: true` capabilities but that is sort of a general concept for all these fields. Since we don't strip all `nil` values when generating the JSON, I thought that a nice place to put the knowledge of whether an object was _empty_ was inside its own defining class.

`BasicObject` is a loaded term though and probably not the best to use, but I struggled with coming up with something better and didn't want to think too long about it. Alternate names very welcome.

lib/elastic_apm/basic_object.rb
```diff
@@ -0,0 +1,74 @@
...
+  module BasicObject
```
mikker, line 21:
```suggestion
  # An interface for creating simple, value holding objects that correspond to
  # object fields in the API.
  #
  # Example:
  #   class MyThing
  #     include BasicObject
  #     field :name
  #     field :address, optional: true
  #   end
  #   
  #   MyThing.new(name: 'AJ').to_h
  #     # => { name: 'AJ' }
  #   MyThing.new().empty?
  #     # => true
  module BasicObject
```

lib/elastic_apm/span.rb
```diff
@@ -106,6 +106,17 @@ def stop(clock_end = Util.monotonic_micros)
       @duration ||= (clock_end - @clock_start)
       @parent.child_stopped
       @self_time = @duration - child_durations.duration
+
+      if exit_span?
+        context.destination ||= Context::Destination.new
+        context.destination.service ||= Context::Destination::Service.new
+        context.destination.service.resource ||= (subtype || type)
+
+        # deprecated fields
```
mikker, line 115:
```suggestion
        # Deprecated fields but required by some versions of APM Server, so
        # we auto-infer them from existing fields
```

lib/elastic_apm/basic_object.rb
estolfo: Alternate suggestions: `APIObject`, `IntakeObject`, `EventObject`

estolfo (commented):
I added some minor comments but otherwise it looks good to me.

spec/elastic_apm/spies/dynamo_db_spec.rb
```diff
@@ -90,8 +90,8 @@ module ElasticAPM
 
         span = @intercepted.spans.first
 
-        expect(span.name).to eq('DynamoDB BatchGetItem')
-        expect(span.outcome).to eq('failure')
+        expect(span.name).to eq("DynamoDB BatchGetItem")
+        expect(span.outcome).to eq("failure")
       end
     end
   end
```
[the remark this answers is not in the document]
mikker: Rubyfmt 🤷‍♂️

I can change it back if you think it's dumb to start mixing the two styles?

spec/elastic_apm/spies/net_http_spec.rb
```diff
@@ -191,33 +189,12 @@ module ElasticAPM
...
-      end
-
-      span, = @intercepted.spans
-
-      expect(span.name).to eq 'GET [::1]'
-      expect(span.context.destination.service.name).to eq 'http://[::1]'
-      expect(span.context.destination.service.resource).to eq '[::1]:80'
-      expect(span.context.destination.service.type).to eq 'external'
-    end
-
```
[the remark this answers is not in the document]
mikker: Already tested further up https://github.com/mikker/apm-agent-ruby/blob/3297713f525f08cb253863443ee754c3ebb80e5c/spec/elastic_apm/spies/net_http_spec.rb#L192-L194

spec/elastic_apm/spies/dynamo_db_spec.rb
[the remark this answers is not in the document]
estolfo: I think we should stick with one style but have no preference as to single or double. I've just always used single unless double was necessary.

estolfo:
looks good!

mikker:
@estolfo Added instrumentation to a bunch more methods 😊

lib/elastic_apm/spies/azure_storage_table.rb
```diff
@@ -0,0 +1,147 @@
...
+                operation_name.to_s.split("_").collect(&:capitalize).join
+              end
+            end
+          end
+
+          def account_name_from_storage_table_host(host)
+            account_names.compute_if_absent(host) do
+              URI(host).host.split(".").first || "unknown"
+            end
+
```
mikker, line 91:
```suggestion
```

mikker:
This is ready for a new look-through @estolfo 😊

lib/elastic_apm/span.rb
```diff
@@ -106,6 +106,18 @@ def stop(clock_end = Util.monotonic_micros)
       @duration ||= (clock_end - @clock_start)
       @parent.child_stopped
       @self_time = @duration - child_durations.duration
+
+      if exit_span?
+        context.destination ||= Context::Destination.new
+        context.destination.service ||= Context::Destination::Service.new
+        context.destination.service.resource ||= (subtype || type)
+
+        # Deprecated fields but required by some versions of APM Server, so
```
estolfo, line 115: "Deprecated fields *are* required"?

estolfo (approved):

lib/elastic_apm/span.rb
mikker: More like _"These are deprecated fields but they are required…"_

mikker:
```suggestion
        # These fields are deprecated but required by some versions of APM Server, so
```

PR #1030, opened 2021-06-24
Validate span.type and .subtype against shared spec

### LINKED ISSUE #1018: Implement span.type shared spec
See PR https://github.com/elastic/apm/pull/443

mikker:
Closes #1018

- ~Found out that we still sent types in the legacy joined form `type.subtype.action`.~ Turns out this is still Good Behaviour™
- Instead of changing any of the metadata we currently report, this PR only checks against and adds to the spec. This is because we want to get a full picture of the status quo before we possibly start aligning between agents.

### DIFF
--- a/spec/fixtures/span_types.json
+++ b/spec/fixtures/span_types.json
@@ -12,10 +12,10 @@
     "subtypes": {
       "inferred": {
         "comment": "Sampling profiler inferred spans"
+      },
+      "controller": {
+        "comment": "Application controller actions"
       }
-    },
-    "controller": {
-      "comment": "Application controller actions"
     }
   },
   "custom": {
--- a/spec/fixtures/span_types.json
+++ b/spec/fixtures/span_types.json
@@ -14,7 +14,13 @@
         "comment": "Sampling profiler inferred spans"
       },
       "controller": {
-        "comment": "Application controller actions"
+        "comment": "Application controller actions (Ruby/Rails)"
+      },
+      "resource": {
+        "comment": "Application resource actions (Ruby/Grape)"
+      },
+      "graphql": {
+        "comment": "Incoming GraphQL requests"
       }
     }
   },
@@ -93,6 +99,10 @@
         "comment": "HTTP client"
       }
     }
+  },
+  "json": {
+    "comment": "JSON Parsing and generating",
+    "optional_subtype": true
   },
   "messaging": {
     "comment": "Messaging",
--- a/spec/fixtures/span_types.json
+++ b/spec/fixtures/span_types.json
@@ -54,6 +54,7 @@
   },
   "db": {
     "__description": "database span",
+    "allow_unlisted_subtype": true,
     "subtypes": {
       "cassandra": {
         "__description": "Cassandra",
@@ -147,7 +148,12 @@
       "sqlite": {
         "__description": "SQLite",
         "__used_by": [
-          "ruby",
+          "ruby"
+        ]
+      },
+      "sqlite3": {
+        "__description": "SQLite 3",
+        "__used_by": [
           "ruby"
         ]
       },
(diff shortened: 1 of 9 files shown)


spec/fixtures/span_types.json
```diff
@@ -0,0 +1,152 @@
...
+      },
+      "controller": {
+        "comment": "Application controller actions (Ruby/Rails)"
+      },
+      "resource": {
+        "comment": "Application resource actions (Ruby/Grape)"
+      },
+      "graphql": {
+        "comment": "Incoming GraphQL requests"
+      }
```
mikker, line 24: @SylvainJuge These…

spec/fixtures/span_types.json
```diff
@@ -0,0 +1,152 @@
...
+      },
+      "postgresql": {
+        "comment": "PostgreSQL"
+      },
+      "redis": {
+        "comment": "Redis"
+      },
+      "sqlite": {
+        "comment": "SQLite"
+      },
```
mikker, line 81: …and this…

spec/fixtures/span_types.json
```diff
@@ -0,0 +1,152 @@
...
+      "http": {
+        "comment": "HTTP client"
+      }
+    }
+  },
+  "json": {
+    "comment": "JSON Parsing and generating",
+    "optional_subtype": true
+  },
```
mikker, line 106: …and this one, were the ones it took.

SylvainJuge (approved):
Just a few minor questions/comments on my end for the shared JSON spec.

I start to feel that we could add another descriptive field to indicate which agent is using what, that will later help to align and find which ones are mostly used, and which ones are the exceptions & oddities, what do you think ?

mikker:
I think marking agent langs in the spec is a good idea so long as we are in the process of aligning 👍

SylvainJuge:
I've just updated the original PR (https://github.com/elastic/apm/pull/443) with `__used_by` field, not sure if the wording/layout of this is the best but at least we can use it.

spec/fixtures/span_types.json
```diff
@@ -0,0 +1,279 @@
...
+  "custom": {
+    "__description": "API custom instrumentation",
+    "__used_by": [
+      "java", "ruby"
+    ],
+    "allow_null_subtype": true
+  },
+  "db": {
+    "__description": "database span",
+    "allow_unlisted_subtype": true,
```
mikker, line 57: As the subtypes are dynamically set in the Ruby agent, I added this. I've added the types I encountered in our tests but felt that adding this to indicate the dynamic nature might be a good idea too. _WDYT_, @SylvainJuge ?

SylvainJuge: Ideally we want to have all types & sub-types to be properly listed and documented here, thus I'm not sure that having `allow_unlisted_subtype` = `true` makes sense. What are the values for which the Ruby agent can't have the supported databases explicitly listed here ?

mikker: It's not so much that we can't list the ones we know about. It was more a signifier that there might be possible _undiscovered_ values out there 👽

But I agree. I'll remove the option.

spec/fixtures/span_types.json
```diff
@@ -0,0 +1,278 @@
...
+        "__used_by": [
+          "ruby"
+        ]
+      }
+    }
+  },
+  "custom": {
+    "__description": "API custom instrumentation",
+    "__used_by": [
+      "java", "ruby"
```
SylvainJuge, line 51: [very minor] maybe keep one entry per line for consistency.

SylvainJuge (approved):

SylvainJuge:
There are only a few differences with the https://github.com/elastic/apm/pull/443 PR, thus you can directly update it once this is merged without doing a PR on PR :-).

PR #1031, opened 2021-06-24
Central Config header key is 'If-None-Match' for the Etag

### LINKED ISSUE #1011: Service Environment Parameter Missing in Remote Config Fetch API
## Describe the bug
We have added Agent Remote Configuration with Service Environment and Elastic APM agent is not able to fetch the config because service.environment parameter is missing in GET API call ('/config/v1/agents') in central_config.rb file.

Kibana Log:
log   [15:32:59.040] [info][apm][plugins] Config was not found for VanillaApp/undefined
undefined is because of absence of service.environment parameter.

## Expected behavior
Support for service.environment should be there to fetch correct agent remote configuration.

## Environment
Ruby: 2.6.2
Elastic Apm Gem Version: 3.15.1
Rails: 5.1.4

estolfo:
Resolves #1011 

We were using the wrong header key when making requests for the central config updates. It should be `If-None-Match`, not `Etag`. The symptom of this bug was that Kibana was never marking the config update as having been applied (the grey dot stayed grey and never turned to green).

### DIFF
--- a/spec/elastic_apm/central_config_spec.rb
+++ b/spec/elastic_apm/central_config_spec.rb
@@ -192,6 +192,7 @@ module ElasticAPM
             request: { headers: { 'Etag': '___etag___' } }
           )
 
+          expect(subject.send(:headers)).to eq('If-None-Match': '___etag___')
           subject.fetch_and_apply_config
           subject.promise.wait
         end
--- a/spec/elastic_apm/central_config_spec.rb
+++ b/spec/elastic_apm/central_config_spec.rb
@@ -189,10 +189,10 @@ module ElasticAPM
 
           stub_response(
             nil,
-            request: { headers: { 'Etag': '___etag___' } }
+            request: { headers: { 'If-None-Match': '___etag___' } },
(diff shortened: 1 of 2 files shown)


spec/elastic_apm/central_config_spec.rb
```diff
@@ -192,6 +192,7 @@ module ElasticAPM
             request: { headers: { 'Etag': '___etag___' } }
           )
 
+          expect(subject.send(:headers)).to eq('If-None-Match': '___etag___')
```
mikker, line 195: I bet we could add this check to the webmock stub instead for a cleaner approach?

mikker (approved):
TIL! You’re so good at debugging little details like this one.

spec/elastic_apm/central_config_spec.rb
estolfo: I updated the PR, it is now what you were suggesting?

mikker (approved):
Exactly something like that ❤️

PR #1035, opened 2021-07-05
Decrease CentralConfig 304 log level to debug

mikker:
As nothing really changed, logging a message on each succesful, not-changed update from CC was maybe a bit too spammy. Let's decrease the level to DEBUG instead.


mikker:
Test failure not relevant

estolfo (approved):

PR #1040, opened 2021-07-09
Add JVM memory metrics

### LINKED ISSUE #954: System Metrics JRuby support
Collect JVM metrics (CPU usage, heap, ...) on JRuby

mikker:
Closes #954

Uses the same format as https://github.com/elastic/apm-agent-java/blob/master/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/JvmMemoryMetrics.java so we can share the custom dashboards for example.

There's also the [GC metrics](https://github.com/elastic/apm-agent-java/blob/master/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/JvmGcMetrics.java). I don't if they differ from what we get from `GC.stat` [in the "vm set"](https://github.com/mikker/apm-agent-ruby/blob/805724440cbef5fc3d86c3d26254a47d83c99fb6/spec/elastic_apm/metrics/vm_set_spec.rb#L80-L83).

This also enables the OS metrics in JRuby when we're running on Linux. Turns out reading the `/proc` files are just as likely to work in JRuby when on *nix.

### DIFF
--- a/lib/elastic_apm/metrics.rb
+++ b/lib/elastic_apm/metrics.rb
@@ -63,7 +63,10 @@ module ElasticAPM
             breakdown: BreakdownSet,
             transaction: TransactionSet
           }
-          sets[:jvm] = JVMSet if defined?(JVMSet)
+          if defined?(JVMSet)
+            debug "Enabling JVM metrics collection"
+            sets[:jvm] = JVMSet
+          end
 
           @sets = sets.each_with_object({}) do |(key, kls), _sets_|
             debug "Adding metrics collector '#{kls}'"
--- a/lib/elastic_apm/metrics.rb
+++ b/lib/elastic_apm/metrics.rb
@@ -57,16 +57,17 @@ module ElasticAPM
         # Only set the @sets once, in case we stop
         # and start again.
         if @sets.nil?
-          @sets = {
+          sets = {
             system: CpuMemSet,
             vm: VMSet,
             breakdown: BreakdownSet,
             transaction: TransactionSet
           }
-          @sets[:jvm] = JVMSet if defined?(JVMSet)
-          @sets.each_with_object({}) do |(key, kls), sets|
+          sets[:jvm] = JVMSet if defined?(JVMSet)
+
+          @sets = sets.each_with_object({}) do |(key, kls), _sets_|
             debug "Adding metrics collector '#{kls}'"
-            sets[key] = kls.new(config)
+            _sets_[key] = kls.new(config)
           end
         end
 
--- a/lib/elastic_apm/metrics.rb
+++ b/lib/elastic_apm/metrics.rb
@@ -41,6 +41,7 @@ module ElasticAPM
       def initialize(config, &block)
         @config = config
         @callback = block
+        @sets = nil
       end
 
       attr_reader :config, :sets, :callback
@@ -55,16 +56,18 @@ module ElasticAPM
 
         # Only set the @sets once, in case we stop
         # and start again.
-        @sets ||= {
-          system: CpuMemSet,
-          vm: VMSet,
-          breakdown: BreakdownSet,
-          transaction: TransactionSet,
-          # jvm: JVMSet <-- we need this, but dynamically somehow
-          # maybe they register themselves akin to how the spies do?
-        }.each_with_object({}) do |(key, kls), sets|
-          debug "Adding metrics collector '#{kls}'"
-          sets[key] = kls.new(config)
+        if @sets.nil?
+          @sets = {
+            system: CpuMemSet,
+            vm: VMSet,
+            breakdown: BreakdownSet,
+            transaction: TransactionSet
+          }
+          @sets[:jvm] = JVMSet if defined?(JVMSet)
+          @sets.each_with_object({}) do |(key, kls), sets|
+            debug "Adding metrics collector '#{kls}'"
+            sets[key] = kls.new(config)
+          end
         end
 
         @timer_task = Concurrent::TimerTask.execute(
--- a/lib/elastic_apm/metrics.rb
+++ b/lib/elastic_apm/metrics.rb
@@ -59,7 +59,9 @@ module ElasticAPM
           system: CpuMemSet,
           vm: VMSet,
           breakdown: BreakdownSet,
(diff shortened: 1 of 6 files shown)


spec/elastic_apm/metrics/cpu_mem_set_spec.rb
```diff
@@ -25,6 +25,10 @@ module Metrics
       let(:config) { Config.new }
       subject { described_class.new config }
 
+      context 'JRuby' do
+
+      end
+
```
mikker, line 31:
```suggestion
```

mikker:
Test error unrelated

estolfo (commented):
This looks good. Is the only outstanding thing how to dynamically define the JVMSet?
We also need to update the metrics documentation `metrics.asciidoc` because it says everything is unsupported on jruby.

lib/elastic_apm/metrics.rb
```diff
@@ -55,7 +59,9 @@ def start
           system: CpuMemSet,
           vm: VMSet,
           breakdown: BreakdownSet,
-          transaction: TransactionSet
+          transaction: TransactionSet,
+          # jvm: JVMSet <-- we need this, but dynamically somehow
+          # maybe they register themselves akin to how the spies do?
```
estolfo, line 64: Maybe I'm missing something, but can't you just do the following?

```ruby
        if @sets.nil?
          @sets = {
            system: CpuMemSet,
            vm: VMSet,
            breakdown: BreakdownSet,
            transaction: TransactionSet
          }
          @sets[:jvm] = JVMSet if defined?(JRUBY_VERSION)
          @sets.each_with_object({}) do |(key, kls), sets|
            debug "Adding metrics collector '#{kls}'"
            sets[key] = kls.new(config)
          end
        end
```

mikker: Sure, yes, that'll work 👍 Maybe this?

```ruby
@sets[:jvm] = JVMSet if defined?(JVMSet)
```

It's only there if JRuby already and I think it looks better, idk 🤷‍♂️

kovyrin:
Just tested within the Enterprise Search environment.

I can see the metrics in APM indexes:
```json
{
  "_index": "apm-7.13.2-metric-000001",
  "_type": "_doc",
  "_id": "L1D2pXoBs_LaRMrg1YQ-",
  "_version": 1,
  "_score": null,
  "fields": {
    "jvm.memory.heap.used": [
      874747140
    ],
    "service.framework.version": [
      "5.1.7"
    ],
    "jvm.memory.non_heap.committed": [
      95289344
    ],
    "observer.name": [
      "instance-0000000000"
    ],
    "service.node.name": [
      "elastic-m1.local"
    ],
    "host.hostname": [
      "elastic-m1.local"
    ],
    "process.pid": [
      28994
    ],
    "service.language.name": [
      "ruby"
    ],
    "host.ip": [
      "99.251.122.181"
    ],
    "processor.event": [
      "metric"
    ],
    "jvm.memory.heap.committed": [
      1173880830
    ],
    "agent.name": [
      "ruby"
    ],
    "host.name": [
      "elastic-m1.local"
    ],
    "service.environment": [
      "development"
    ],
    "jvm.memory.non_heap.used": [
      91406032
    ],
    "jvm.memory.heap.max": [
      1908932610
    ],
    "service.name": [
      "Enterprise Search"
    ],
    "service.framework.name": [
      "Ruby on Rails"
    ],
    "processor.name": [
      "metric"
    ],
    "service.runtime.name": [
      "jruby"
    ],
    "service.runtime.version": [
      "9.2.13.0"
    ],
    "observer.version_major": [
      7
    ],
    "observer.hostname": [
      "8477c28a175b"
    ],
    "host.architecture": [
      "universal"
    ],
    "metricset.name": [
      "app"
    ],
    "event.ingested": [
      "2021-07-14T17:00:54.973Z"
    ],
    "@timestamp": [
      "2021-07-14T17:00:43.619Z"
    ],
    "service.version": [
      "ce0e85f8b8c1473ff6dde6e8d270e421418d85c0"
    ],
    "ecs.version": [
      "1.8.0"
    ],
    "host.os.platform": [
      "java"
    ],
    "observer.type": [
      "apm-server"
    ],
    "observer.version": [
      "7.13.2"
    ],
    "labels.ent_search_version": [
      "7.15.0"
    ],
    "service.language.version": [
      "2.5.7"
    ],
    "jvm.memory.non_heap.max": [
      -1
    ],
    "agent.version": [
      "4.2.0"
    ],
    "process.title": [
      "/Users/kovyrin-elastic/work/elastic/ent-search/web.xml"
    ]
  },
  "sort": [
    1626282043619
  ]
}
```

Unfortunately, I was not able to find any places within the dashboard where this data would be exposed. Am I missing something?

kovyrin:
By updating the Metrics App index pattern to "apm-*-metric-*,metrics-*,metricbeat-*" I was able to use the metrics explorer feature to find JVM metrics and plot them on a graph - looks great!

Going to test it in a Docker environment on linux now to see OS metrics.

kovyrin:
When running in Docker (a Linux environment):

### Docker icon

image

### CPU and Memory usage in the hosts table

image

### CPU and memory usage graphs on the host page

image

Seems to work as expected!

lib/elastic_apm/metrics.rb
```diff
@@ -51,14 +56,19 @@ def start
...
-          debug "Adding metrics collector '#{kls}'"
-          sets[key] = kls.new(config)
+        if @sets.nil?
+          sets = {
+            system: CpuMemSet,
+            vm: VMSet,
+            breakdown: BreakdownSet,
+            transaction: TransactionSet
+          }
+          sets[:jvm] = JVMSet if defined?(JVMSet)
```
kovyrin, line 66: I've added this small change in my local environment to make it easier to tell if this conditional worked. Seems useful.
```suggestion
          if defined?(JVMSet)
            debug "Enabling JVM metrics collection"
            sets[:jvm] = JVMSet
          end
```

kovyrin (approved):
Worked in my local tests both on a mac and on linux

estolfo:
jenkins run the tests

PR #1042, opened 2021-07-13
Add experimental config option log_ecs_formatting and document it

### LINKED ISSUE #931: [META 373] Extend ECS logging documentation
Extend the ECS logging documentation to cover ECS formatters replacement (if applicable) as described in https://github.com/elastic/apm/issues/373

estolfo:
This PR adds a config option to the agent, `log_ecs_formatting`. If this option is set to "`override`", the agent will attempt to load the `ecs-logging` gem and create a `EcsLogging::Logger` instance for the agent to use. 

The `ecs-logging` gem must be installed before the agent is started. If `log_ecs_formatting` is set to "`override`", the agent will attempt to require the gem and if it cannot be loaded, it will fall back to using the standard Ruby `::Logger` and log the load error.

When using Rails, the agent will not use a `EcsLogging::Logger` instance and instead use the Rails logger. The only way for the agent to use a `EcsLogging::Logger` instance when on Rails, is to set the `logger` attribute on the `config`.

Closes #931

### DIFF
--- a/lib/elastic_apm/config.rb
+++ b/lib/elastic_apm/config.rb
@@ -70,6 +70,7 @@ module ElasticAPM
     option :ignore_url_patterns,               type: :list,   default: [],      converter: RegexpList.new
     option :instrument,                        type: :bool,   default: true
     option :instrumented_rake_tasks,           type: :list,   default: []
+    option :log_ecs_formatting,                type: :string, default: 'off'
     option :log_level,                         type: :int,    default: Logger::INFO, converter: LogLevelMap.new
     option :log_path,                          type: :string
     option :metrics_interval,                  type: :int,    default: '30s',   converter: Duration.new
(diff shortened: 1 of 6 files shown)


lib/elastic_apm/config.rb
```diff
@@ -246,9 +247,28 @@ def load_env
...
+      if self.log_ecs_formatting == 'override'
+        begin
+          require 'ecs_logging/logger'
+          return ::EcsLogging::Logger.new(log_path == '-' ? $stdout : log_path).tap do |logger|
+            logger.level = log_level
+          end
+        rescue LoadError
+          load_error = true
+        end
+      end
```
mikker, line 260: I think we could make this easier to read for ourselves by splitting it up into a few smaller private methods instead.

mikker (changes requested):
Overall looks good. Nice investigation. I'll bring it home, enjoy your vacation time 😊

mikker:
Continued in https://github.com/elastic/apm-agent-ruby/pull/1053

PR #1054, opened 2021-08-13
ES Spy: Fake out verification for 7.14+

mikker:


mikker:
Test errors unrelated

PR #1148, opened 2021-08-25
Add ability to queue events and send them synchronously

estolfo:
POC for queuing events and sending them synchronously when the agent is shutdown.
This is necessary for testing the lambda extension.


estolfo:
This shouldn't need to be merged, it's just being used to experiment with lambda extensions so I'm going to close it.

PR #1152, opened 2021-08-30
Make Subscriber compatible with rails 3

vollnhals:

This pull requests makes the Subscriber compatible with Rails 3, so that the `elastic-apm` gem can be used in a rails 3 application.

It does this by providing a `call` method in the Subscriber, that rails 3 will call instead of `start` and `finish`.

The resulting code works both in rails 3 and modern rails versions.

## Why is it important?

We have a large application that still uses rails 3. We would like to monitor this application. 

If rails 3 support is possible with minimal changes and minimal burden of maintenance then we think it would beneficial to the rails community (that is stuck on older versions) be able to use `elastic-apm`.

## Details

Because we cannot rely on `start` to record the correct monotonic microseconds timestamp, we have to fix the times when the subscriber is called by rails 3.

Besides this change the new methods on Hash like `transform_values` are required when running on an old ruby version, like 2.3.x. We use the [backports](https://github.com/marcandre/backports) gem for that. 

## Checklist

- [ ] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/). 
- [ ] I have rebased my changes on top of the latest master branch

- [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-ruby/blob/master/CONTRIBUTING.md#testing) pass locally with my changes

### DIFF
--- a/lib/elastic_apm/span.rb
+++ b/lib/elastic_apm/span.rb
@@ -80,7 +80,9 @@ module ElasticAPM
       :outcome,
       :subtype,
       :trace_context,
-      :type
+      :type,
+      :timestamp,
+      :clock_start
     )
     attr_reader(
       :context,
@@ -89,7 +91,6 @@ module ElasticAPM
       :sample_rate,
       :self_time,
       :stacktrace,
-      :timestamp,
       :transaction
     )
 
--- a/lib/elastic_apm/subscriber.rb
+++ b/lib/elastic_apm/subscriber.rb
@@ -88,6 +88,31 @@ module ElasticAPM
     end
     # rubocop:enable Metrics/CyclomaticComplexity
 
+    # @deprecated provide `call` method for Rails 3 backward compatibility
+    # instead of having start and finish, rails 3 only calls `call` method
+    # so we have to be creative and fixup the start span
+    def call(name, start_time, finish_time, id, payload)
+      return unless (transaction = @agent.current_transaction)
+      # create start event and then fixup the recorded times in the span
+      notifications = start(name, id, payload)
+      span = notifications.last.span
+
+      if span
+        # set timestamp to match start_time
+        start_micros = Util.micros(start_time)
+        span.timestamp = start_micros
+
+        # when setting clock_start we have to be more careful because clock_start uses monotonic time and the values we get
+        # are relative to different origin than what normal Time.now uses
+        finish_micros = Util.micros(finish_time)
+        duration = finish_micros - start_micros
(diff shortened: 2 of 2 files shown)


mikker:
As Rails 3.2 has been EOL'ed for at least five years, I feel hesitant to add code to support it. Is this not something you could patch in your own code?

That way we wouldn't have to add support for a version that's already officially unsupported.

vollnhals:
Would you be willing to accept a patch that just makes `:timestamp` and `:clock_start` attr_accessible on `Span`?

This way we could more easily monkey patch it ourselves.

mikker:
Could you not do that in a patch too?

Something like 

```ruby
module Rails3Patch
  self.included(cls)
    cls.class_eval do
      attr_accessor :timestamp, :clock_start
    end
  end
end

ElasticAPM::Subscriber.include(Rails3Patch)
```

(Untested, probably errors but you get the idea 😊)

mikker:
@vollnhals Did you figure something out? Would be nice to have whatever you made up here for others to find via search 😉

mikker:
Please follow up with your solution if you figure something out. Thanks for the effort and for considering ElasticAPM.

vollnhals:
@mikker We are using the following patch in `config/initializers/20_elastic_apm_subscriber_rails_3_compat.rb` to make it work with rails 3. 

```ruby
module ElasticApmSubscriberRails3Compat
  # provide `call` method for Rails 3 backward compatibility
  # instead of having start and finish, rails 3 only calls `call` method
  # so we have to be creative and fixup the start span
  def call(name, start_time, finish_time, id, payload)
    return unless (transaction = @agent.current_transaction)
    # create start event and then fixup the recorded times in the span
    notifications = start(name, id, payload)
    span = notifications.last.span

    if span
      # set timestamp to match start_time
      start_micros = ElasticAPM::Util.micros(start_time)
      span.timestamp = start_micros

      # when setting clock_start we have to be more careful because clock_start uses monotonic time and the values we get
      # are relative to different origin than what normal Time.now uses
      finish_micros = ElasticAPM::Util.micros(finish_time)
      duration = finish_micros - start_micros
      # set clock start relative to monotonic time source by subtracting the event duration
      span.clock_start = ElasticAPM::Util.monotonic_micros - duration
    end

    finish(name, id, payload)
  end
end

ElasticAPM::Subscriber.prepend(ElasticApmSubscriberRails3Compat)

ElasticAPM::Span.class_eval do
  attr_accessor :timestamp, :clock_start
end
```

mikker:
Thank you for taking the time 🙏

PR #1154, opened 2021-08-31
Update exit span definition

mikker:
Empty strings in destination fields: https://github.com/elastic/apm/pull/494

### DIFF
--- a/lib/elastic_apm/fields.rb
+++ b/lib/elastic_apm/fields.rb
@@ -42,6 +42,8 @@ module ElasticAPM
       end
 
       attr_reader :key, :value, :optional, :default
+
+      alias :optional? :optional
     end
 
     module InstanceMethods
@@ -60,7 +62,7 @@ module ElasticAPM
       def empty?
         self.class.schema.each do |key, field|
           next unless send(key)
-          next if field.optional
+          next unless (field.optional? && field.value.nil?)
           return false
         end
 
--- a/lib/elastic_apm/fields.rb
+++ b/lib/elastic_apm/fields.rb
@@ -25,7 +25,7 @@ module ElasticAPM
   #   class MyThing
   #     include Fields
   #     field :name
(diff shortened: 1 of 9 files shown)


lib/elastic_apm/fields.rb
```diff
@@ -33,56 +33,69 @@ module ElasticAPM
...
   module Fields
+    class Field
+      def initialize(key, value: nil, optional: false, default: nil)
+        @key = key
+        @value = value
+        @optional = optional
+        @default = default
+      end
+
+      attr_reader :key, :value, :optional, :default
```
estolfo, line 44: we could add
`alias :optional? :optional`

spec/elastic_apm/fields_spec.rb
```diff
@@ -34,12 +35,25 @@ class MyObject
...
+        expect(subject.online).to be false
+      end
     end
 
     describe "#empty?" do
       it "is when missing all values" do
         subject = MyObject.new
+        subject.online = nil
         expect(subject).to be_empty
       end
```
estolfo, line 58: I think we should add a test for when only optional fields have a value. For example:

```ruby
    describe "#empty?" do
      class MyOptionalObject
        include Fields

        field :age, optional: true
        field :language, optional: true
      end

      it "doesn't consider optional fields" do
        subject = MyOptionalObject.new
        subject.age = 11
        subject.language = 'english'
        expect(subject).to be_empty
      end
    end
```
Is that even the behavior we would expect?

estolfo (changes requested):

spec/elastic_apm/fields_spec.rb
mikker: Thinking about it again, I don't even know what _optionals_ are supposed to signify.

estolfo (approved):

PR #1155, opened 2021-09-01
Use Socket.gethostname instead

### LINKED ISSUE #1151: APM hostname collection on detected_hostname raising exceptions
## Describe the bug
The ElasticAPM worker fails to start and the exception points to 
https://github.com/mikker/apm-agent-ruby/blob/9a0d6aa7b52a0b0c699578fe85d0a40db38d4a52/lib/elastic_apm/metadata/system_info.rb#L52-L54

as the exception location

## Steps to reproduce
* Currently running this inside Openshift, ruby 2.5.5, puma 5.4.0, apm gem 4.3.0
* Tried multiple configurations for hostname, service_node_name, etc without success
* Also tried with puma in single mode and clustered mode and it did not make a difference

```
#<Thread:0x00000000034bc960@/opt/app-root/src/bundle/ruby/2.5.0/gems/elastic-apm-4.3.0/lib/elastic_apm/transport/base.rb:140 run> terminated with exception (report_on_exception is true):
--
  | /opt/app-root/src/bundle/ruby/2.5.0/gems/elastic-apm-4.3.0/lib/elastic_apm/metadata/system_info.rb:53:in ``': No such file or directory - hostname (Errno::ENOENT)
  | from /opt/app-root/src/bundle/ruby/2.5.0/gems/elastic-apm-4.3.0/lib/elastic_apm/metadata/system_info.rb:53:in `detect_hostname'
  | from /opt/app-root/src/bundle/ruby/2.5.0/gems/elastic-apm-4.3.0/lib/elastic_apm/metadata/system_info.rb:28:in `initialize'
  | from /opt/app-root/src/bundle/ruby/2.5.0/gems/elastic-apm-4.3.0/lib/elastic_apm/metadata.rb:26:in `new'
  | from /opt/app-root/src/bundle/ruby/2.5.0/gems/elastic-apm-4.3.0/lib/elastic_apm/metadata.rb:26:in `initialize'
  | from /opt/app-root/src/bundle/ruby/2.5.0/gems/elastic-apm-4.3.0/lib/elastic_apm/transport/connection.rb:42:in `new'
  | from /opt/app-root/src/bundle/ruby/2.5.0/gems/elastic-apm-4.3.0/lib/elastic_apm/transport/connection.rb:42:in `initialize'
  | from /opt/app-root/src/bundle/ruby/2.5.0/gems/elastic-apm-4.3.0/lib/elastic_apm/transport/worker.rb:52:in `new'
  | from /opt/app-root/src/bundle/ruby/2.5.0/gems/elastic-apm-4.3.0/lib/elastic_apm/transport/worker.rb:52:in `initialize'
  | from /opt/app-root/src/bundle/ruby/2.5.0/gems/elastic-apm-4.3.0/lib/elastic_apm/transport/base.rb:141:in `new'
  | from /opt/app-root/src/bundle/ruby/2.5.0/gems/elastic-apm-4.3.0/lib/elastic_apm/transport/base.rb:141:in `block in boot_worker'
```

estolfo:
Resolves #1151 

Thanks @kovyrin

### DIFF
--- a/lib/elastic_apm/metadata/system_info.rb
+++ b/lib/elastic_apm/metadata/system_info.rb
@@ -50,7 +50,7 @@ module ElasticAPM
       private
 
       def detect_hostname
-        `hostname`.chomp
+        Socket.gethostname.chomp
       end
     end
   end
--- a/lib/elastic_apm/metadata/system_info.rb
+++ b/lib/elastic_apm/metadata/system_info.rb
@@ -51,6 +51,8 @@ module ElasticAPM
 
       def detect_hostname
         Socket.gethostname.chomp
+      rescue
+        ''
       end
     end
   end
--- a/lib/elastic_apm/metadata/system_info.rb
+++ b/lib/elastic_apm/metadata/system_info.rb
@@ -52,7 +52,6 @@ module ElasticAPM
       def detect_hostname
         Socket.gethostname.chomp
       rescue
-        ''
       end
     end
   end
--- a/spec/elastic_apm/metadata/system_info_spec.rb
+++ b/spec/elastic_apm/metadata/system_info_spec.rb
@@ -41,7 +41,7 @@ module ElasticAPM
           end
 
           it 'returns an empty string' do
-            expect(subject.detected_hostname).to eq('')
+            expect(subject.detected_hostname).to eq(nil)
           end
         end
       end
--- a/spec/elastic_apm/metadata/system_info_spec.rb
+++ b/spec/elastic_apm/metadata/system_info_spec.rb
@@ -40,7 +40,7 @@ module ElasticAPM
             allow(Socket).to receive(:gethostname).and_raise(StandardError)
           end
 
-          it 'returns an empty string' do
+          it 'returns nil' do
             expect(subject.detected_hostname).to eq(nil)
           end
         end
--- a/spec/elastic_apm/metadata/system_info_spec.rb
+++ b/spec/elastic_apm/metadata/system_info_spec.rb
@@ -34,6 +34,16 @@ module ElasticAPM
         it 'has no newline at the end' do
           expect(subject.detected_hostname).not_to match(/\n\z/)
         end
+
+        context 'when there is an exception' do
+          before do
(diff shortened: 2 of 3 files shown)


lib/elastic_apm/metadata/system_info.rb
```diff
@@ -50,7 +50,7 @@ def gem_platform
       private
 
       def detect_hostname
-        `hostname`.chomp
+        Socket.gethostname.chomp
```
mikker, line 53: Do we know whether it's already chomped by Socket?

mikker: And does it ever raise?

estolfo: I believe it can, looking at the c source [here](https://ruby-doc.org/stdlib-2.5.3/libdoc/socket/rdoc/Socket.html#method-c-gethostname)
I'll add in error handling

estolfo: @mikker I updated this. Looking at the c source, `chomp` is not called on the hostname and all the possible errors raised are `StandardErrors` so I just `rescue` and return the empty string.

spec/elastic_apm/metadata/system_info_spec.rb
```diff
@@ -34,6 +34,16 @@ module ElasticAPM
...
           expect(subject.detected_hostname).not_to match(/\n\z/)
         end
+
+        context 'when there is an exception' do
+          before do
+            allow(Socket).to receive(:gethostname).and_raise(StandardError)
+          end
+
+          it 'returns an empty string' do
+            expect(subject.detected_hostname).to eq('')
```
mikker, line 44: Why empty string and not `nil`? The latter would skip the whole field in the serializer code above.

estolfo: I came to the same conclusion but forgot to update the test. The `detected_hostname` method returns `nil` when there's an exception.

mikker (approved):

PR #1157, opened 2021-09-03
Handle NaN in metrics

### LINKED ISSUE #1156: Failed converting event to JSON because of NaN values
## Describe the bug

I'm seeing some "Failed converting event to JSON" errors in APM logs in Kibana.

## Steps to reproduce

Sample message in the error logs (notice the `NaN` values that cannot be serialized with JSON):

```
"[ElasticAPM] {:metricset=>{:timestamp=>1630572097705853, :samples=>{:\"system.cpu.total.norm.pct\"=>{:value=>NaN}, :\"system.memory.actual.free\"=>{:value=>13053140992}, :\"system.memory.total\"=>{:value=>16509009920}, :\"system.process.cpu.total.norm.pct\"=>{:value=>NaN}, :\"system.process.memory.size\"=>{:value=>1940881408}, :\"system.process.memory.rss.bytes\"=>{:value=>207499264}}}}"
```

## Expected behavior

Don't see errors in Kibana APM logs.

## Environment

- **OS**: LInux
- **Ruby version**: 2.7.4
- **Framework and version**: Rails 6.1.4
- **APM Server version**: 7.13.4
- **Agent version**: 4.3.0

## Additional context

Not sure if you could prevent getting `NaN`s or just not send the request if it failed to get the correct values:

```ruby
>> nan = 0.0/0
=> NaN
>> { a: nan }.to_json
JSON::GeneratorError (865: NaN not allowed in JSON)
```

mikker:
Fixes #1156

### DIFF
--- a/lib/elastic_apm/metrics/metric.rb
+++ b/lib/elastic_apm/metrics/metric.rb
@@ -57,6 +57,8 @@ module ElasticAPM
         @mutex.synchronize do
           collected = @value
 
+          return nil if collected.is_a?(Float) && collected.nan?
+
           @value = initial_value if reset_on_collect?
 
           return nil if reset_on_collect? && collected == 0
--- a/lib/elastic_apm/metrics/metric.rb
+++ b/lib/elastic_apm/metrics/metric.rb
@@ -57,7 +57,7 @@ module ElasticAPM
         @mutex.synchronize do
           collected = @value
 
-          return nil if collected.is_a?(Float) && collected.nan?
+          return nil if collected.is_a?(Float) && !collected.finite?
 
           @value = initial_value if reset_on_collect?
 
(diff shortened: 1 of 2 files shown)


lib/elastic_apm/metrics/metric.rb
```diff
@@ -57,6 +57,8 @@ def collect
         @mutex.synchronize do
           collected = @value
 
+          return nil if collected.is_a?(Float) && collected.nan?
```
estolfo, line 60: How about checking `collected.finite?` just to cover all bases?
https://ruby-doc.org/core-2.7.0/Float.html#method-i-finite-3F

mikker:
```suggestion
          return nil if collected.is_a?(Float) && !collected.finite?
```

mikker: Good suggestion!

estolfo (approved):

PR #1180, opened 2021-11-01
Add span compression

mikker:
See #1021 

Currently passing all tests but one, which is concerning this part from the spec:

```ruby
    if (!child.isCompressionEligible()) {
        if (buffered != null) {
            report(buffered) # <-- reporting buffered from sibling
            buffered = null  # <-- resetting buffer
        }
        report(child)
        return
    }
```

Keeping track of children and buffers and such is very difficult and the fact that we use the instrumenter to _report_ instead of letting the spans themselves do it doesn't make it easier.


mikker:
Hey friends! Closing this one because cleaning my GH up a bit. Will leave the branch – but maybe copy it somewhere if you want to make sure to not lose the work. Hope you're all well ❤️

PR #1183, opened 2021-11-02
span type/subtype alignment

SylvainJuge:

Adds changes to shared spec in https://github.com/elastic/apm/pull/513 
- allows to identify what breaks int he ruby agent by removing deprecated type/subtype values
- unblocks merging https://github.com/elastic/apm/pull/513 

## Checklist

- [x] remove all references to any deprecated field in spec
- [ ] modify ruby agent implementation to fit the updated spec
- [ ] once merged (or about to be), back-port changes to the shared spec

### DIFF
--- a/spec/fixtures/span_types.json
+++ b/spec/fixtures/span_types.json
@@ -20,32 +20,37 @@
           "java"
         ]
       },
+      "internal": {
+        "__description": "Application generic internal span for controller/handler/processing delegation",
+        "__used_by": [
+        ]
+      },
       "controller": {
-        "__description": "Application controller actions",
+        "__description": "Deprecated: use app/internal instead",
         "__used_by": [
           "ruby"
         ]
       },
       "graphql": {
-        "__description": "Incoming GraphQL requests",
+        "__description": "Deprecated: use app/internal instead",
         "__used_by": [
           "ruby"
         ]
       },
       "mailer": {
-        "__description": "Application mailer actions",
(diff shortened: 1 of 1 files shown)


jaggederest:
Hi @SylvainJuge does this need attention from me? Just going through the open PRs, I know this is a draft so if it's not moving forward can we close it? Happy to take over if you prefer that as well.

SylvainJuge:
Hi @jaggederest , thanks for cleaning-up old PRs, this was my attempt to modify the specification for the ruby agent regarding span type/subtype alignment.

From what I remember, the ruby agent does enforce the span types & subtypes from the shared spec in https://github.com/elastic/apm/blob/main/tests/agents/json-specs/span_types.json.

So here, the next steps here are:
- update this PR with the latest state of `main`
- make sure that the copy of `span_types.json` that is stored in this repo is in sync with https://github.com/elastic/apm/blob/main/tests/agents/json-specs/span_types.json (the later was updated since).
- make the ruby agent avoid any deprecated type/subtype, which can be done in multiple steps.

While not being "strictly urgent", providing alignment here helps for quite other features that indirectly use span type and subtype for resource destination, compressed spans, destination metrics, ... On the other end, it is a "potentially breaking change" as any user relying on the current values would be impacted (quite minimal, but worth being aware of).

jaggederest:
Great, I'll work on that then 👍

PR #1186, opened 2021-11-04
Update jruby version in docker images

estolfo:
We need to update the jruby version so we get the fix for this bug https://github.com/jruby/jruby/issues/6547


mikker (approved):
Maybe we should add 9.3.x too?

PR #1208, opened 2021-12-06
Make Intercept event submission thread-safe and improve mongo spy

### LINKED ISSUE #1207: Re-enable and fix MongoDB spy spec on JRuby
To get the release out, I disabled the spec when JRuby. It should be possible to make it work, so we should probably.

The test:
https://github.com/elastic/apm-agent-ruby/blob/master/spec/elastic_apm/spies/mongo_spec.rb#L120

The PR:
https://github.com/elastic/apm-agent-ruby/pull/1202

estolfo:
This PR fixes some thread safety issues with the way we test using Intercept and improves the MongoSpy:

The Intercept event arrays (`@transactions`, `@spans`, `@errors`, `@metricsets`) were not thread-safe so submitting events to them from multiple threads would inevitably end up causing event loss on JRuby.

I've removed the unnecessary `@collection` object in the Mongo Spy. That object is shared amongst threads. Even though it wouldn't be a problem with the `events` object being a thread local object, I'm wary of leaving an object shared amongst threads in the spy.

Lastly, the test didn't need to create 1000 threads to test the thread safety of the code and it created identical events in each thread. I've changed the test to give a unique operation id to each event and to test only using 50 threads, as that is sufficient.

Resolve #1207

### DIFF
--- a/lib/elastic_apm/spies/mongo.rb
+++ b/lib/elastic_apm/spies/mongo.rb
@@ -37,14 +37,8 @@ module ElasticAPM
 
         EVENT_KEY = :__elastic_instrumenter_mongo_events_key
 
-        class Collection
-          def events
-            Thread.current[EVENT_KEY] ||= {}
-          end
-        end
-
-        def initialize
-          @collection = Collection.new
+        def events
+          Thread.current[EVENT_KEY] ||= {}
         end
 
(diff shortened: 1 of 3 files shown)


mikker (approved):
This is great!

PR #1209, opened 2021-12-07
MongoDB: Make sure to end the right spans

mikker:


mikker:
@estolfo We're green thanks to you 😊

lib/elastic_apm/spies/mongo.rb
```diff
@@ -98,7 +98,7 @@ def pop_event(event)
           span = events.delete(event.operation_id)
           return unless (curr = ElasticAPM.current_span)
 
-          curr == span && ElasticAPM.end_span
+          curr == span && ElasticAPM.end_span(span)
```
estolfo, line 101: Will this change actually make a difference? 
Line 101 will only call `end_span` if the span is the last span in the list of `current_spans` (because `curr` is compared to `span`). So passing a span in to `ElasticAPM.end_span` won't change anything.
So I think in order to actually make a difference, we need to get rid of line 99 and just have `ElasticAPM.end_span(span)`, in case it's not the last span in the list of current spans.

lib/elastic_apm/spies/mongo.rb
```diff
@@ -98,7 +98,7 @@ def pop_event(event)
           span = events.delete(event.operation_id)
           return unless (curr = ElasticAPM.current_span)
```
estolfo, line 99:
```suggestion
          ElasticAPM.end_span(span)
```

lib/elastic_apm/spies/mongo.rb
estolfo:
```suggestion
```

estolfo (changes requested):
..

lib/elastic_apm/spies/mongo.rb
```diff
@@ -96,9 +96,8 @@ def push_event(event)
 
         def pop_event(event)
           span = events.delete(event.operation_id)
```
mikker, line 98: If the agent is not recording, will this not be a problem? We have an early return in `push_event` if there's no transaction. Shouldn't we have the same check here?

estolfo:
I'm going to close this, as it's superseded by #1211

PR #1211, opened 2021-12-09
Use stack for mongo spy spans

estolfo:
Instead of relying on the operation id, let's just use a stack to keep track of mongo spy spans.

### DIFF
--- a/lib/elastic_apm/spies/mongo.rb
+++ b/lib/elastic_apm/spies/mongo.rb
@@ -41,10 +41,6 @@ module ElasticAPM
           Thread.current[EVENT_KEY] ||= []
         end
 
-        def initialize
-          @mutex = Mutex.new
-        end
-
         def started(event)
           push_event(event)
         end
@@ -87,13 +83,13 @@ module ElasticAPM
                   event.command_name].compact.join('.')
 
           span =
-              ElasticAPM.start_span(
-                name,
-                TYPE,
-                subtype: SUBTYPE,
-                action: ACTION,
-                context: build_context(event)
-              )
+            ElasticAPM.start_span(
+              name,
+              TYPE,
+              subtype: SUBTYPE,
+              action: ACTION,
+              context: build_context(event)
+            )
 
           events << span
         end
--- a/lib/elastic_apm/spies/mongo.rb
+++ b/lib/elastic_apm/spies/mongo.rb
@@ -41,6 +41,10 @@ module ElasticAPM
           Thread.current[EVENT_KEY] ||= []
         end
 
+        def initialize
+          @mutex = Mutex.new
+        end
+
         def started(event)
           push_event(event)
         end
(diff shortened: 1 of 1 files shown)


lib/elastic_apm/spies/mongo.rb
```diff
@@ -91,14 +91,13 @@ def push_event(event)
...
-          events[event.operation_id] = span
+          events << span
         end
 
         def pop_event(event)
-          span = events.delete(event.operation_id)
           return unless (curr = ElasticAPM.current_span)
 
-          curr == span && ElasticAPM.end_span
+          curr == events[-1] && ElasticAPM.end_span(events.pop)
```
mikker, line 100: Do we need a Mutex around this line?

estolfo: That's a good question. The `events` list is a thread-local variable so we wouldn't need a mutex to protect mutating that, however, what could happen, is one thread sets `curr` in line 98 and then another thread compares `curr` to its thread local `event` stack.
So we'll need a mutex around the whole thing, to make sure the current span is being compared to the correct thread's `events` stack.

lib/elastic_apm/spies/mongo.rb
```diff
@@ -64,41 +68,44 @@ def succeeded(event)
...
+              ElasticAPM.start_span(
+                name,
+                TYPE,
+                subtype: SUBTYPE,
+                action: ACTION,
+                context: build_context(event)
+              )
+
+            events << span
+          end
```
estolfo, line 100: This could alternatively be a mutex just around 

```ruby
return unless ElasticAPM.current_transaction
events << span
```

but then we would waste creating the span if there is no current transaction. The tradeoff is having that whole block be synchronous.

lib/elastic_apm/spies/mongo.rb
estolfo: Update: the current span is a thread local variable so we don't need a mutex at all.

mikker: Yes, I was thinking that another span might get pushed during `&&` in this line, right?

estolfo: No, because the `events` list is local to a thread. So another thread can't possibly push to another thread's `events` list.

mikker: Right 👍

estolfo:
@jclusso Do you want to try this PR out? We think it might solve the MongoDB spans issue. Please let us know how it goes!

jclusso:
@estolfo giving it a shot now!

jclusso:
Looks promising so far. I'll update you after a day or so.
image

jclusso:
Looks promising so far. I'll update you after a day or so.
image

estolfo:
@jclusso Great! We look forward to hearing back tomorrow!

jclusso:
@estolfo no luck :( - on the bright side I did find a span that is traced and you can see all the stacktraces that don't make any sense.

mikker (approved):

jclusso:
@estolfo I noticed we're still using your fork on the `mongo-forking` branch. Should we still be using this or is that merged in? I can't find the issue where you told us to give that a try.

PR #1214, opened 2021-12-16
Clear thread local variables when forked

estolfo:
Thread local variables are copied over when a process is forked so we should clear the MongoSpy thread locals when forked.

### DIFF
--- a/lib/elastic_apm/agent.rb
+++ b/lib/elastic_apm/agent.rb
@@ -288,6 +288,8 @@ module ElasticAPM
       instrumenter.handle_forking!
       metrics.handle_forking!
 
+      Spies::MongoSpy::Subscriber.handle_forking!
+
       @pid = Process.pid
     end
   end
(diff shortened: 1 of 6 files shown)


mikker:
Wow, what a find! 👏

jclusso:
@estolfo still no luck for us. We're still seeing all those random spans getting collected on that one worker.

jclusso:
@estolfo we're still using this branch and I was wondering if this is going to get merged or what exactly is happening here. Wanted to get us back on the main version.

estolfo:
Hi @jclusso I made these changes hoping they would fix the issues you were seeing but when we tested, they didn't. So I am hesitant to merge these changes without a specific reason, though in theory, they make sense. @jaggederest is working on the agent now so I'll ask him to review these changes and help decide if they should be merged.
By the way, have you tried the latest stable of the Ruby agent recently? Does it work for you?

jclusso:
@estolfo should we close this since it's merged?

PR #1218, opened 2022-01-04
ruby: 3.1 is now available

v1v:

Add 3.1 ruby in the CI

## Why is it important?

3.1 has been just released


v1v:
3.1 is not supported yet

PR #1278, opened 2022-06-02
Query apm server for version and details about supported features

### LINKED ISSUE #1274: Query APM server for version and details about supported features
We would like to gate some features in the agent on the APM server version (e. g. #1194 ) and currently the agent does not query for this information

jaggederest:

This adds a `version` method to the Config class, which internally checks the server for the value of it's version in the configuration data returned from the root url

## Why is it important?

This will be used to gate some features based on server version, most directly elastic/apm#545

## Checklist

## Related issues

Closes #1274

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -42,6 +42,7 @@ endif::[]
 ===== Added
 
 - Added transaction_name to reported error to allow grouping by transaction name {pull}1267[#1267]
+- Added ability to query server for version (useful in the future) {pull}1278[#1278]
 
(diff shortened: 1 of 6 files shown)


estolfo (approved):
Looks good!
We should think about whether we want to get the version lazily or explicitly at agent start.

felixbarny:
> We should think about whether we want to get the version lazily or explicitly at agent start.

Ideally, querying the version should not block both the application startup and the agent startup. In other words, it's fine if the agent starts creating transactions and we only know about the server version some time later.

PR #1279, opened 2022-06-16
Add new fields to service to track new data type and name

### LINKED ISSUE #1244: [META 621] Improve Granularity for SQL Databases
[issue details](https://github.com/elastic/apm/issues/621)

[issue details](https://github.com/elastic/apm/issues/622)

- start collecting `span.service.target.name` and `span.service.target.type` fields and deprecate `span.destination.service.resource`.
- depending on APM server version send 
  - `span.destination.service.resource` (APM server < 8.3)
  - `span.service.target.name` and `span.service.target.type` (APM server >= 8.3)
- replace the internal usage of the `span.destination.service.resource` field with the new fields (e.g. compressed spans algorithm, inferring `span.service.target.name`, `span.service.target.type` instead of `span.destination.service.resource`).

jaggederest:

Adds new fields to agent-server protocol, specifically span.context.service.target.type and span.context.service.target.name

## Why is it important?

These fields will be used in the future as the correct way to capture service details in the agent-server API

## Checklist

## Related issues

Relates #1244 (partially completes)

### DIFF
--- a/lib/elastic_apm/span.rb
+++ b/lib/elastic_apm/span.rb
@@ -145,6 +145,9 @@ module ElasticAPM
         service: service,
         cloud: cloud
       )
+      context.service = Span::Context::Service.new(
+        target: Span::Context::Service::Target.new(type: context.destination.service.name, name: context.destination.service.type )
+      )
     end
 
     def inspect
--- a/lib/elastic_apm/span.rb
+++ b/lib/elastic_apm/span.rb
@@ -146,7 +146,7 @@ module ElasticAPM
         cloud: cloud
       )
       context.service = Span::Context::Service.new(
-        target: Span::Context::Service::Target.new(type: context.destination.service.name, name: context.destination.service.type )
+        target: Span::Context::Service::Target.new(name: context.destination.service.name, type: context.destination.service.type )
       )
     end
 
--- a/lib/elastic_apm/span/context/service.rb
+++ b/lib/elastic_apm/span/context/service.rb
@@ -1,0 +1,55 @@
+# Licensed to Elasticsearch B.V. under one or more contributor
+# license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright
+# ownership. Elasticsearch B.V. licenses this file to you under
+# the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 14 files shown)


lib/elastic_apm/span/context/service.rb
```diff
@@ -0,0 +1,55 @@
...
+
+        def initialize(target: nil, **attrs)
+          super(**attrs)
+
+          self.target = build_target(target)
+        end
+
+        private
+
+        def build_target(cloud = nil)
```
estolfo, line 45: What is this `cloud` arg for?

estolfo (changes requested):

estolfo (changes requested):
We also need to add in the functionality to actually capture these values for SQL databases.

estolfo (changes requested):
I think it would make sense to add tests for the database instrumentations, right?
Also, don't we have to keep the existing fields for compatibility with the UI?

jaggederest:
> I think at this point, we are only supposed to add the new fields and not stop sending any existing fields.

Just for posterity's sake, the existing sequel spy did not send destination.service.name or destination.service.type before this PR, so they are not being sent. Had to double check that to make sure I didn't accidentally delete them.

lib/elastic_apm/span.rb
```diff
@@ -145,6 +145,9 @@ def set_destination(address: nil, port: nil, service: nil, cloud: nil)
         service: service,
         cloud: cloud
       )
+      context.service = Span::Context::Service.new(
+        target: Span::Context::Service::Target.new(type: context.destination.service.name, name: context.destination.service.type )
```
estolfo, line 149:
```suggestion
        target: Span::Context::Service::Target.new(name: context.destination.service.name, type: context.destination.service.type )
```

estolfo: I also think that this should have been caught by a test. Can we add one that would cover this part of the code?

estolfo (changes requested):
I noticed a typo so proposed a correction but I also suggest adding a test that would have caught the typo.

lib/elastic_apm/span.rb
jaggederest: Oh good catch, and yes, absolutely, adding a test now 👍

estolfo (approved):

estolfo:
I just saw that the tests are failing but I approve the code changes once they are passing. Let's get the tests passing before merging. I'm happy to do another code review once the test failures are fixed. Thanks!

PR #1280, opened 2022-06-17
Implement span links API

jaggederest:

Implements a span links api to allow linkage of non-parent spans (e.g. originating asynchronous spans) to spans via a span context

## Why is it important?
This is a part of the open telemetry spec as seen here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#links-between-spans

It will also be useful in future for e.g. SQS or RabbitMQ instrumentation

## Checklist

- [ ] I have rebased my changes on top of the latest main branch

- [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-ruby/blob/main/CONTRIBUTING.md#testing) pass locally with my changes

### DIFF
--- a/lib/elastic_apm/spies/sqs.rb
+++ b/lib/elastic_apm/spies/sqs.rb
@@ -56,6 +56,7 @@ module ElasticAPM
             service: { resource: "#{SUBTYPE}/#{queue_name}" },
             cloud: { region: region }
           }
+          # span links added here?
         )
       end
 
--- a/lib/elastic_apm/span/context.rb
+++ b/lib/elastic_apm/span/context.rb
@@ -52,8 +52,8 @@ module ElasticAPM
           end
         @links =
         case links
-        when Links then links
-        when Array then Links.new(links)
+          when Links then links
+          when Array then Links.new(links)
         end
       end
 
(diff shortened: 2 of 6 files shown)


estolfo (approved):
Looks good but tests are failing so we should get them fixed before merging.

lib/elastic_apm/spies/sqs.rb
```diff
@@ -56,6 +56,7 @@ def self.span_context(queue_name, region)
             service: { resource: "#{SUBTYPE}/#{queue_name}" },
             cloud: { region: region }
           }
+          # span links added here?
```
jaggederest, line 59: @estolfo Is this logically where I should be adding the span links? I assume I'll need to grab the traceparent details from the initiating request

estolfo: I am not sure off the top of my head and would need to spend some time to investigate to give an answer. This is what I would suggest to investigate:
1. Test SQS by using localstack and step through the instrumentation to learn about where spans are created
2. Read the Span link spec in the apm repo
3. Asking the authors of the spec if you need any clarification
4. I believe span links are relevant when a batch of messages is processed. So I would set up a test scenario processing a batch of messages and find the point at which you can create a link between the spans
5. Look at the instrumentation in other agents. I know the Python agent has implemented this, for example

PR #1288, opened 2022-07-08
Fix for redis.call breaking when given an array

jclusso:
The redis.call method can take an array. I came across this in while debugging this [issue](https://github.com/MiniProfiler/rack-mini-profiler/issues/513) with Rack Mini Profiler. Calling `.to_s` before we call `.upcase` seems to be the best way to go about this.

### DIFF
--- a/lib/elastic_apm/spies/redis.rb
+++ b/lib/elastic_apm/spies/redis.rb
@@ -25,7 +25,7 @@ module ElasticAPM
       # @api private
       module Ext
         def call(command, &block)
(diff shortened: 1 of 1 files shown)


jaggederest:
Hi @jclusso I'd be happy to merge this if you fill out the CLA, seems like a reasonable change to me 👍

jclusso:
@jaggederest I already did. I also just signed it again to be sure.

jaggederest:
run elasticsearch-ci/docs

PR #1298, opened 2022-08-30
Pin webmock to fix CI

jaggederest:
Skipping checklist since it's just a Gemfile change. If CI passes, will merge after review


jaggederest:
> Why can't we fix our tests to use the latest version of Webmock? What is the reason we need to pin the version?

For posterity, #1299 documents the issue and provides supporting information 👍

Issue #1292, opened 2022-07-20
Pod name is getting reported as host.name rather than pod.name 


matschaffer:
## Describe the bug

When viewing the Inventory UI with Ruby APM service active in a kubernetes pod, we end up with the ruby service showing up as a "host"

## Steps to reproduce

1. Run a ruby APM instruemented service in a k8s pod
2. Ingest that into ES
3. Open the Infra UI

Can use edge-oblt for the above.

## Expected behavior

Only _hosts_ (VMs) are expected to be present in the infra UI. Since Ruby APM data contains `host.name` we end up showing that as a host as well. Other APM agents don't exhibit the same behavior.

## Additional context

Screenshot from edge-oblt

<img width="947" alt="Screenshot 2022-07-19 at 11 20 49 (1)" src="https://user-images.githubusercontent.com/690/180127465-f9332fc1-79ac-4405-b020-99e050192337.png">

https://github.com/elastic/kibana/issues/133132 was opened for a UI-layer fix, but if other APM agents don't do this, we might be looking at a bug (or at least unexpected behavior) in the Ruby agent.

cc @miltonhultgren @kuisathaverat

miltonhultgren:
I don't know much more than what I wrote in the linked issue and what can be seen on the UI side. 

We have some Ruby service that uses the APM Ruby agent that the Observability Robots team runs inside of Kubernetes, and unlike the other APM agents, the Ruby agent reports the pod name as `host.name` which makes it show up in the Inventory view as a "host".

matschaffer:
Cool. I took a swag at populating the issue template then. @estolfo are you able to help with getting this into an appropriate triage queue?

estolfo:
Thanks for reporting this @matschaffer and @miltonhultgren. We'll take a look and get back to you soon.

@jaggederest can you please take a look at this issue and see what it would take to fix it? Thanks

jaggederest:
So I don't see in the Ruby Agent code anywhere we're setting a host name as `host.name`. We report 
`lib/elastic_apm/metadata/system_info/container_info.rb:58`
```
              {
                namespace: kubernetes_namespace,
                node: { name: kubernetes_node_name },
                pod: {
                  name: kubernetes_pod_name,
                  uid: kubernetes_pod_uid
                }
              }
```

and we report

`lib/elastic_apm/transport/serializers/metadata_serializer:84`
```
          base = {
            configured_hostname: keyword_field(system.configured_hostname),
[...]

PR #1314, opened 2022-10-06
Use hostname as the default Kubernetes pod name

estolfo:
Use the hostname as the Kubernetes pod name in the Container Info metadata if the pod id is parsed from `/proc/self/cgroup`.
This will help solve part of https://github.com/elastic/apm-agent-ruby/issues/1292, which is partly due to a server issue and partly due to this missing usage of the hostname as the default pod name in the Ruby agent.


miltonhultgren:
Thanks for fixing this! I'm not sure if my review means anything here since I don't really understand the details of Ruby or APM Agents.

trentm (approved):
LGTM (with the caveat that my Ruby is ... poor :)

PR #1320, opened 2022-11-10
action: enable matrix testing (version, framework)

v1v:

Use GitHub actions 

## Why is it important?

Before the upcoming sunset for JenkinsCI, let's start moving things forward

## Checklist

- [ ] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).
- [ ] I have rebased my changes on top of the latest main branch

- [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-ruby/blob/main/CONTRIBUTING.md#testing) pass locally with my changes

## Some facts

While the provisioner didn't happen GH runners were already provisioned

image

## Actions

Should we push those docker images somewhere else, since their definitions are already in this project and open to be viewed, by doing so, there is no need to docker login hence security wise everyone could contribute...

```
pull access denied for docker.elastic.co/observability-ci/jruby, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```


.pre-commit-config.yaml
```diff
@@ -13,6 +13,4 @@ repos:
     rev: current
     hooks:
     -   id: check-bash-syntax
-    -   id: check-jenkins-pipelines
-    -   id: check-jjbb
```
v1v, line 17: These checks are not really something to be linted when moving to GitHub actions, so this should help us to avoid the need to use the internal docker registry.

.ci/.jenkins_ruby.yml
```diff
@@ -1,4 +1,4 @@
-RUBY_VERSION:
+VERSION:
```
v1v, line 1: Being as generic as possible so the GH action in charge of generating the matrix can work

.ci/Jenkinsfile
```diff
@@ -62,34 +62,6 @@ pipeline {
         }
       }
     }
-    stage('Sanity checks') {
```
v1v, line 65: Moved to the pre-commit section in the GH workflow

.ci/Jenkinsfile
```diff
@@ -362,7 +321,8 @@ def runScript(Map params = [:]){
   def framework = params.framework
   log(level: 'INFO', text: "${label}")
   retry(2){
-    withEnv(["HOME=${env.WORKSPACE}", "PATH=${env.PATH}:${env.WORKSPACE}/bin"]) {
+    withEnv(["HOME=${env.WORKSPACE}", "PATH=${env.PATH}:${env.WORKSPACE}/bin",
+             "RUBY_VERSION=${ruby}"]) {
```
v1v, line 325: Just in case the `RUBY_VERSION` env variable might be needed, though maybe it should not be required..

reakaleek:
Done in https://github.com/elastic/apm-agent-ruby/pull/1366

PR #1329, opened 2022-12-07
Support daily snapshots

amannocci:
Signed-off-by: Adrien Mannocci <adrien.mannocci@elastic.co>

* Add support for daily snapshots.

Screenshot 2022-12-07 at 17 34 46

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -25,6 +25,7 @@ pipeline {
     OPBEANS_REPO = 'opbeans-ruby'
     REFERENCE_REPO = '/var/lib/jenkins/.git-references/apm-agent-ruby.git'
     SLACK_CHANNEL = '#apm-agent-ruby'
+    RUBY_DOCKER_TAG = 'ruby:2.6'
   }
   options {
     timeout(time: 2, unit: 'HOURS')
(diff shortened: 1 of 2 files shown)


kuisathaverat (approved):

.ci/Jenkinsfile
```diff
@@ -25,6 +25,7 @@ pipeline {
     OPBEANS_REPO = 'opbeans-ruby'
     REFERENCE_REPO = '/var/lib/jenkins/.git-references/apm-agent-ruby.git'
     SLACK_CHANNEL = '#apm-agent-ruby'
+    RUBY_DOCKER_TAG = 'ruby:2.6'
```
estolfo, line 28: what does this tag mean?

amannocci: It's the docker image version used to install some gems and perform a `rake build`.

estolfo: Ruby 2.6 is EOL'ed so maybe we should use a later version? 3.2.0 is the latest stable.

v1v (approved):

.ci/Jenkinsfile
amannocci: Does it matter if we generate a gem with a 3.2.x rather than 2.6?
Is there an impact on the APM-agent compatibility?

estolfo: Sorry if I'm not understanding the context properly of this part of the code. What actions are taken or code is executed with this version of Ruby?
The only point worth considering is that if there is some security issue with Ruby 2.6 and it's not actively maintained, the gems built using it might be at risk.

amannocci: Ruby 2.6 is currently used to build release gems. This PR build gems in the same way and upload them in a GCS bucket.
Are there any specific details that can be impacted by a ruby bump version?
If not then we can bump the ruby version to the latest one?

estolfo: I am pretty sure it doesn't matter what version you use to build the gem. I just tried locally to build the gem with Ruby v3.2.0, then installed it with v3.0.0 and v2.7.0.
I'm having trouble installing earlier versions of Ruby (2.3.0, our minimum supported Ruby) on my machine right now but I wanted to check that I could install the gem with 2.3.0, because then we'd know we had all our bases covered.

It should also be fine to leave the version here at 2.6, if you prefer. My concerns were only very minor.

estolfo: Yes, sure

estolfo (approved):

PR #1351, opened 2023-01-21
Close the read pipe at the right moment

### LINKED ISSUE #1350: Timeouts from APM server causes a file descriptor leak when HTTP compression is enabled
## Describe the bug

After the HTTP request to the APM server gets a time-out, `ElasticAPM::Transport::Connection` cannot close the _write_ pipe because it silently raises `Errno::EPIPE Exception: Broken pipe`.

## Steps to reproduce

1. Configure `ELASTIC_APM_SERVER_URL` so that you can get time-outs
2. Watch file descriptors to grow

## Expected behaviour

The number of file descriptors stays constant and does not grow continuously.

## Environment

- **OS**: Linux
- **Ruby version**: 3.2.0
- **Agent version**: 4.6.0

matheussilvasantos:

Closes the read pipe whether the HTTP request thread was executed or not.

## Why is it important?

If we only close the read pipe after the HTTP request inside the child thread, a file descriptor leak can happen.

## Checklist

## Related issues

- Closes #1350

### DIFF
--- a/lib/elastic_apm/transport/connection/http.rb
+++ b/lib/elastic_apm/transport/connection/http.rb
@@ -75,7 +75,15 @@ module ElasticAPM
           @closed.make_true
 
           @wr&.close
-          return if @request.nil? || @request&.join(5)
+
+          if @request&.join(5)
+            @rd&.close
+            return
+          end
+
+          @rd&.close
+
+          return if @request.nil?
 
           error(
             '%s: APM Server not responding in time, terminating request',
@@ -117,8 +125,6 @@ module ElasticAPM
               error(
                 "Couldn't establish connection to APM Server:\n%p", e.inspect
               )
-            ensure
-              @rd&.close
             end
           end
         end
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -35,6 +35,11 @@ endif::[]
 [[release-notes-4.x]]
 === Ruby Agent version 4.x
 
+[float]
+===== Fixed
+
+- Fix growing number of open file descriptors when HTTP request to APM is never sent {pull}1351[#1351]
+
 [[release-notes-4.6.0]]
 ==== 4.6.0
 
--- a/spec/elastic_apm/transport/connection/http_spec.rb
+++ b/spec/elastic_apm/transport/connection/http_spec.rb
@@ -95,6 +95,26 @@ module ElasticAPM
         end
       end
 
+      describe 'when exception is raised before write' do
+        subject! { described_class.new(config) }
+
+        before do
+          allow(subject).to receive(:post).and_raise
+        end
+
+        it 'closes the connection on close' do
+          stub = build_stub(body: /{"msg": "hey!"}/, headers: headers)
+          subject.open(url)
+
+          sleep 0.2
+          subject.write('{"msg": "hey!"}')
+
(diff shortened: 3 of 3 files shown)


estolfo:
jenkins please test

matheussilvasantos:
Hi @estolfo, is there anything I have to fix on my side?

estolfo:
Hi @matheussilvasantos thanks for following up. There are some test failures that I need to look at. They don't have to do with this code; it looks like one of the gems we test with dropped support for a version of Ruby we have in our test matrix.
I'll take a look at the test failures and have an update soon, but there's nothing I need from your side right now. Thanks again for your contribution!

estolfo:
Hi @matheussilvasantos can you rebase against main now? The test failures have been resolved in the latest commit.

matheus-silva-santos:
Hi @estolfo. I've rebased it.

matheussilvasantos:
Hi @estolfo. I've rebased it.

estolfo:
jenkins please test

lib/elastic_apm/transport/connection/http.rb
```diff
@@ -75,7 +75,15 @@ def close(reason)
...
-          return if @request.nil? || @request&.join(5)
+
+          if @request&.join(5)
+            @rd&.close
+            return
+          end
+
+          @rd&.close
+
+          return if @request.nil?
```
estolfo, line 86: I would suggest simplifying this to:

```ruby
          @request&.join(5) # will return nil if @request is nil, will return nil if 5s expires, will return the thread if the thread exits
          @rd&.close
          return if @request.nil?
```

because you will always call `@request&.join(5)` and then `@rd&.close` no matter what #join returns.

estolfo (changes requested):

lib/elastic_apm/transport/connection/http.rb
matheussilvasantos: The condition before was `return if @request.nil? || @request&.join(5)`, so we need to return if `@request&.join(5)` returns a "truthy" value. `return if @request.nil?` won't make it return when `@request&.join(5)` returns a truthy value.

estolfo: I understand now, thanks for the explanations.

estolfo:
Thanks @matheussilvasantos !

matheussilvasantos:
Thank you too, @estolfo. Do you intend to release it at RubyGems any time soon?

estolfo:
@matheussilvasantos we just did a release within the last two weeks so I'm going to wait a little longer before doing another one, given that this is a bit of an edge case.

matheussilvasantos:
Hi, again, Emily. Would you have an estimate of when you will release it? This caused issues in production once, and we want to avoid creating a dependency on GitHub by pointing our gem to the repo. Thanks!

estolfo:
Hi @matheussilvasantos I was hoping to get [this issue](https://github.com/elastic/apm-agent-ruby/issues/1358) resolved and in the next release so was holding off. I'll give it a few more days and if it's not resolved by then, I'll do a release including this issue's fix.

estolfo:
@matheussilvasantos v4.6.1 is released with this fix. Thanks again!

matheussilvasantos:
@estolfo, thank you too!

PR #1352, opened 2023-01-25
latest sqlite3 does not support < ruby v2.7

estolfo:
The latest test failures show that the most recent release of [sqlite3 (1.6.0)](https://rubygems.org/gems/sqlite3/versions/1.6.0-x86_64-linux) doesn't support < ruby v2.7


spec/scripts/features.sh
```diff
@@ -43,6 +43,7 @@ IMAGE_NAME=${IMAGE_NAME} RUBY_VERSION=${VERSION} USER_ID="$(id -u):$(id -g)" \
   -e INCLUDE_SCHEMA_SPECS=1 \
   -e JDK_JAVA_OPTIONS="${JDK_JAVA_OPTIONS}" \
   -e JRUBY_OPTS="${JRUBY_OPTS}" \
+  -e FRAMEWORKS="${FRAMEWORKS}" \
```
estolfo, line 46: This is necessary, otherwise the `bundle update` line below won't know the framework and install the correct version of sqlite.

PR #1382, opened 2023-04-13
Faraday response can be nil

jclusso:
This bug came up in production for us.

`NoMethodError · undefined method '[]' for nil:NilClass` since  `e.response` can be nil. There is a `response_status` method which will work regardless of if response is nil.


jqr:
Came here to make nearly the same PR. :+1:

estolfo:
jenkins run tests

ericproulx:
@estolfo do you know when 4.6.3 will be released ?

jclusso:
@ericproulx you can just bundle off main for now. that's what we're doing.

PR #1395, opened 2023-06-17
Add keyword args for `span_method` helper so you can pass in additional options

jclusso:
I wanted to be able to use the other options available on [ElasticAPM.start_span](https://www.elastic.co/guide/en/apm/agent/ruby/current/api.html#api-agent-start_span).


estolfo:
Hi @jclusso thanks for this contribution. Would you mind adding some tests? Thanks!

jclusso:
There are no tests for me to base this off of which was why I didn’t add any. I couldn’t find any tests that covered the existing parameters.

PR #1415, opened 2023-09-07
changes/fixes to metadata.cloud.* fields collected for GCP

trentm:
- cloud.instance.id must be a string
- change cloud.project.id to be the 'projectId' (per spec change)
- drop cloud.project.name
- fix cloud.machine.type

Closes: https://github.com/elastic/apm-agent-ruby/issues/1414

### DIFF
--- a/lib/elastic_apm/metadata/cloud_info.rb
+++ b/lib/elastic_apm/metadata/cloud_info.rb
@@ -102,7 +102,7 @@ module ElasticAPM
         self.project_id = metadata["project"]["projectId"]
         self.availability_zone = zone
         self.region = zone.split("-")[0..-2].join("-")
-        self.machine_type = metadata["instance"]["machineType"].split("/")&.at(-1)
+        self.machine_type = metadata["instance"]["machineType"].split("/")[-1]
       rescue HTTP::TimeoutError, HTTP::ConnectionError
         nil
       end
--- a/lib/elastic_apm/metadata/cloud_info.rb
+++ b/lib/elastic_apm/metadata/cloud_info.rb
@@ -97,13 +97,12 @@ module ElasticAPM
         zone = metadata["instance"]["zone"]&.split("/")&.at(-1)
 
         self.provider = "gcp"
-        self.instance_id = metadata["instance"]["id"]
+        self.instance_id = metadata["instance"]["id"].to_s
         self.instance_name = metadata["instance"]["name"]
-        self.project_id = metadata["project"]["numericProjectId"]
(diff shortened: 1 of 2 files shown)


lib/elastic_apm/metadata/cloud_info.rb
```diff
@@ -97,13 +97,12 @@ def fetch_gcp
         zone = metadata["instance"]["zone"]&.split("/")&.at(-1)
 
         self.provider = "gcp"
-        self.instance_id = metadata["instance"]["id"]
+        self.instance_id = metadata["instance"]["id"].to_s
```
estolfo, line 100: The [issue description](https://github.com/elastic/apm/issues/826) says to check that the int is being handled correctly but doesn't say it should be collected as a string. Is using a string the solution?

lib/elastic_apm/metadata/cloud_info.rb
```diff
@@ -97,13 +97,12 @@ def fetch_gcp
...
         self.instance_name = metadata["instance"]["name"]
-        self.project_id = metadata["project"]["numericProjectId"]
-        self.project_name = metadata["project"]["projectId"]
+        self.project_id = metadata["project"]["projectId"]
         self.availability_zone = zone
         self.region = zone.split("-")[0..-2].join("-")
-        self.machine_type = metadata["instance"]["machineType"]
+        self.machine_type = metadata["instance"]["machineType"].split("/")&.at(-1)
```
estolfo, line 105:
```suggestion
        self.machine_type = metadata["instance"]["machineType"].split("/")[-1]
```

estolfo (changes requested):
Thanks for doing this for the Ruby agent!

lib/elastic_apm/metadata/cloud_info.rb
trentm: This was a bug in a the Ruby impl unrelated to the spec change for https://github.com/elastic/apm/issues/826.

The spec had already said:

> `instance.id` as a string

and the intake schema requires a string (https://github.com/elastic/apm-agent-ruby/blob/8d4a30cd3f592d6fcb7fc97826d71ec6d908b4b7/spec/fixtures/metadata.json#L52-L57).

estolfo (approved):
Thanks!

estolfo:
@trentm can you rebase against main? I've just merged in a fix for the test failure on ruby 2.4.

trentm:
> @trentm can you rebase against main? I've just merged in a fix for the test failure on ruby 2.4.

Done.

PR #1419, opened 2023-09-25
Use response method on Faraday error for older versions of the library

### LINKED ISSUE #1418: Elastic APM v4.7.0 breaks versions of Faraday < 1.3.0
In an application with Faraday v0.17.4, experienced new errors after upgrading the Elastic APM gem from v4.6.0 to v4.7.0.  Elastic APM is looking for a `response_status` on Faraday errors that does not exist in older versions of Faraday.

errors were:
```
undefined method `response_status' for #<Faraday::TimeoutError:0x00007f401377b2d0>
undefined method `response_status' for #<Faraday::SSLError:0x00007fbcac2e7960>
undefined method `response_status' for #<Faraday::ConnectionFailed:0x00007fae7dcd8420> 
```

Here is the change in v4.7.0: https://github.com/elastic/apm-agent-ruby/compare/v4.6.0...v4.7.0#diff-4b5630a4907f497d4416de76c8dd5a162f4a525222dc3c11bdcef2182a9cdbcdR112

`response_status` was introduced to Faraday in this commit: https://github.com/lostisland/faraday/commit/023cc413351ce6c4a7cc01a3037dc2fcbe3d7dbc and is only available in faraday versions >= 1.3.0

estolfo:
Resolves #1418

### DIFF
--- a/lib/elastic_apm/spies/faraday.rb
+++ b/lib/elastic_apm/spies/faraday.rb
@@ -109,7 +109,8 @@ module ElasticAPM
                   yield req if block
                 end
               rescue Faraday::ClientError, Faraday::ServerError => e # Faraday::Response::RaiseError
-                status = e.response_status
+                status = e.response_status if e.respond_to?(:response_status)
+                status ||= e.response&.fetch(:status)
                 http = span&.context&.http
                 if http && status
                   http.status_code = status.to_s


picandocodigo (approved):
LGTM! I reproduced the original error and this change fixes it :+1:

PR #1423, opened 2023-11-03
Fix `ActionDispatchSpy#render_exception` for Rails 7.1

jclusso:
In Rails 7.1 `render_exception` was changed to take an ActionDispatch::ExceptionWrapper instead of an Exception. I've added support for the new way while keeping support for pre 7.1.

### DIFF
--- a/lib/elastic_apm/spies/action_dispatch.rb
+++ b/lib/elastic_apm/spies/action_dispatch.rb
@@ -24,11 +24,19 @@ module ElasticAPM
     class ActionDispatchSpy
       # @api private
       module Ext
-        def render_exception(env, exception)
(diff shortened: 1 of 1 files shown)


dalibor (approved):
@estolfo Could we get this released, please?

janko:
We're experiencing this issue as well, in the form of `undefined method `cause' for #<ActionDispatch::ExceptionWrapper...>` errors in our error monitoring service.

picandocodigo (approved):

picandocodigo:
@elasticmachine run elasticsearch-ci/docs

assembler:
+1 for releasing this

PR #1435, opened 2024-01-18
Address machineType not being returned in GCP metadata

picandocodigo:
Related: https://github.com/elastic/apm/issues/847

### DIFF
--- a/lib/elastic_apm/metadata/cloud_info.rb
+++ b/lib/elastic_apm/metadata/cloud_info.rb
@@ -94,15 +94,18 @@ module ElasticAPM
         return unless resp.status == 200
         return unless (metadata = JSON.parse(resp.body.to_s))
 
-        zone = metadata["instance"]["zone"]&.split("/")&.at(-1)
-
         self.provider = "gcp"
-        self.instance_id = metadata["instance"]["id"].to_s
-        self.instance_name = metadata["instance"]["name"]
         self.project_id = metadata["project"]["projectId"]
+
+        return unless metadata['instance']
+
+        zone = metadata["instance"]["zone"]&.split("/")&.at(-1)
         self.availability_zone = zone
         self.region = zone.split("-")[0..-2].join("-")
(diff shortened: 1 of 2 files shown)


lib/elastic_apm/metadata/cloud_info.rb
```diff
@@ -102,7 +102,7 @@ def fetch_gcp
         self.project_id = metadata["project"]["projectId"]
         self.availability_zone = zone
         self.region = zone.split("-")[0..-2].join("-")
-        self.machine_type = metadata["instance"]["machineType"].split("/")[-1]
+        self.machine_type = metadata["instance"]["machineType"]&.split("/")&.[](-1)
```
estolfo, line 105: Should we use Hash#dig instead in case `"instance"` doesn't exist for some reason?

picandocodigo: Some previous values depend on `instance`, should I rewrite it and update it to use `fetch` in the whole function or just for `machineType`? It does feel more readable than the safe navigation operator, so I'll update at least this line to `fetch` anyway.

estolfo: Ah right, we use `instance` above already. It's up to you, whatever you think is most readable and consistent.

picandocodigo: I updated it to return early if `instance` doesn't exist, and I changed the code to look just like the code that sets the zone so these two lines are consistent:
```ruby
zone = metadata["instance"]["zone"]&.split("/")&.at(-1)

self.machine_type = metadata["instance"]["machineType"]&.split("/")&.at(-1)
```

estolfo (approved):

lib/elastic_apm/metadata/cloud_info.rb
estolfo: perfect, thank you!

PR #1437, opened 2024-01-24
Use Faraday's URI util in Faraday context

dalibor:
Inside `FaradaySpy` i.e. within the Faraday context we need to use the Faraday's URI utility, otherwise we get errors in the elastic-apm-ruby that did not fail in the Faraday code. Specifically, here's an example for when using `Addressable::URI` as Faraday's `default_uri_parser`:

```
Faraday::Utils.default_uri_parser = Addressable::URI.method(:parse)
```

Results:

```
>> URI('http://examplé.com')
Traceback (most recent call last):
        2: from (irb):14
        1: from (irb):15:in `rescue in irb_binding'
URI::InvalidURIError (URI must be ascii only "http://exampl\u00E9.com")

>> Addressable::URI.parse('http://examplé.com')
=> #<Addressable::URI:0x2245c URI:http://examplé.com>

>> Faraday::Utils.URI('from@examplé.com')
=> #<Addressable::URI:0x235a0 URI:http://examplé.com>
```

Fixes an error in FaradaySpy.

## Why is it important?

To fix errors in Elastic APM context.

## Checklist


dalibor:
@picandocodigo Any chance we could get this one reviewed please?

dalibor:
@picandocodigo Thanks for looking into this. There are some details in the PR description, but basically I'm configuring Faraday to use `Addressable::URI` as its URI parser which works well with URLs that have Internationalized Domain Names (IDN) while `URI` standard library does not. And then I'm getting URI parsing error inside the ElasticAPM code i.e. in`ElasticAPM::Spies::FaradaySpy::Ext` context that's external to the Faraday's `run_request` method.

I did some more investigation now and figured that this patch is not sufficient because it then fails at a later point inside `ElasticAPM::Span::Context::Destination` because ElasticAPM internally uses URI, so it fails at a different point.

I think I'll do URL sanitization on my side like the following:

```
>> uri = Addressable::URI.parse("http://examplé.com"); p uri; p URI.parse(uri.normalize.to_s); nil
#<Addressable::URI:0x3b1dc URI:http://examplé.com>
#<URI::HTTP http://xn--exampl-gva.com/>
```

Not sure if it makes sense for ElasticAPM to use `Addressable::URI` instead of `URI`? Ideally, ElasticAPM should not fail inside its spy context. It should handle its own failures and log a warning if that happens which is the issue here.

I will close this PR since I have a work-around and leave it to you if you want to do any of these two ideas. Thanks!

PR #1531, opened 2025-03-28
Add ability to configure sampling rate based on transaction span name

### LINKED ISSUE #1334: Ability to override transaction_sample_rate on a per controller basis
## Is your feature request related to a problem? Please describe.
Sampling is not very usefull when you have a few controllers that make up most of your applications traffic.

## Describe the solution you'd like
Ability to override the transaction_sample_rate on a per controller basis.

## Describe alternatives you've considered
Playing around with `ElasticAPM.add_filter`, but not sure this is ideal.

jclusso:

Adds new configuration option so you can set the sampling rate by transaction name. I'm not 100% certain this is the correct way to do this, so please let me know if I should be going about this in a different manner.

```ruby
{
  transaction_sample_rate_by_name: {
    "UsersController#index": 1.0
    "HealthController#ping": 0
  }
}
```

## Why is it important?

When you have an application with certain portions that get a significant amount of traffic, there is currently no good way to control the sampling. Either you are blind to low volume transactions or have to collect a lot more samples than you need. You can currently get around this with [Tail-based sampling](https://www.elastic.co/guide/en/observability/current/apm-tail-based-samling-config.html), but this is not available on serverless. This also still requires you send a ton of data to your APM server that you may not need to.

## Checklist

- [X] I have signed the [Contributor License Agreement](https://www.elastic.co/contributor-agreement/).
- [X] My code follows the style guidelines of this project (See `.rubocop.yml`)
- [X] I have rebased my changes on top of the latest main branch
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing [**unit** tests](https://github.com/elastic/apm-agent-ruby/blob/main/CONTRIBUTING.md#testing) pass locally with my changes
- [X] I have made corresponding changes to the documentation
- [ ] I have updated [docs/release-notes/index.md](../docs/release-notes/index.md)
- [ ] I have updated [docs/reference/supported-technologies.md](../docs/reference/supported-technologies.md)

## Related issues

Resolves #1334


jclusso:
I have this deployed in production. I thought everything was working until I realized that transactions that are under Requests in Rails seem to have their throughput heavily increased. For Sidekiq related transactions, everything seems fine though.

@estolfo any suggestions?

jclusso:
I think I found the issue. I wasn't setting the `transaction.sample_rate` to be 0 when the transaction wasn't sampled.

jclusso:
@estolfo any chance you could give some input on this? This is the only solution for moving to serverless if you rely on tail-based sampling.

estolfo:
Hi @jclusso I'll be able to take a look within the next week. In the meantime, can you please rebase? Tests are passing on main now so they should be reliable for this PR as well.

jclusso:
@estolfo i rebased it and the tests seem to be passing if I did everything correctly.

estolfo:
Hi @jclusso thanks for this contribution. I've taken a look but I was wondering if you could explain a bit more what the goals are of this feature. The description says that you'd like to have different sample rates based on transaction name. But the code shows that the span name determines the sample rate on the transaction.
And that the first span on a transaction, if it has a special sample rate, determines the sample rate for the whole transaction. 
Would you mind explaining a bit more how you want it to work? thanks!

jclusso:
> Hi @jclusso thanks for this contribution. I've taken a look but I was wondering if you could explain a bit more what the goals are of this feature. 

The goal of the feature is to be able to sample based on transaction name. In Rails this commonly is a controller or worker name.

> The description says that you'd like to have different sample rates based on transaction name. But the code shows that the span name determines the sample rate on the transaction. And that the first span on a transaction, if it has a special sample rate, determines the sample rate for the whole transaction. Would you mind explaining a bit more how you want it to work? thanks!

For Rails controllers, the transaction name appears to be set on the first span and the transaction name itself is Rack. Later on it gets set to the first span though. That is why this needs to look at the first span.

jclusso:
Might also be worth noting that we're using this in production in two applications since early April. From what we can tell it appears to be working as described in my example config.

estolfo (approved):

estolfo:
run docs-build

PR #1597, opened 2026-01-14
chore: delete SECURITY.md to use organization-wide policy

ismisepaul:
This PR deletes the repository-specific SECURITY.md file to consolidate to a single source of truth for the security policy, which is https://github.com/elastic/.github/blob/main/SECURITY.md.


hegerchr (approved):
LGTM

PR #1619, opened 2026-05-20
fix(ci,test): stabilize JRuby and Rails matrix by hardening dependency resolution, test synchronization, and legacy image builds

hegerchr:
### Summary

This PR started as a dev workflow fix and was expanded to address several CI instabilities uncovered during matrix runs. The final result is a clean CI run by resolving framework compatibility drift, JRuby test races, and legacy Debian image build failures.

### What changed

- BDD runner reliability: 
  - Updated the BDD runner invocation to use Bundler-managed executables so cucumber is always available in CI environments.
- Matrix compatibility updates:
  - Excluded grape master on Ruby versions no longer supported upstream.
  - Pinned delayed_job for Rails 4.x and 5.x compatibility to avoid ActiveJob adapter NameError failures.
  - Added a targeted shoryuken compatibility constraint for the Ruby 3.4 + Rails 6.1 combination.
- Legacy Debian image resilience:
  - Added Debian Buster archive fallback handling for CI spec image builds.
  - Added required build tools for JRuby-based images, including curl and GPG tooling needed by gosu verification.
  - Added a safe default base image argument in the spec Docker build definition.
- JRuby test flake fixes:
  - Updated the Rails JSON schema integration test wait condition to wait for the full expected payload set.
  - Hardened MockIntake synchronization with mutex-protected writes and synchronized count/snapshot reads.
  - Fixed wait_for logic so each expected event type must independently satisfy its threshold, preventing cross-category cancellation.

### Why this was needed

- Upstream framework changes introduced new resolver failures in the matrix.
- JRuby true concurrency exposed race and visibility issues in test intake polling.
- EOL Debian Buster repositories caused deterministic image build failures in legacy targets.
- Missing base-image utilities caused secondary CI build regressions in JRuby jobs.

### Outcome

CI is now passing after these follow-up commits. Changes are focused on test, CI, and build stability, with no intended production runtime behavior change.

### Notes for reviewers

The PR now includes both the original dev workflow fix and the required CI hardening discovered during validation.
Most changes are in test and CI support paths and are scoped to restore matrix reliability across legacy and JRuby targets.

### DIFF
--- a/bin/dev
+++ b/bin/dev
@@ -36,7 +36,10 @@ FRAMEWORKS = options.fetch(:frameworks, default_frameworks)
 IMAGE_PATH_SAFE = RUBY_IMAGE.tr(':', '_')
 IMAGE_NAME = "apm-agent-ruby:#{IMAGE_PATH_SAFE}"
 VENDOR_PATH = "/vendor/#{IMAGE_PATH_SAFE}"
-DOCKER_PLATFORM = RUBY_IMAGE.include?('jruby') ? 'linux/amd64' : nil
+legacy_ruby_image = RUBY_IMAGE.match?(/^ruby:(2\.[0-9]+|1\.[0-9]+)/)
+DOCKER_PLATFORM = if RUBY_IMAGE.include?('jruby') || legacy_ruby_image
+                    'linux/amd64'
+                  end
 
 def run(cmd)
   env = ["IMAGE_NAME=#{IMAGE_NAME}"]
--- a/bin/dev
+++ b/bin/dev
@@ -26,14 +26,23 @@ end.parse!
 USER_ID_GROUP = %w[u g].map { |f| `id -#{f}`.chomp }.join(':')
 
 RUBY_IMAGE = options.fetch(:image, 'ruby:latest')
-FRAMEWORKS = options.fetch(:frameworks, 'rails,sinatra,grape')
+default_frameworks = if RUBY_IMAGE.include?('jruby')
+                       'rails-6.1,sinatra-2.2,grape-1.6'
+                     else
+                       'rails,sinatra,grape'
+                     end
+FRAMEWORKS = options.fetch(:frameworks, default_frameworks)
 
 IMAGE_PATH_SAFE = RUBY_IMAGE.tr(':', '_')
 IMAGE_NAME = "apm-agent-ruby:#{IMAGE_PATH_SAFE}"
 VENDOR_PATH = "/vendor/#{IMAGE_PATH_SAFE}"
+DOCKER_PLATFORM = RUBY_IMAGE.include?('jruby') ? 'linux/amd64' : nil
 
 def run(cmd)
-  "IMAGE_NAME=#{IMAGE_NAME} #{cmd}".tap do |str|
+  env = ["IMAGE_NAME=#{IMAGE_NAME}"]
+  env << "DOCKER_DEFAULT_PLATFORM=#{DOCKER_PLATFORM}" if DOCKER_PLATFORM
+
+  "#{env.join(' ')} #{cmd}".tap do |str|
     puts str
     system str
   end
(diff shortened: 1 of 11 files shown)


bin/dev
```diff
@@ -26,14 +26,26 @@ end.parse!
...
+                       'rails,sinatra,grape'
+                     end
+FRAMEWORKS = options.fetch(:frameworks, default_frameworks)
 
 IMAGE_PATH_SAFE = RUBY_IMAGE.tr(':', '_')
 IMAGE_NAME = "apm-agent-ruby:#{IMAGE_PATH_SAFE}"
 VENDOR_PATH = "/vendor/#{IMAGE_PATH_SAFE}"
+legacy_ruby_image = RUBY_IMAGE.match?(/^ruby:(2\.[0-9]+|1\.[0-9]+)/)
+DOCKER_PLATFORM = if RUBY_IMAGE.include?('jruby') || legacy_ruby_image
+                    'linux/amd64'
```
estolfo, line 41: Do you know why this is necessasry?

spec/support/mock_intake.rb
```diff
@@ -29,6 +29,7 @@
 
 class MockIntake
   def initialize
+    @mutex = Mutex.new
```
estolfo, line 32: I'm curious what failure this is fixing?

docker-compose.yml
```diff
@@ -14,10 +13,18 @@ services:
...
+        FRAMEWORKS: ${FRAMEWORKS:-rails,sinatra,grape}
+        VENDOR_PATH: ${VENDOR_PATH:-/vendor/ruby_3.3}
     image: '$IMAGE_NAME'
     environment:
       HOME: '/tmp'
       MONGODB_URL: 'mongodb:27017'
+      USER_ID: ${USER_ID:-1000}
+      LOCAL_USER_ID: ${LOCAL_USER_ID:-1000}
+      LOCAL_GROUP_ID: ${LOCAL_GROUP_ID:-1000}
+      RUBY_VERSION: ${RUBY_VERSION:-3.3}
```
estolfo, line 27: Why is this needed?

estolfo (commented):
This is a lot of gymnastics and I don't know the motivation behind a lot of the changes, but maybe you can explain where I put specific comments?

bin/dev
hegerchr: These lines are needed to make local runs reliable on Apple Silicon and match CI behavior for older images.

Many JRuby and legacy Ruby images are only published for amd64 (or behave inconsistently on arm64), so without forcing linux/amd64 Docker may fail to pull/run them, or run with different dependency resolution than CI. The condition in dev ensures only JRuby and old Ruby targets are pinned to amd64, while newer Ruby images can still use native arm64 for speed.

It might only be relevant for local testing and development. Do you think they should be omitted because they may not be relevant for CI?

spec/support/mock_intake.rb
hegerchr: `@mutex = Mutex.new` fixes a JRuby race in the mock intake collector where the test thread could observe partially published state from the intake thread.

In CI this showed up as intermittent failures in the Rails JSON schema example (`metadata` being `nil`, or `transactions.fetch(0)` raising `IndexError`) even after `wait_for` appeared satisfied. JRuby runs true parallel threads, so unsynchronized writes/reads to the shared arrays in mock_intake.rb can be observed out of order across threads. The mutex makes these updates/read checks synchronized and removes that flake.

docker-compose.yml
hegerchr: This is to make `docker compose` runs deterministic when those env vars are not pre-exported.

Without defaults, Compose can pass empty values into the container, which caused user/group mapping and image-tag interpolation to break in some local runs. Adding defaults in docker-compose.yml keeps permissions and runtime selection stable out of the box, while still allowing CI or developers to override values explicitly.

estolfo (approved):
I can't say I've done deep research on the justification behind all the changes but they seem to make the tests pass so I think it's ok to merge.

PR #1624, opened 2026-06-16
Use version 7.0 as http gem upper bound in gemspec

estolfo:
Increase upper bound for `http` gem to 7.0 from 6.0 in the gemspec.


hegerchr (approved):
LGTM

PR #1629, opened 2026-06-18
Use headers method directly for forwards compatibility

estolfo:
Up until version 6.0 of the `http` gem, there was a mixin included in `HTTP::Request` that allow headers to be access via `[]` on the object. As of 6.0, that mixin is removed so `.headers` needs to be called explicitly on the request object to access the headers.
This change ensures compatibility with all http versions after 3.0 up to the current 6.0 series.


hegerchr (approved):
LGTM

